PR #14, opened 2015-06-11
Avoid to increment ptr in iio_buffer_first() for disabled channels

rofferom:
Previous implementation of iio_buffer_first() considered that a
struct iio_buffer always contains samples for all the channels of the
device, for disabled channels. This is actually wrong, and disabled
channels should be ignored during offset calculation of a given enabled
channel.

It implements the same strategy as iio_buffer_foreach_sample().

Change-Id: I6df5a3ec766d0490d94d72cf67db2e6c1d3b76c2


pcercuei:
This makes perfect sense. Thanks for the fix!

PR #15, opened 2015-06-17
Local backend: free device elements when device creation fails

jlbilou93:
Simple clean when device creation fails.
free iio_device allocated elements


pcercuei:
All the device elements are freed within free_device(), so the code as it is right now does not leak memory in any error case.

Issue #16, opened 2015-06-22
Enhance buffer API


rofferom:
I have some requests about buffer API. Actually I have some needs, but I would like to have a feedback before starting my development, mainly to know if the ideas themself are correct for libiio, and to have some clues about the expected implementation. The goal is to provide a pull request that has a sense...

My goal is to fetch samples from three differents sensors that are running at different rates, in an unique thread. I would like to :
* Make a blocking read on my faster sensor (no problem here, it's already available in libiio API)
* Read the other sensors. If no data are available, get EAGAIN. There is already a iio_context_set_timeout() but it is global to the context, not locally defined for a buffer

Furthermore, I would like to have a way to control the size of the buffer on kernel side. From what I understand, libiio configures the buffer at the exact size of the read done by iio_buffer_refill(), and this size is configured in iio_device_create_buffer(). Actually I would like to use kernel buffer to store some extra data for me, in case I'm a little bit late in my processing, so a little tuning of the kernel buffer size would help me a lot.

To conclude, I would like to add these functionalities in the buffer API :
* Support of non-blocking reads
* Kernel buffer size configuration

An extra feature is to export the filedescriptor of the buffer to read (or return -1 if no pollable filedescriptor is available), to do the poll() myself in my acquisition loop.

pcercuei:
Hi,

First, about the kernel size configuration. Right now the high-speed interface (mmap + ioctl based) uses 4 kernel buffers (hard-coded value). The low-speed interface (stdio) uses only one. For now I will increase the number of buffers to 4 as well as a quick fix, but it would be great to have the number of buffers configurable through an API function so that you can adjust the latency.

The question about non-blocking reads is a bit tricky, if you consider that the applications using libiio must run the exact same independently of the backend used (local or network) and the OS.

For instance, exporting the file descriptor is not a solution, as you would get a different behaviour whether you are running with the local backend or the network backend; with the latter, you have to ask the server for data first, so polling the file descriptor won't work.

The solution might be to add API functions "iio_buffer_refill_nonblock" and "iio_buffer_push_nonblock" (alternatively, "iio_buffer_set_nonblock" and have "iio_buffer_refill" and "iio_buffer_push" return -EAGAIN). Then, you can handle this differently in the backends so that it works homogeneously across all backends.

rofferom:
[...]

PR #17, opened 2015-06-30
Enhance buffer API

rofferom:
Implements buffer API enhancement discussed in issue #16

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -103,6 +103,11 @@ int iio_buffer_get_poll_fd(struct iio_buffer *buffer)
 	return iio_device_get_poll_fd(buffer->dev);
 }
 
+int iio_buffer_set_blocking_mode(struct iio_buffer *buffer, bool blocking)
+{
+	return iio_device_set_blocking_mode(buffer->dev, blocking);
+}
+
 ssize_t iio_buffer_refill(struct iio_buffer *buffer)
 {
 	ssize_t read;
--- a/buffer.c
+++ b/buffer.c
@@ -98,6 +98,11 @@ void iio_buffer_destroy(struct iio_buffer *buffer)
 	free(buffer);
 }
 
+int iio_buffer_get_poll_fd(struct iio_buffer *buffer)
+{
+	return iio_device_get_poll_fd(buffer->dev);
+}
+
 ssize_t iio_buffer_refill(struct iio_buffer *buffer)
 {
 	ssize_t read;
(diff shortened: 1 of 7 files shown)


pcercuei:
Hi,

I still have some comments before I merge this. I will embed the comments in your commits, that will be simpler.

Also, please add your Signed-Off-By in the commit messages, and rename the "add iio_device_set_nb_buffers" commit message to reflect the new function name.

Thanks!

rofferom:
Ok, tell me when you have terminated your review and I will update my commits

pcercuei:
Ok, I think I'm done :)

One last comment: please decouple the commits that add API functions to the ones that modify the local backend; e.g. first add iio_buffer_get_poll_fd( ) in one commit, then the backend function in a second one. This makes things easier for me when tracking changes and issues.

Thanks for your hard work!

rofferom:
New version pushed !

About https://github.com/Parrot-Developers/libiio/commit/95a67789042a1cc4a0f3a119a6f811f2c05bdb7f#commitcomment-11926460, I don't understand why it would crash, there can be some free(NULL) calls but it is allowed by free() function. Or did I miss something ?

pcercuei:
You're right, I didn't know that free(NULL) was allowed.

However I noticed real mistakes in the last commit:
 * you don't free pdata->blocks and pdata->addrs in enable_high_speed( ) in the case where an error happens in the rest of the function; they will eventually be freed but it would be better to free them sooner (as they won't be used anyway).
 * you reset pdata->nb_blocks to 0. Then, enable_high_speed( ) will exit with an error code, which will make local_open( ) revert to the low-speed interface, and set a buffer size of 0. You should not reset pdata->nb_blocks at all.
 * You free pdata->blocks and pdata->addrs in local_shutdown( ), and not in local_close( ), so you will leak memory if you open/close the device multiple times in one session.

rofferom:
You're right, I should have been more cautious with my commits... Here is a new version that fixes what you have mentioned

pcercuei:
There is a "return -errno" in local.c:712 that should jump to the end of the function, so that pdata->blocks and pdata->addrs are freed. The rest looks good to me.

Almost there :)

rofferom:
Last version seems to build, but push artifact operation has failed : *Error pushing artifact: The operation has timed out*

pcercuei:
The appveyor build failed but it's because appveyor is crap :)

I will do one last round of reviews in the next couple of days, to make sure that it doesn't break anything - and I'll merge it by the end of the week.

Thanks!

rofferom:
Ok, thank you !

pcercuei:
Merged. Thanks!

PR #18, opened 2015-07-03
Local backend: Handle channels declared with a extend_name

rofferom:
Here is a quick fix to handle channels declared with an "extend_name".

IIO drivers can declare a channel with a predefined modifier, or define
a custom string (called "extend_name" in IIO framework). libiio only handle a subset of predefined modifier. I want to use a driver that uses extend_name, so this check is too restrictive for me.

Note that this commit does not handle attributes defined with a modifier and an
extend_name.

Linux sources references to know how devices attribute names are generated : 
https://github.com/torvalds/linux/blob/master/include/linux/iio/iio.h#L226
https://github.com/torvalds/linux/blob/master/drivers/iio/industrialio-core.c#L578

### DIFF
--- a/iio-private.h
+++ b/iio-private.h
@@ -40,23 +40,6 @@
 #define CLEAR_BIT(addr, bit) \
 	*(((uint32_t *) addr) + BIT_WORD(bit)) &= ~BIT_MASK(bit)
 
-enum iio_modifier {
-	IIO_NO_MOD,
-	IIO_MOD_X,
-	IIO_MOD_Y,
-	IIO_MOD_Z,
-	IIO_MOD_LIGHT_BOTH,
-	IIO_MOD_LIGHT_IR,
-	IIO_MOD_ROOT_SUM_SQUARED_X_Y,
-	IIO_MOD_SUM_SQUARED_X_Y_Z,
-	IIO_MOD_LIGHT_CLEAR,
-	IIO_MOD_LIGHT_RED,
-	IIO_MOD_LIGHT_GREEN,
-	IIO_MOD_LIGHT_BLUE,
-	IIO_MOD_I,
-	IIO_MOD_Q,
-};
-
 struct iio_backend_ops {
 	struct iio_context * (*clone)(const struct iio_context *ctx);
 	ssize_t (*read)(const struct iio_device *dev, void *dst, size_t len,
--- a/local.c
+++ b/local.c
@@ -92,45 +92,24 @@ static const char * const device_attrs_blacklist[] = {
 	"uevent",
 };
 
-static const char * const modifier_names[] = {
-	[IIO_MOD_X] = "x",
-	[IIO_MOD_Y] = "y",
-	[IIO_MOD_Z] = "z",
-	[IIO_MOD_ROOT_SUM_SQUARED_X_Y] = "sqrt(x^2+y^2)",
-	[IIO_MOD_SUM_SQUARED_X_Y_Z] = "x^2+y^2+z^2",
-	[IIO_MOD_LIGHT_BOTH] = "both",
-	[IIO_MOD_LIGHT_IR] = "ir",
-	[IIO_MOD_LIGHT_CLEAR] = "clear",
-	[IIO_MOD_LIGHT_RED] = "red",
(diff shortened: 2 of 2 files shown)


lclausen-adi:
Can you give a specific example of a channel that is currently parsed incorrectly and what you'd expect it to be parsed like instead?

rofferom:
I have a device with these kind of files in scan_elements : in_voltage0_**sm1**_* (en/index/type).I expect *voltage0_sm1* to be recognize as a channel. Current libiio implementation only recognize a channel *voltage0* with sm1_en/sm1_index/type attributes.

If I have many channels of type voltage (or any other type) with differents extend_name (instead of different modifiers), libiio will only create on voltage channel that does not exist in driver, with wrong attributes.

To give a more explicit example, in the *scan_elements* folder of my device, I have :
```
in_voltage0_sm1_en in_voltage0_sm1_type in_voltage0_sm1_index
in_voltage0_sm2_en in_voltage0_sm2_type in_voltage0_sm2_index
```

Here is a part of the XML generated by local backend, without the patch :
```xml
<channel id="voltage0" type="input">
  <scan-element index="-2" format="le:u0/0>>0"/>
  <attribute name="sm1_en" filename="scan_elements/in_voltage0_sm1_en"/>
  <attribute name="sm1_index" filename="scan_elements/in_voltage0_sm1_index"/>
  <attribute name="sm1_type" filename="scan_elements/in_voltage0_sm1_type"/>
  <attribute name="sm2_en" filename="scan_elements/in_voltage0_sm2_en"/>
  <attribute name="sm2_index" filename="scan_elements/in_voltage0_sm2_index"/>
  <attribute name="sm2_type" filename="scan_elements/in_voltage0_sm2_type"/>
   ...
</channel>
```

And an XML generated with my patch :
```xml
<channel id="voltage0_sm1" type="input">
  <scan-element index="0" format="be:U16/16>>0"/>
  <attribute name="en" filename="scan_elements/in_voltage0_sm1_en"/>
  <attribute name="type" filename="scan_elements/in_voltage0_sm1_type"/>
  <attribute name="index" filename="scan_elements/in_voltage0_sm1_index"/>
</channel>
<channel id="voltage0_sm2" type="input">
  <scan-element index="0" format="be:U16/16>>0"/>
  <attribute name="en" filename="scan_elements/in_voltage0_sm2_en"/>
  <attribute name="type" filename="scan_elements/in_voltage0_sm2_type"/>
  <attribute name="index" filename="scan_elements/in_voltage0_sm2_index"/>
</channel>
```

This is comparable to an other device with predefined modifiers :
```
in_accel_x_en in_accel_x_index in_accel_x_type
in_accel_y_en in_accel_y_index in_accel_y_type
in_accel_z_en in_accel_z_index in_accel_z_type
```
but in this case libiio will recognize the channels *accel_x*, *accel_y* and *accel_z*

pcercuei:
Hi,

I won't merge this pull request, as it will break everything. The following filename: "in_voltage0_sample_rate" would be detected as the channel "in_voltage0_sample" with a parameter named "rate"...

However, check out the master branch. I think I fixed your problem.

-Paul

rofferom:
I have tested your commit, but unfortunately it doesn't fix my issue. I still have the same wierd channel.

Do you have a suggestion of an alternative strategy to handle my case without beaking everything ? I'm not really aware of all the cases to handle, but if you have an idea to improve my patch, I can provide a new version.

pcercuei:
Actually the original XML generated without your patch is the expected output.

Your problem is a kernel issue: the spec forbids you to have two channels of the same type with the same ID, even if they have different extend_name.

This is valid:
*in_voltage0_sm1_en* *in_voltage1_sm2_en*

This is **not** valid:
*in_voltage0_sm1_en* *in_voltage0_sm2_en*

If you fix the second channel ID in the kernel, then libiio should detect correctly the two channels: one "in_voltage0" channel with the name "sm1" and one "in_voltage1" channel with the name "sm2".

rofferom:
I didn't know these requirements about channel naming. I'm going to update the names in the driver. With you last commit it seems to work as expected.

Thanks for the information !

PR #19, opened 2015-07-22
sort the channels array on device or trigger init

beOn:
related to https://ez.analog.com/message/209741#209741

Signed-off-by: Ben Acland <benacland@gmail.com>

PR #20, opened 2015-07-22
use a dictionary for channels

beOn:
an alternative to https://github.com/analogdevicesinc/libiio/pull/19

related to https://ez.analog.com/message/209741#209741


pcercuei:
I think #19 makes more sense, for several reasons:
* the channel ID alone is not an identifier; the couple (id, output) is;
* a dictionary would not allow you to address a channel by its name (if any), only by its ID;
* it would break the existing Python based applications.

beOn:
Cool. An alternative might be making the channels available by name as attributes of the device object. That wouldn't mess with the existing apps...

But I'll leave off. Long live #19.

pcercuei:
I will close this one then, as I merged #19.

PR #22, opened 2015-12-07
iio_readdev: fix ABI usage for sampling frequency with trigger devices

mtitinger:
be in agreement with Documentation/ABI/testing/sysfs-bus-iio#L37 and use
'sampling_frequency' rather than 'frequency'

Signed-off-by: Marc Titinger <mtitinger@baylibre.com>

### DIFF
--- a/tests/iio_readdev.c
+++ b/tests/iio_readdev.c
@@ -25,6 +25,7 @@
 #define MY_NAME "iio_readdev"
 
 #define SAMPLES_PER_READ 256
+#define DEFAULT_FREQ_HZ  100
 
 static const struct option options[] = {
 	  {"help", no_argument, 0, 'h'},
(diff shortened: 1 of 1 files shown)


pcercuei:
Thanks! I will merge it as soon as I can test. Because I remember it worked at some point, so I just want to make sure it's not a problem in the documentation. Or did the ABI change?

mtitinger:
Hi Paul,

yes it is likely to have changed, I found the following related patch: https://lkml.org/lkml/2015/3/12/136

This is OOT, but I have a colleague who made a prototype of sigrock support for iio, and he is wondering what is best to support multiple devices. Do you confirm that as of today iiod will use the same socket/port for all devices ? I'm curious have you already experimented with a multi port version of iiod (on port per device leading to a different socket fd at client/libiio side)

Thanks and Regards,
Marc.

pcercuei:
Hi,

Ok, so they really broke the ABI. Nice.
Maybe it would make sense to try writing "sampling_frequency" first, and fallback to "frequency" if it returns -ENOENT?

For your other question - Sigrock support for IIO, now that's an awesome news.
Libiio already supports handling multiple devices at the same time, in one thread context or in different threads. One socket is allocated per IIO context, plus one per iio_buffer created.

mtitinger:
Yes the retry with "frequency" is sensible since it worked at some point and you want to remain compatible.

lclausen-adi:
To shed some light on this. The driver were the attribute name was changed is the RTC trigger driver. I don't think I have ever seen anybody using that driver and it's probably going to be removed from upstream at some point since there now is a better replacement.

But the frequency attribute is also used by the hrtimer-trigger that is in the ADI tree. Which is not ABI compliant, but it is out there so we are going to keep supporting that for the moment. The hrtimer trigger that was merged in upstream recently using sampling_frequency. So once the upstream hrtimer trigger gets merged into the ADI tree we'll maybe wait a year or so to give everybody a chance to upgrade and then drop support for the non-standard frequency attribute.

PR #26, opened 2016-03-07
Support Android builds

gburca:
Added `Android.mk` file so that libiio can be built on Android.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -6,6 +6,10 @@ LOCAL_MODULE := libiio
 LOCAL_SRC_FILES := channel.c device.c context.c buffer.c utilities.c
 LOCAL_CFLAGS += -Wall -Wextra
 
+# LOG_LEVEL 0=off, 1=error, 2=warning, 3=info, 4=debug
+LOCAL_CFLAGS += -DLOG_LEVEL=4 -DLOG_TAG=\"libiio\"
+LOCAL_SHARED_LIBRARIES += liblog
+
 LOCAL_CFLAGS += -fvisibility=hidden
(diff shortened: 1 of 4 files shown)


pcercuei:
Pretty cool, thanks!

Two things:
* Could you add your signed-off-by line?
* Android.mk:26 refers to an "external" directory, is that supposed to be present in the source tree?

Thanks,
-Paul

gburca:
`external/libxml2` is part of AOSP, however I haven't tested the network backend, so that's why it's not compiled in by default.

I don't think the signed-off-by line can be added without rewriting the git history. How does github handle merge requests whose SHA1 suddenly disappears?

lclausen-adi:
Just send a new pull-request with the signed-off added.

rgetz:
Does this want to get merged in before 0.9? I think I can use travis to build test things going forward.

pcercuei:
So I had comments on the commits that got dropped when the branch was force-pushed.
Mainly it was for the documentation tweaks to be done in a separate commit since they are unrelated.

This PR was done in March 2016, it's safe to assume that it won't work anymore with the current libiio master, so no need to rush it for the 0.9 release.

adisuciu:
https://github.com/analogdevicesinc/libiio/pull/728 adds support for Android using NDK and CMake https://developer.android.com/ndk/guides/cmake. 

This PR proposes another build method for Android, indepenedent from CMake. It's better if we stick to a single build method for all platforms.

I think we can close this.

pcercuei:
Closing then.

PR #30, opened 2016-06-09
IIOD: Link pthread with no network backend

mfornero:
If iiod is built without the network backend, CMake will not link in
pthreads. Update the CMake configuration to accomodate this case.

Signed-off-by: Matthew Fornero <mfornero@mathworks.com>


mfornero:
This patch was also submitted to buildroot against v0.6, to resolve build failures when building iiod without the network backend.

See [buildroot mailing list](http://lists.busybox.net/pipermail/buildroot/2016-June/163362.html) [thread includes bump to v0.6 and fix for iiod linking]

pcercuei:
Makes sense. Thanks for your contribution!

Issue #28, opened 2016-05-09
Make /sbin/ install directory configurable


anatol:
libiio should not hardcode iiod install path to sbin/

Instead it should use Cmake configuration variables https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

PR #38, opened 2016-08-26
cmake: use GNUInstallDirs module to initialize standard install paths

radhermit:
This lets users override the install location in an easier fashion, for
example:

mkdir build && cd build && cmake CMAKE_INSTALL_PREFIX=/usr/local ..


pcercuei:
Hi Tim,

I don't really understand the purpose of this commit.
For starters, GNUInstallDirs is already included line 14. Also, this commit changes something important, is that the /usr directory is set as the default prefix.
Finally, you can already override the install location using the exact same command that you pasted (well, with a '-D' before CMAKE_INSTALL_PREFIX).

radhermit:
Oh sorry, I've been carrying that patch in my local repo for more than a year now and just noticed someone else suggested the same thing in #28 which you fixed last May... closing.

PR #40, opened 2016-09-14
[RFC PATCH v2] iio: repeat channel support and dummy client example

lucasrangit:

---

Changes since v1:

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -236,7 +236,8 @@ ssize_t iio_buffer_foreach_sample(struct iio_buffer *buffer,
 					processed += ret;
 			}
 
-			ptr += length;
+			ptr += length *
+				(chn->format.repeat ? chn->format.repeat : 1);
 		}
 	}
 	return processed;
@@ -259,7 +260,8 @@ void * iio_buffer_first(const struct iio_buffer *buffer,
 
 	for (i = 0; i < buffer->dev->nb_channels; i++) {
 		struct iio_channel *cur = buffer->dev->channels[i];
-		len = cur->format.length / 8;
+		len = cur->format.length / 8 *
+			(cur->format.repeat ? cur->format.repeat : 1);
 
 		/* NOTE: dev->channels are ordered by index */
 		if (cur->index < 0 || cur->index == chn->index)
--- a/channel.c
+++ b/channel.c
@@ -184,14 +184,17 @@ static char *get_attr_xml(struct iio_channel_attr *attr, size_t *length)
 
 static char * get_scan_element(const struct iio_channel *chn, size_t *length)
 {
-	char buf[1024], *str;
+	char buf[1024], repeat[8] = "", *str;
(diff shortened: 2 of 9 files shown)


pcercuei:
I have one problem with this, adding a 'repeat' field in the iio_data_format will break the ABI :(

lucasrangit:
That is a significant issue. What about if I move the repeat into `struct iio_channel` which is in `iio-private.h`?

lucasrangit:
@pcercuei see https://github.com/analogdevicesinc/libiio/pull/41 for an implementation that moves `repeat` from the public `struct iio_data_format` to the private `struct iio_channel`. If this doesn't break the ABI and you prefer this approach, please close this pull request.

Thanks for the feedback guys!

pcercuei:
So if you move 'repeat' at the end of the iio_data_format structure, applications compiled with an old version of libiio and running with the latest one would still work fine. On the other end, the contrary would not be true. I wonder if that's acceptable? @larsclausen would probably put his veto on this.

lucasrangit:
@pcercuei excellent idea! I like that approach because maintaining forward compatibility is very valuable and future tools can add a version check around using a new {{libiio}} feature. Thanks for the input.

lclausen-adi:
I think it is OK since we never write to a struct that has not been allocated inside the libiio itself. Which means we can't accidentally do a out-of-bounds write on a struct allocated by an application that assumes the old layout.

The only issue I can think of is if an application compiled against a version of libiio that has the field links against a version of libiio that does not have the field. In this case the app would perform a out-of-bounds read when trying to access the repeat field. But the risk for this should be fairly minimal.

rgetz:
Shouldn't most applications be doing a version check anyway? (Which I know you can't guarantee).

lclausen-adi:
We try to be backward and forward compatible, since anything else is a huge inconvenience for application developers and users.

This means applications compiled against an older version will work fine when dynamically linked against a newer version. It also means applications compiled against a newer version will work fine when dynamically linked against an older version as long as they are no using any features that are only available in the newer version. If the application is using newer features the dynamic linker will generate a symbol error. Now if we add the repeat field to the struct an app using the field but dynamically linked against an older version will not get an error (neither during linking, nor at runtime), but rather just read bogus values which might result in undefined behavior.

lclausen-adi:
But sure an app can use iio_library_get_version() to ensure that repeat is really available. We should document this that it is recommended to do so.

lucasrangit:
@lclausen-adi those are great points. thank you.

So if I understand correctly, we can keep the `repeat` field in `struct iio_data_format` but should move it to the end to minimize the impact of struct field offset changes in previously compiled code.

I can will add the ` iio_library_get_version()` to the example code and documentation for review as well.

lclausen-adi:
Yes, sounds about right. Thanks for the good work.

lucasrangit:
I've incorporated all your valuable feedback and created a v4 of this series at https://github.com/analogdevicesinc/libiio/pull/42 .

PR #42, opened 2016-09-16
[RFC PATCH v4] iio: repeat channel support and dummy client example

lucasrangit:

---

Changes since v1:

Changes since v2:
 * iio: moved `struct iio_data_format` member `repeat` to the end of the struct to maintain forward ABI compatibility.
 * iio: added a note specifying that the member was added in version v0.8 (assuming this will be merged). this shows up in the doxygen documentation where users are more likely to see a potential version dependency.
 * docs: added note to main page about abi compatibility and recommended usage of `iio_library_get_version()`.
 * examples: dummy: even though this is versioned with the latest source, it may be run against older library versions so it includes an example of checking the library for a feature and conditionally using that feature if supported.

### DIFF
--- a/examples/dummy-iiostream.c
+++ b/examples/dummy-iiostream.c
@@ -1,0 +1,335 @@
+/*
+ * libiio - Dummy IIO streaming example
+ *
+ * This example libiio program is meant to exercise the features of IIO present
+ * in the sample dummy IIO device. For buffered access it relies on the hrtimer
+ * trigger but could be modified to use the sysfs trigger. No hardware should
+ * be required to run this program.
+ *
+ * How to setup the sample IIO dummy device and hrtimer trigger:
+ *
+ *   1. sudo modprobe industrialio kfifo_buf industrialio-sw-trigger
+ *   2. sudo modprobe iio_dummy iio-trig-hrtimer
+ *   3. sudo mkdir /configfs
+ *   4. sudo mount -t configfs none /config
+ *   5. sudo mkdir /config/iio/triggers/hrtimer/instance1
+ *
+ * Copyright (c) 2016, DAQRI. All rights reserved.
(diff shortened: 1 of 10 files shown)


iio.h
```diff
@@ -1232,6 +1233,9 @@ struct iio_data_format {
 
 	/** @brief Contains the scale to apply if with_scale is set */
 	double scale;
+
+	/** @brief Number of times length repeats (added in v0.8) */
```
lucasrangit: I am assuming this will be merged and released with v0.8.

examples/dummy-iiostream.c
```diff
@@ -0,0 +1,335 @@
...
+	// Listen to ctrl+c and assert
+	signal(SIGINT, handle_sig);
+
+	unsigned int major, minor;
+	char git_tag[8];
+	iio_library_get_version(&major, &minor, git_tag);
+	printf("Library version: %u.%u (git tag: %s)\n", major, minor, git_tag);
+
+	/* check for struct iio_data_format.repeat support */
+	has_repeat = major >= 0 && minor >= 8 ? true : false;
```
lucasrangit: I am assuming this will be merged and released with v0.8.

pcercuei:
I don't like how you use "(chn->format.repeat ? chn->format.repeat : 1)" everywhere.
It does not make sense to have chn->format.repeat set to 0 anyway, why not initialize it to 1 by default?

lucasrangit:
I did that so that the libiio repeat field matched the value in the kernel,
which doesn't initialize to 1. However, I see your point and will change
it.

Thanks for the feedback!

lucasrangit:
Changed the initial value of repeat in pull request https://github.com/analogdevicesinc/libiio/pull/43.

PR #43, opened 2016-09-20
[RFC PATCH v5] iio: repeat channel support and dummy client example

lucasrangit:

---

Changes since v1:

Changes since v2:
 * iio: moved `struct iio_data_format` member `repeat` to the end of the struct to maintain forward ABI compatibility.
 * iio: added a note specifying that the member was added in version v0.8 (assuming this will be merged). this shows up in the doxygen documentation where users are more likely to see a potential version dependency.
 * docs: added note to main page about abi compatibility and recommended usage of `iio_library_get_version()`.
 * examples: dummy: even though this is versioned with the latest source, it may be run against older library versions so it includes an example of checking the library for a feature and conditionally using that feature if supported.

Changes since v4:
 * iio: initialize repeat to 1 to simplify computations. a repeat of 1 is effectively no repeat.
 * examples: dummy: repeat is now 1 so code can be simplified

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -236,7 +236,7 @@ ssize_t iio_buffer_foreach_sample(struct iio_buffer *buffer,
 					processed += ret;
 			}
 
-			ptr += length;
+			ptr += length * chn->format.repeat;
 		}
 	}
 	return processed;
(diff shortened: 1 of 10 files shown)


pcercuei:
Merged. Thanks!

lucasrangit:
Thanks @pcercuei ! Sorry for not checking the Windows build. Wont happen again.

Looks like my commit 356420d6806e175f37144991db973fbdc9ccfb77 snuck in as the 1000th commit! I hope this means I'll be invited to the libiio party ;)

pcercuei:
I'm sure you did set the commit date on purpose ;)

PR #44, opened 2016-09-29
iio.h: add checks for MATLAB workarounds

radhermit:
MATLAB requires a few workarounds for its header parsing, library
loading, and mex usage to work properly.

MATLAB_MEX_FILE is defined via the MATLAB mex build process while
MATLAB_LOADLIBRARY is defined via a header wrapper used by the libiio
MATLAB bindings.

### DIFF
--- a/bindings/matlab/CMakeLists.txt
+++ b/bindings/matlab/CMakeLists.txt
@@ -1,0 +1,18 @@
+find_program(
+	MATLAB_EXECUTABLE
+	NAMES matlab
+	DOC "Matlab main program"
+)
+mark_as_advanced(MATLAB_EXECUTABLE)
+
+if (MATLAB_EXECUTABLE)
+	option(MATLAB_BINDINGS "Install MATLAB bindings" ON)
+
+	if (MATLAB_BINDINGS)
+		install(
+			DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+			DESTINATION "${CMAKE_INSTALL_PREFIX}/share/libiio"
+			PATTERN "CMakeLists.txt" EXCLUDE
+		)
+	endif()
+endif()
--- a/bindings/matlab/CMakeLists.txt
+++ b/bindings/matlab/CMakeLists.txt
@@ -14,5 +14,10 @@ if (MATLAB_EXECUTABLE)
 			DESTINATION "${CMAKE_INSTALL_PREFIX}/share/libiio"
 			PATTERN "CMakeLists.txt" EXCLUDE
 		)
+		install(
+			CODE "execute_process(
+				COMMAND ${MATLAB_EXECUTABLE} -nodesktop
+					-nodisplay -r \"cd('${CMAKE_INSTALL_PREFIX}/share/libiio/matlab');iio_installer_script;exit;\"
+				OUTPUT_QUIET)")
 	endif()
 endif()
--- a/bindings/matlab/libiio_if.m
+++ b/bindings/matlab/libiio_if.m
@@ -1,0 +1,697 @@
+% Copyright 2014(c) Analog Devices, Inc.
+%
+%  All rights reserved.
+%
+%  Redistribution and use in source and binary forms, with or without modification,
+%  are permitted provided that the following conditions are met:
+%      - Redistributions of source code must retain the above copyright
+%        notice, this list of conditions and the following disclaimer.
+%      - Redistributions in binary form must reproduce the above copyright
+%        notice, this list of conditions and the following disclaimer in
+%        the documentation and/or other materials provided with the
+%        distribution.
+%      - Neither the name of Analog Devices, Inc. nor the names of its
+%        contributors may be used to endorse or promote products derived
+%        from this software without specific prior written permission.
+%      - The use of this software may or may not infringe the patent rights
+%        of one or more patent holders.  This license does not release you
+%        from the requirement that you obtain separate licenses from these
+%        patent holders to use this software.
+%      - Use of the software either in source or binary form or filter designs
+%        resulting from the use of this software, must be connected to, run
+%        on or loaded to an Analog Devices Inc. component.
+%
+%  THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+%  INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT, MERCHANTABILITY AND FITNESS FOR A
+%  PARTICULAR PURPOSE ARE DISCLAIMED.
+%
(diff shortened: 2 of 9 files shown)


radhermit:
This should allow us to quit using a partially munged iio.h file for the MATLAB bindings and instead just copy the file as it is (or even install the header alongside the library on all targets and use that). Using this with the current MATLAB bindings requires using a simple wrapper header to define a macro for loadlibrary() usage and using the addheader param in the loadlibrary() call to load the actual iio.h header.

One warning that MATLAB currently throws when connecting to remote boards after those changes relates to the iio_context_info struct being defined in iio-private.h, the warning output is as follows:

```
Warning: The data type 'iio_context_infoPtrPtrPtr' used by
function iio_scan_context_get_info_list does not exist.
> In loadlibrary
  In libiio_if/init (line 401)
  In AD9361_Filter_Wizard>connect2target_Callback (line 2628)
  In gui_mainfcn (line 95)
  In AD9361_Filter_Wizard (line 96)
  In matlab.graphics.internal.figfile.FigFile/read>@(hObject,eventdata)AD9361_Filter_Wizard('connect2target_Callback',hObject,eventdata,guidata(hObject))
  In uiwait (line 81)
  In AD9361_Filter_Wizard>AD9361_Filter_Wizard_OpeningFcn (line 255)
  In gui_mainfcn (line 220)
  In AD9361_Filter_Wizard (line 96)
```

pcercuei:
Could you just add your signed-off-by to the commit?
Thanks!

radhermit:
Yep, done.

rgetz:
Does the warning need to get resolved? (It seems pretty noisy, and happens ever time someone accesses a model inside MATLAB/Simulink).

@pcercuei ?

-Robin

pcercuei:
Yes, it's annoying if it adds warnings...

radhermit:
One method of avoiding that warning without touching the libiio headers might be to pre-generate and edit the prototype interface headers that matlab uses, but I haven't figured out how to make matlab happier yet using those.

mfornero:
If the warning is expected and not "something bad", you could suppress the warning before calling loadlibrary and then restore it immediately afterward. 

See the help on the "warning" function for details on how to do this (and the lastwarn function to get the warning ID)

radhermit:
Yes, that's a decent workaround as well which I've implemented. Also, this now pulls the matlab bindings into this repo and includes initial cmake support to install them at $PREFIX/share/libiio/matlab.

Replacing the bindings in the filter wizard with a well-placed addpath() makes things work as expected when the bindings are installed to the system.

I think the only thing missing now is adding installer support for Windows targets.

bindings/matlab/libiio_if.m
```diff
@@ -398,7 +398,10 @@ function delete(obj)
             % Load the libiio library
             if(~libisloaded(obj.libname))
                 try
+                    % ignore unknown type warnings due to header parsing limitations
+                    warning('off', 'MATLAB:loadlibrary:TypeNotFound');
                     [notfound, warnings] = loadlibrary(obj.libname, obj.hname, 'addheader', 'iio.h');
+                    warning('on', 'MATLAB:loadlibrary:TypeNotFound');
                 catch exception
```
mfornero: Right, the best practice here would be:
```matlab
try
    % ignore unknown type warnings due to header parsing limitations
    warnState = warning('off', 'MATLAB:loadlibrary:TypeNotFound');
    cleanupObj = onCleanup(@()warning(warnState));
    [notfound, warnings] = loadlibrary(obj.libname, obj.hname, 'addheader', 'iio.h');
    cleanupObj = []; % restore the warning state
```

The onCleanup() function returns an object that will execute the specified function when it is cleared. We explicitly clear it after loading the library, but it will be implicitly cleared when you return from the function in the catch block.

bindings/matlab/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+find_program(
+	MATLAB_EXECUTABLE
+	NAMES matlab
+	PATHS ${Matlab_ROOT_DIR} ${Matlab_ROOT_DIR}/bin
```
radhermit: I was mostly copying straight from the FindMatlab module available in newer cmake versions so its purpose is if a user has multiple matlab versions installed (or has it installed in a non-standard directory) and wants to point specifically at it by setting that variable.

If it's not specified, whatever version is the default or found first is used.

radhermit: I should note we could bundle the FindMatlab module and just use that, but I don't think we need all the extra functionality that provides at this point.

radhermit:
Another idea of where to install this would be somewhere under the targeted MATLAB root dir, perhaps somewhere like $MATLAB_ROOT/toolbox/iio? Then the install process could even automate updating MATLAB's search path for the targeted version so addpath() wouldn't have to be used anywhere in consumers.

mfornero:
I'd generally recommend against installing to MATLAB_ROOT-- this may be on a network drive for some users, and is often not writable

When building add-ons with standalone installers, we generally have the installer (or the make install action) call MATLAB in command line mode, and execute a script that takes care of adding the relevant paths via addpath() and savepath()

See 
https://www.mathworks.com/help/releases/R2016b/matlab/ref/matlabwindows.html for Windows options (and links to Linux / Mac). Of note, Windows will likely need to use the -wait option

I generally do something along the lines of:
```matlab
matlab -r cd(<add-on dir>);InstallerScript;exit;
```

By cd'ing into the add-on dir, you can ensure your InstallerScript is visible (this also deals with spaces in the path). The installer script itself can make use of mfilename() to determine where it is installed, and generate the directories for addpath(). E.g.

```matlab
function InstallerScript(varargin)
    if nargin > 0
        install = varargin{1}; % use the command line arguement
    else
        install = true; % assume install
    end
    thisDir = fileparts(mfilename('fullpath')); % path to InstallerScript.m
    
    if install
        pathfunc = @addpath; % add paths for installation
    else
        pathfunc = @rmpath; % remove paths for uninstall
    end
    
    pathfunc(thisDir);
    pathfunc(fullfile(thisDir,'include'));
    savepath;
end
```

bindings/matlab/CMakeLists.txt
radhermit: Yeah, I realized that which is why I mentioned we could just bundle the module instead of me copying more of its setup over, but I guess removing the line is fine for now.

radhermit:
Thanks for the example, using it now and things seem to work for Linux. However, this does assume that the default pathdef.m file is writable and the installing user matches the user registered for MATLAB usage.

If the users don't match (e.g. installing with root perms) the registration wizard pops up instead and once you confirm your regular user still exists the process doesn't actually add the path to the search path list.

pcercuei:
Merged. Thanks!

PR #63, opened 2016-11-15
InnoSetup: install matlab bindings via the windows installer

radhermit:
Places the matlab bindings alongside the C# bindings dll in "Common Files\libiio".

The last piece of the puzzle in order for people to install libiio and start using the bindings seamlessly would be to have the installer also run the iio_installer_script.m to add the installed binding directory to the system path. Obviously we'd either want to make that an install option or done automatically if matlab is installed.

For now people will have to run it themselves or manually add the install dir to their syspath for matlab.

### DIFF
--- a/libiio.iss.cmakein
+++ b/libiio.iss.cmakein
@@ -63,3 +63,5 @@ Source: "C:\libs\32\libserialport-0.dll"; DestDir: "{sys}"; Flags: onlyifdoesnte
 Source: "C:\libs\64\libserialport-0.dll"; DestDir: "{sys}"; Check: Is64BitInstallMode; Flags: onlyifdoesntexist
 
 Source: "C:\projects\libiio\build-win32\bindings\csharp\libiio-sharp.dll"; DestDir: "{cf}\libiio"
+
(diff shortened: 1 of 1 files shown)


radhermit:
One note to make, I haven't tried using it installed specifically like this yet so I'm unsure if matlab's default header search path will find where iio.h is currently being placed. I'll try this out tomorrow/later and see if it works.

radhermit:
Right, that's what I was planning to do.

mfornero:
We (Pilots) use Inno Setup to build some of our installers. We generally prompt the user for their MATLAB install path, then test for the presence of the MATLAB executable (e.g. matlab.exe or matlab_helper) to verify that the path given is correct.

We then use the Exec command to run MATLAB with the -r (/r) switch, which can then be used to run an install/uninstall script.

Note that if you do create a MATLAB app, you would not want to install in this way, as you'd be running from within MATLAB.

rgetz:
@mfornero : Do we still need this one? or this should be dropped now?

Thanks
-Robin

mfornero:
I don't think we need this, as the bindings are showing up inside the .zip

rgetz:
OK - closing this then.

rgetz:
@pcercuei or @radhermit - did you want to rm this branch then?

radhermit:
Done.

PR #64, opened 2016-11-15
Usb context info with iserial

mhennerich:


pcercuei:
Maybe you could make it easier to parse?

e.g.
0x456:0xb672 (Analog Devices Inc. M2K), serial=01234567

Then it's easy to search for "serial=" to get the number.

mhennerich:
If Lars agrees - this would be my favorite too. 

-Michael

rgetz:
Or comma, 

"%04x:%04x (%s, %s, %s)" 

but that causes issues if there is a comma in the manufacturer (like "Analog Devices, Inc"), or the other fields... so that isn't going to work.

I guess, "serial=xxx" works for me (as unwieldy string parsing is in C).

-Robin

mhennerich:
Updated the description string per this discussion and updated the feature branch

usb.c
```diff
@@ -953,10 +953,10 @@ static int usb_fill_context_info(struct iio_context_info *info,
 		unsigned int interface)
 {
 	struct libusb_device_descriptor desc;
-	char manufacturer[64], product[64];
+	char manufacturer[64], product[64], serial[64];
 	char uri[sizeof("usb:127.255.255")];
 	char description[sizeof(manufacturer) + sizeof(product) +
-		sizeof("0000:0000 ( )")];
+		sizeof(serial) + sizeof("0000:0000 ( )")];
```
pcercuei: it should be sizeof("0000:0000 ( ), serial=")
Otherwise the buffer is too small.

mhennerich:
Thanks for pointing out - fixed

pcercuei:
Pushed to master, thanks.

PR #65, opened 2016-11-25
Merge : Travis ci

rgetz:
This adds a few badges to the README, and updates the CMake and travis files for building and deploying things to tags, and coverity (for master only).


rgetz:
It looks like appveyor doesn't like things yet -- specifically the minor number being set in the Cmake (I think), so closing this - until I get that resolved.

PR #71, opened 2016-11-30
iio_buffer: fix samples alignment computation

rlefevre:
Samples are aligned to their own size, but the alignment is
from the buffer start, not an absolute (virtual) memory alignment.

For example, timestamps are added like this in the kernel:

    size_t ts_offset = indio_dev->scan_bytes / sizeof(int64_t) - 1;
    ((int64_t *)data)[ts_offset] = timestamp;

Also:

```
  static int iio_compute_scan_bytes(struct iio_dev *indio_dev,
                                const unsigned long *mask, bool timestamp)
  {
        const struct iio_chan_spec *ch;
        unsigned bytes = 0;
        int length, i;

        /* How much space will the demuxed element take? */
        for_each_set_bit(i, mask, indio_dev->masklength) {
                ...
                bytes = ALIGN(bytes, length);
                        ^^^^^^^^^^^^^^^^^^^^
                bytes += length;
  }
```

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -201,6 +201,7 @@ ssize_t iio_buffer_foreach_sample(struct iio_buffer *buffer,
 			void *, size_t, void *), void *d)
 {
 	uintptr_t ptr = (uintptr_t) buffer->buffer,
+		  start = ptr,
 		  end = ptr + buffer->data_length;
 	const struct iio_device *dev = buffer->dev;
 	ssize_t processed = 0;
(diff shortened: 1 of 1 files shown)


lclausen-adi:
Hi,

Thanks for the patch, this makes sense, well spotted. Can you add a Signed-off-by tag to the commit message.

Thanks,
- Lars

rlefevre:
Hi, 

Yes, I had issues with alignment when using custom channels of large size using repeat syntax (for which I may provide another patch later). I added the Signed-off-by.

Thanks
Rémi

lclausen-adi:
Ok, makes sense. I was wondering when this would occur since the pointer itself should always be at least aligned to 8 bytes. But with larger repeat values this could cause issue.

rlefevre:
Well, I think that a kernel channel storagebits value of 128 is valid (power of 2 and on 8 bits), so with a 8 bytes malloc alignment, an issue could occur with such a channel of 16 bytes (assuming the iio_buffer->buffer is not aligned on 16 bytes and there are other channels after this one), even without a repeat value.

rlefevre:
PS: I could not find any other places where this has some impact, but I may not be versed enough with the code to spot them, do you think of other parts that could be impacted?

rlefevre:
Hi, anything needed to have this patch going forward?

Thx

pcercuei:
Looks good to me. Let's see if it breaks ✌️

PR #74, opened 2016-12-15
matlab: Add scan_block helper API

mfornero:
Add scan_block helper functions for use in languages (e.g. MATLAB) that
cannot easily iterate over a list of pointers.

The scan_block functions enable the creation / scanning of a scan context
and the retrival of context_info via an index value.

Signed-off-by: Matt Fornero <matt.fornero@mathworks.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,7 +294,14 @@ set(CPACK_PACKAGE_CONTACT "Engineerzone <https://ez.analog.com/community/linux-d
 set(CPACK_DEBIAN_PACKAGE_DEPENDS "libaio-dev (>= 0.3.109-4), libavahi-client-dev (>= 0.6.31-4), libavahi-common-dev (>= 0.6.31-4), libc6-dev (>= 2.19-0), libusb-1.0-0-dev (>= 2:1.0.17-1), libxml2-dev (>= 2.9.1+dfsg1-3)")
 include(CPack)
 
-add_library(iio ${LIBIIO_CFILES} ${LIBIIO_HEADERS})
+add_subdirectory(bindings)
+
+if (WITH_MATLAB_BINDINGS_API)
+	set(LIBIIO_EXTRA_HEADERS ${LIBIIO_EXTRA_HEADERS} bindings/matlab/iio-wrapper.h)
+	add_definitions(-DMATLAB_BINDINGS_API=1)
+endif()
+
+add_library(iio ${LIBIIO_CFILES} ${LIBIIO_HEADERS} ${LIBIIO_EXTRA_HEADERS})
 set_target_properties(iio PROPERTIES
 	VERSION ${VERSION}
 	SOVERSION ${LIBIIO_VERSION_MAJOR}
@@ -391,7 +398,5 @@ endif()
 if(WITH_TESTS)
 	add_subdirectory(tests)
 endif()
-
-add_subdirectory(bindings)
 
 configure_file(iio-config.h.cmakein ${CMAKE_CURRENT_BINARY_DIR}/iio-config.h @ONLY)
--- a/bindings/matlab/iio-wrapper.h
+++ b/bindings/matlab/iio-wrapper.h
@@ -1,4 +1,46 @@
 #ifndef MATLAB_LOADLIBRARY
 #define MATLAB_LOADLIBRARY
 #include <iio.h>
+
+#ifndef __api
+#define __api
+#endif
+
+struct iio_scan_block;
+
+/** @brief Create a scan block
+* @param backend A NULL-terminated string containing the backend to use for
+* scanning. If NULL, all the available backends are used.
+* @param flags Unused for now. Set to 0.
+* @return on success, a pointer to a iio_scan_block structure
+* @return On failure, NULL is returned and errno is set appropriately */
+__api struct iio_scan_block * iio_create_scan_block(
+		const char *backend, unsigned int flags);
+
+
+/** @brief Destroy the given scan block
+* @param ctx A pointer to an iio_scan_block structure
+*
+* <b>NOTE:</b> After that function, the iio_scan_block pointer shall be invalid. */
+__api void iio_scan_block_destroy(struct iio_scan_block *blk);
+
+
+/** @brief Enumerate available contexts via scan block
+* @param blk A pointer to a iio_scan_block structure.
+* @returns On success, the number of contexts found.
+* @returns On failure, a negative error number.
+*/
+__api ssize_t iio_scan_block_scan(struct iio_scan_block *blk);
+
+
+/** @brief Get the iio_context_info for a particular context
+* @param blk A pointer to an iio_scan_block structure
+* @param index The index corresponding to the context.
+* @return A pointer to the iio_context_info for the context
+* @returns On success, a pointer to the specified iio_context_info
+* @returns On failure, NULL is returned and errno is set appropriately
+*/
+__api struct iio_context_info *iio_scan_block_get_info(
+		struct iio_scan_block *blk, unsigned int index);
+
 #endif
--- a/iio-private.h
+++ b/iio-private.h
@@ -44,6 +44,10 @@
 #   define __api
 #endif
 
+#ifdef WITH_MATLAB_BINDINGS_API
+#include "bindings/matlab/iio-wrapper.h"
+#endif
+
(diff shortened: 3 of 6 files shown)


pcercuei:
Hi Matt,

I don't like the idea of adding 4 functions to the public API that are completely redundant with what's already there. It would be pretty confusing for everybody.

That said, I'm not against having those in the library if the function prototypes are moved to bindings/matlab/iio-wrapper.h and the code moved to (e.g.) scan-matlab.c that would be compiled only if the bindings for Matlab are enabled.

mfornero:
If we move the API calls into bindings/matlab/iio-wrapper.h, we need to duplicate the platform detection and __api definition code to properly export the symbols at compile time.

Please see the proposed update, which adds a CMake option to enable / disable the MATLAB scan API, but leaves the function definitions in iio.h-- please let me know your thoughts. 

Maybe we'd also want to rename the "scan_block" function to "matlab_scan" or something along those lines to make the intended use case clear?

pcercuei:
Hi Matt,
I don't understand why you need to duplicate __api? When compiling for Matlab, it's an empty macro.

pcercuei:
Ok sorry, I spoke without having my morning coffee. The symbols still need to be defined in the library.

However I'd rather still duplicate the platform detection and __api definition, than having the functions in iio.h, just so that the C end-users don't get confused - which is my only concern here.

mfornero:
Hi Paul,

Please see the latest changes-- I believe they now meet the above criteria.

I did have to move a few things around in the top level CMakeLists, and include some MATLAB-API specific logic there-- CMake 3.1 adds the target_sources function that would eliminate the need for this, but it's not readily available on some of the supported platforms. Let me know if you have some other suggestions here.

-Matt

bindings/matlab/iio-wrapper.h
```diff
@@ -1,4 +1,56 @@
...
+
+/** @brief Get the iio_context_info for a particular context
+* @param blk A pointer to an iio_scan_block structure
+* @param index The index corresponding to the context.
+* @return A pointer to the iio_context_info for the context
+* @returns On success, a pointer to the specified iio_context_info
+* @returns On failure, NULL is returned and errno is set appropriately
+*/
+__api struct iio_context_info *iio_scan_block_get_info(
+struct iio_scan_block *blk, unsigned int index);
```
pcercuei: Two tabs alignment

scan.c
```diff
@@ -41,6 +45,26 @@ const char * iio_context_info_get_uri(
 	return info->uri;
 }
 
+#ifdef WITH_MATLAB_BINDINGS_API
```
pcercuei: Could you put those two functions at the bottom of the file? So that they can be wrapped together with iio_create_scan_block() in the same #ifdef WITH_MATLAB_BINDINGS_API.

scan.c
```diff
@@ -158,3 +182,32 @@ void iio_scan_context_destroy(struct iio_scan_context *ctx)
 #endif
 	free(ctx);
 }
+
+#ifdef WITH_MATLAB_BINDINGS_API
+struct iio_scan_block *iio_create_scan_block(
+	const char *backend, unsigned int flags)
```
pcercuei: Two tabs alignment

scan.c
```diff
@@ -41,6 +45,26 @@ const char * iio_context_info_get_uri(
...
+	iio_context_info_list_free(blk->info);
+	blk->info = NULL;
+	blk->ctx_cnt = iio_scan_context_get_info_list(blk->ctx, &blk->info);
+	return blk->ctx_cnt;
+}
+
+struct iio_context_info *iio_scan_block_get_info(
+		struct iio_scan_block *blk, unsigned int index)
+{
+	if (!blk->info || (ssize_t)index >= blk->ctx_cnt){
```
pcercuei: Missing a space before the `{`. Yes I'm picky :)

CMakeLists.txt
```diff
@@ -294,7 +294,14 @@ set(CPACK_PACKAGE_CONTACT "Engineerzone <https://ez.analog.com/community/linux-d
 set(CPACK_DEBIAN_PACKAGE_DEPENDS "libaio-dev (>= 0.3.109-4), libavahi-client-dev (>= 0.6.31-4), libavahi-common-dev (>= 0.6.31-4), libc6-dev (>= 2.19-0), libusb-1.0-0-dev (>= 2:1.0.17-1), libxml2-dev (>= 2.9.1+dfsg1-3)")
 include(CPack)
 
-add_library(iio ${LIBIIO_CFILES} ${LIBIIO_HEADERS})
+add_subdirectory(bindings)
+
+if (WITH_MATLAB_BINDINGS_API)
+	set(LIBIIO_EXTRA_HEADERS ${LIBIIO_EXTRA_HEADERS} bindings/matlab/iio-wrapper.h)
+	add_definitions(-DMATLAB_BINDINGS_API=1)
```
pcercuei: Please indent those two lines with one tab

iio-private.h
```diff
@@ -206,6 +206,12 @@ struct iio_context_info {
 	char *uri;
 };
 
+struct iio_scan_block {
```
pcercuei: You could have this directly in scan.c, I think it won't be used by any other .c file.

pcercuei:
Ok, I just have some comments about cosmetic changes, see above.

Btw, do you want iio-wrapper.h to be installed as well?

mfornero:
Will update with cosmetic fixes.

I believe iio-wrapper.h is included in the .zip package currently, which is what we plan to use. For other uses, iio-wrapper.h should be installed when the system object (the contents of bindings/matlab) are being installed-- which I believe the existing installer takes care of.

-Matt

CMakeLists.txt
mfornero: These already are indented with one tab, unless I'm missing something

mfornero:
Updated per cosmetic fixes above, other than CMakeLists.txt (let me know if I'm missing something there)

CMakeLists.txt
pcercuei: I blame Github's web view then :)

scan.c
```diff
@@ -29,6 +33,14 @@ struct iio_scan_context {
 	bool scan_local;
 };
 
+#ifdef WITH_MATLAB_BINDINGS_API
+struct iio_scan_block {
```
pcercuei: Put 'struct iio_scan_block' right before 'iio_scan_block_scan' in the #ifdef and I'm happy with the commit.

mfornero: Done-- let me know if I'm missing anything else.

bindings/matlab/iio-wrapper.h
```diff
@@ -1,4 +1,56 @@
...
 #include <iio.h>
+
+#ifdef _WIN32
+#   ifdef LIBIIO_EXPORTS
+#	define __api __declspec(dllexport)
+#   else
+#	define __api __declspec(dllimport)
+#   endif
+#elif __GNUC__ >= 4 && !defined(MATLAB_MEX_FILE) && !defined(MATLAB_LOADLIBRARY)
+#   define __api __attribute__((visibility ("default")))
```
pcercuei: Does this work? !defined(MATLAB_LOADLIBRARY) will always be false.

mfornero: I didn't modify that part-- I believe it was added in b20f340421.

It's an odd format (maybe __MATLAB_IIO_WRAPPER__ would be better?), but I believe the ifdefs exist to prevent multiple inclusion. This header should only be included if the MATLAB bindings are being (1) built into the library or (2) called from MATLAB

pcercuei: Actually MATLAB_LOADLIBRARY is for when Matlab reads iio-wrapper.h then iio.h, so that it doesn't choke on __api, since it does not handle GCC's __attribute__ stuff. In that case, __api should not be defined...

Ok, what about this:
* in iio-wrapper.h, add this:
`#ifndef __api  #define __api  #endif`
* Don't include bindings/matlab/iio-wrapper.h from scan.c; but include bindings/matlab/iio-wrapper.h from iio-private.h, after __api is defined.

pcercuei: I made a patch for your patch: https://gist.github.com/pcercuei/e2e4cd23dfa09e4e4aaa9140e05fba1a
With this change it builds fine on Linux and the symbols are available.

pcercuei:
Merged. Thanks!

PR #76, opened 2016-12-19
Local backend: return negative errno code on local_read/write() error

rlefevre:
Since commit 4ebd0743, returned negative errno code is always `-EIO`
on `read()` error instead of using `-errno`.

For example, `-EAGAIN` was not returned anymore in blocking mode when
no data was available.

Also fix documentation regarding `-EAGAIN` from `iio_buffer_refill()` and
`iio_buffer_push()`.

### DIFF
--- a/iio.h
+++ b/iio.h
@@ -1102,7 +1102,7 @@ __api int iio_buffer_get_poll_fd(struct iio_buffer *buf);
 
 /** @brief Make iio_buffer_refill() and iio_buffer_push() blocking or not
  *
- * After this function has been called with blocking == true,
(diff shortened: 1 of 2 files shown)


pcercuei:
I wonder why that commit changed -errno by -EIO. @lclausen-adi, do you remember?

The same fix should be applied to local_write, btw.

rlefevre:
I added the fix for local_write().

lclausen-adi:
I don't think there was a specific reason, this is probably just a mistake.

pcercuei:
Ok, I'll merge it then. Thanks!

PR #77, opened 2016-12-21
Add to API to make runtime query of built library backends possible

oliverbunting:
Currently applications cannot query at runtime which backends are available. I've added some functions to the top level section of the API to query the built configuration at runtime.

If you'd prefer, it'd also be possible to do the same with a single function returning flags: a matter of taste.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,7 @@ if (ENABLE_IPV6)
 	endif()
 endif()
 
-set(LIBIIO_CFILES channel.c device.c context.c buffer.c utilities.c scan.c)
+set(LIBIIO_CFILES channel.c device.c context.c buffer.c utilities.c scan.c iio-config.c)
 set(LIBIIO_HEADERS iio.h)
 
 add_definitions(-D_POSIX_C_SOURCE=200809L -D__XSI_VISIBLE=500 -DLIBIIO_EXPORTS=1)
--- a/iio-config.c
+++ b/iio-config.c
@@ -1,0 +1,66 @@
+
+/*
+ * libiio - Library for interfacing industrial I/O (IIO) devices
+ *
+ * Copyright (C) 2014-2015 Analog Devices, Inc.
+ * Author: Paul Cercueil <paul.cercueil@analog.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * */
+
+#include "iio-config.h"
+#include <stdbool.h>
+
+bool iio_backend_has_local() 
+{
+#ifdef WITH_LOCAL_BACKEND
+	return true;
+#else
+	return false;
+#endif
+}
+
+bool iio_backend_has_xml() 
+{
+#ifdef WITH_XML_BACKEND
+	return true;
+#else
+	return false;
+#endif
+}
+
+bool iio_backend_has_network() 
+{
+#ifdef WITH_NETWORK_BACKEND
+	return true;
+#else
+	return false;
+#endif
+}
+
+bool iio_backend_has_usb() 
+{
+#ifdef WITH_USB_BACKEND
+	return true;
+#else
+	return false;
+#endif
+}
+
+bool iio_backend_has_serial() 
+{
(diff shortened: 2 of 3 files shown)


rgetz:
You can do this today - set up context - see what err is set. That is the way we do it in the iio scope application. 

- Robin

oliverbunting:
OK. For reference, it's not done that way in iio_info.c though, which is what was guiding me.

oliverbunting:
Actually, looking through the API, its not at all obvious how to create a serial or usb context. You could do so with a URI returned from a scan context, but what that scans is hard-coded at compile time.

rgetz:
Yeah, I'm not saying that it isn't a good idea, and we wouldn't use it - just saying it's possible (although it is noisy - it prints an error to the console saying "can't connect").

https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/dialogs.c#L1063

oliverbunting:
Would perhaps the more general solution to be create a version of 

`__api struct iio_context * iio_create_default_context(void);`
[](https://github.com/analogdevicesinc/libiio/blob/master/iio.h#L263)

with the signature 
`__api struct iio_context * iio_create_named_context(const char *);`

and perhaps also change the functions I've submitted for merge with a single function
`__api __pure struct const char ** iio_get_available_backends();`

I'd be happy to do so.

 My motivation by the way is I'm in the middle of providing a Haskell wrapper to the library, and reimplementing iio_info.c faithfully is proving tricky.

pcercuei:
Well, your `iio_create_named_context` is `iio_create_context_from_uri`...

The correct way of creating a USB or serial context is through URI. Either by scanning, or by composing the URI string yourself (which needs some doc).

Instead of one API function per backend to query if it's been compiled in, why not:
`iio_has_backend(const char *name)`?

oliverbunting:
That seems fine. It is perhaps redundant now, as scan_context can provide all uri's.

pcercuei:
Well, not exactly, the scan API will discover the available IIO contexts, but the list may not include URIs for all the supported backends. For instance when the USB backend is compiled in but there's no compatible USB device plugged in.

rgetz:
Paul:

What about get_capabilities() where it returns an enum defined in the header file? That is mostly how ffmpeg works. 

- Robin

pcercuei:
Not an enum but a u32 value containing flags, yes. Olivier proposed that in his first message.

oliverbunting:
I'm happy for me to modify the pull request to this end. I'd be surprised if that happens before new year, I've got a lot on.

lclausen-adi:
We might want to consider having a API that returns a list of strings of the supported backends. The same strings that can be passed to iio_create_scan_context() as the backend parameter.

oliverbunting:
I favour the strings approach personally.

rgetz:
Ok - strings it is, as @larsclausen indicated - strings which can be passed to  iio_create_scan_context(), which are defined/used at:

https://github.com/analogdevicesinc/libiio/blob/master/scan.c#L125
https://github.com/analogdevicesinc/libiio/blob/master/context.c#L270

and a few other places.

Did we want to change the strings into a header file somewhere, so they are always the same (and easier to maintain?) It might make user code easier (was looking at /tests/iio_genxml.c where it defines its own enum).

Something like this in iio.h

enum iio_backend {
	IIO_LOCAL,
	IIO_XML,
	IIO_NETWORK,
        IIO_USB,
        IIO_SERIAL,
};

static const char *IIO_BACKEND_STRING[] = {
    "local",  "xml",  "ip", "usb", "serial"
};

and two functions:

bool iio_has_backend(const char * backend)
which returns true/false for a backend

char * iio_backends() 
which returns all the available back ends.

Then it would "simplify" things to:

if (iio_has_backend(IIO_BACKEND_STRING[IIO_LOCAL]) &&
                 !strcmp(uri, IIO_BACKEND_STRING[IIO_LOCAL]))

or something like that - it would eliminate the ifdef in many things (things should get compiled out since the pre-processor has all the strings.

Is that what we want to do?

-Robin

pcercuei:
There's not really a need for the enum or IIO_BACKEND_STRING, and it might get in the way in the future if we support dynamically loaded backends as .so/DLL modules (like in PR #9).

iio_has_backend() is fine.
char ** iio_backends(), would that work on Matlab? Also, returning an array of strings isn't easy, I'd prefer something like iio_get_backends_count() + iio_get_backend() which would return pointer to a static const string.

rgetz:
char * iio_backends() would work, if it returned a single space delimited string. "local xml ip usb serial".
users would then do something like if (strstr(iio_backends(), "local")) or something like that.

If it returns a char**, it is not going to work in MATLAB's callib.

-Robin

pcercuei:
I made backend-query branch, if you want to test:
https://github.com/analogdevicesinc/libiio/tree/backend-query

rgetz:
This worked for what I need it to do.

@oliverbunting - does this work for you?

I think we wanted to tag a 0.9 tomorrow if you can have a quick look.

-Robin

oliverbunting:
Looks good to me!

Issue #81, opened 2017-01-02
runtime error on ubuntu 15


rgetz:
Same device works on Windows and other Linux installs

picozed@picozed:~$ iio_info -n 192.168.2.1
Library version: 0.8 (git tag: 9838779)
Compiled with backends: local xml network usb
IIO context created with network backend.
Backend version: 0.8 (git tag: v0.8   )
Backend description string: 192.168.2.1 Linux pluto 4.6.0-g1470706-dirty #66 SMP PREEMPT Fri Dec 23 12:40:07 CET 2016 armv7l
IIO context has 1 attributes:
	local,kernel: 4.6.0-g1470706-dirty
IIO context has 5 devices:
	iio:device0: adm1177
... works ....

picozed@picozed:~$ lsusb
Bus 001 Device 002: ID 1bcf:288a Sunplus Innovation Technology Inc. 
Bus 001 Device 006: ID 0456:b673 Analog Devices, Inc. 
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 005 Device 001: ID 1d6b:0001 Linux Foundation 1.1 root hub
Bus 004 Device 001: ID 1d6b:0001 Linux Foundation 1.1 root hub
Bus 003 Device 001: ID 1d6b:0001 Linux Foundation 1.1 root hub
Bus 002 Device 001: ID 1d6b:0001 Linux Foundation 1.1 root hub

device is there

[ 2083.447606] usb 1-1: new high-speed USB device number 6 using ehci-pci
[ 2083.582909] usb 1-1: New USB device found, idVendor=0456, idProduct=b673
[ 2083.582935] usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[ 2083.582952] usb 1-1: Product: PlutoSDR (ADALM-PLUTO)
[ 2083.582967] usb 1-1: Manufacturer: Analog Devices Inc.
[ 2083.582982] usb 1-1: SerialNumber: 100000235523730700230029090216ea7c
[ 2083.590405] rndis_host 1-1:1.0 eth0: register 'rndis_host' at usb-0000:00:1d.7-1, RNDIS device, 00:e0:22:b6:0f:8b
[ 2083.591828] usb-storage 1-1:1.2: USB Mass Storage device detected
[ 2083.592234] scsi host7: usb-storage 1-1:1.2
[ 2083.593623] cdc_acm 1-1:1.3: ttyACM0: USB ACM device
[...]

PR #83, opened 2017-01-03
Be a little more verbose

rgetz:
Hopefully this will avoid issues like #81 where people (like me) forget to install udev rules, or run as root.

### DIFF
--- a/tests/iio_info.c
+++ b/tests/iio_info.c
@@ -24,6 +24,14 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef _WIN32
+#include <windows.h>
+#include <stdio.h>
+#include <lmcons.h>
+#else
+#include <unistd.h>
+#include <sys/types.h>
+#endif
 
 #define MY_NAME "iio_info"
 
@@ -68,12 +76,48 @@ static void usage(void)
 					options_descriptions[i]);
 }
 
+static bool isadmin(void)
+{
+	bool ret = false;
+
+#ifdef _WIN32
+	HANDLE hToken = NULL;
+
+	if(OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &hToken)) {
+		TOKEN_ELEVATION Elevation;
+		DWORD cbSize = sizeof(TOKEN_ELEVATION);
+
+		if(GetTokenInformation(hToken, TokenElevation, &Elevation,
+					sizeof(Elevation), &cbSize))
+			ret = Elevation.TokenIsElevated;
+
+		if(hToken)
+			CloseHandle(hToken);
+	}
+#else
+	uid_t uid = getuid(), euid = geteuid();
+
+	if (uid == 0 || uid != euid)
+		ret = true;
+#endif
+
+	return ret;
+
+}
 static void scan(void)
 {
 	struct iio_scan_context *ctx;
 	struct iio_context_info **info;
 	unsigned int i;
 	ssize_t ret;
+
+	if (!isadmin()) {
+		fprintf(stderr, "As recommeneded, you are not running as root or administrator,\n"
+				"however, you may not see all devices.\n" );
+	} else {
+		fprintf(stderr, "*** WARNING: running as root or admin. Please do not do this!\n"
+				"***          Please modify device permissions or install correct udev rules.\n");
+	}
 
 	ctx = iio_create_scan_context(NULL, 0);
 	if (!ctx) {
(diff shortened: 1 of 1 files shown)


rgetz:
I will need to rework this to run on Windows before merge.

usb.c
```diff
@@ -1120,8 +1121,12 @@ int usb_context_scan(struct iio_scan_backend_context *ctx,
 		unsigned int interface = 0;
 
 		ret = libusb_open(dev, &hdl);
-		if (ret)
+		if (ret) {
+			if (ret != LIBUSB_ERROR_ACCESS)
+				perms = true;
 			continue;
+		}
+		perms = true;
```
lclausen-adi: Should probably only be set if the device is a IIO device. There might be other devices in the system that are accessible to the user. There might be other devices in the system that are accessible, but this does not mean that there is a IIO udev rule.

And in that case perms = false will be equivalent of simply no IIO device being found. So I think we can just skip the patch and simply print the warning when no IIO contexts have been found.

rgetz: Sort of - if no devices are found - and you can open at least one - perms = true (which is the case when run as root), so it does differentiate between no devices, and no IIO devices.

If you can open one different device (not as root), then no IIO devices, and the wrong udev rule will be the same.

lclausen-adi: For me it makes no sense that some random other accessible device affects the result of this function. There is no correlation between other devices being accessible and the correct udev rules for IIO devices being installed.

The other issue here is that returning an error here will cause iio_scan_context_get_info_list() to fail, even though other contexts returned valid results. So this is pretty much a no-go.

rgetz: Yes - I'm not pleased with the state either, however without walking through sysfs (Linux only, and only for primary contexts), there isn't any way to see what is going on. 

My biggest issue - is just because you have one udev rule installed (by pid/vid for Pluto), means you will not get the warning - and still will not see a different PID/VID (like M2k).

I will revert the libiio changes - but other than printing a warning in libiio about not running as root/administrator - there isn't much you can do - you don't know if you are seeing the actual device you are looking for (M2k), or the device that the udev was installed for (Pluto). 

I guess - we should install an ATTRS{interface}=="IIO" udev rule, and just be done? Or is the pointer/printf about root worth it? It still might be - not sure.

rgetz:
Ok - I reverted changes to libiio, and just print something when not running as root. This is similar to what lsusb -v does. Just need to get working on Windows...

Thoughts/comments?

lclausen-adi:
Sounds good.

rgetz:
So, Windows builds
https://ci.appveyor.com/project/analogdevicesinc/libiio/build/2015.1.717/artifacts

I just need to run time test it on a Windows machine where the local user is not an admin. If anyone wants to try before I can - please report results.

pcercuei:
I don't really like the way it's done, to be honest. Here it suggests the user to run the application as root, and it shouldn't do that. It shouldn't detect if it's running as root, either; this introduces highly non-standard code into iio_info, so even if it works on Windows and Linux, it will break on other systems (BSD, ARM MBED).

rgetz:
The only way to ensure all devices are found is to run as root. This is what lsusb -v does.  

Any other suggestions?

pcercuei:
And then you will have people running Scopy as root.
The way to ensure all devices are found is to have a proper udev rule installed by the Debian and RPM packages.

rgetz:
Right - but how to remind people to install the "proper" udev rule in Linux. My understanding is that it seems to be distribution specific, and you didn't want it to be part of libiio.

rgetz:
BTW - BSD supports getuid, geteuid (same as Linux).
https://www.freebsd.org/cgi/man.cgi?query=geteuid&sektion=2

ARM mbed - doesn't support libusb (running on the target).

I agree - it's a little hokey - but the goal was to remind people to set the correct permissions, without opening bug reports of things not working.

I can add an additional "#ifdef WITH_USB_BACKEND", and update the comment to indicate udev rules might be necessary.

pcercuei:
That doesn't change anything. You can't add a warning about a missing udev rule if (!isadmin()), because the whole point of the udev rule if that you never have to run iio_info (and other libiio stuff) as admin. Having the udev rule installed won't magically give iio_info super-user rights.

rgetz:
Paul:

The goal is avoid false bug reports (plugged in device, and can't see it).

They way I thought to do that was just to remind people about permission issues, and to run things as root (similar to lsusb -v) which hopefully reminds them to install the udev rule. If you don't like that - do you have any other suggestions?

pcercuei:
I understand the goal, but I don't agree with the method - telling people to run libiio apps as root won't make them install the udev rule.

rgetz:
I can change the string, "If you need to run as root, that means you don't have a correct udev rule installed (please fix)". Or something that's pointing people to what we want them to do...

-Robin

rgetz:
I updated the statements that print out (based on the discussion) - let me know if that is OK, or not. If you still don't like it, let me know.

-Robin

rgetz:
Now we are installing udev rules: https://github.com/analogdevicesinc/libiio/commit/4414a30f9f2bd4093e9456780e0faaafba303a54

I'm closing this, without a merge.

PR #88, opened 2017-01-18
Python 3 support

jeras:
I installed libiio from debian including Python bindings, but import only worked in Python 2.7 while Python 3 did not see the package.

So I tried to use the bindings locally, now the issue was with compatibility. After the given patches the example/iio_info.py works properly again. I am sure the example code is correct.

Regarding the wrapper itself, I do not know if how the 'xrange' to 'range' change will affect Python 2.7.
And I do not know if the change is actually tested by the example.

Also this change does not affect install for Python 3, so this is something to look into.

### DIFF
--- a/bindings/python/examples/iio_info.py
+++ b/bindings/python/examples/iio_info.py
@@ -18,42 +18,42 @@ import iio
 def main():
 	ctx = iio.Context()
 
-	print 'Library version: %u.%u (git tag: %s)' % iio.version
+	print ('Library version: %u.%u (git tag: %s)' % iio.version)
 
-	print 'IIO context created: ' + ctx.name
-	print 'Backend version: %u.%u (git tag: %s)' % ctx.version
-	print 'Backend description string: ' + ctx.description
+	print ('IIO context created: ' + ctx.name)
+	print ('Backend version: %u.%u (git tag: %s)' % ctx.version)
+	print ('Backend description string: ' + ctx.description)
 
-	print 'IIO context has %u devices:' % len(ctx.devices)
+	print ('IIO context has %u devices:' % len(ctx.devices))
 
 	for dev in ctx.devices:
-		print '\t' + dev.id + ': ' + dev.name
+		print ('\t' + dev.id + ': ' + dev.name)
 
 		if dev is iio.Trigger:
-			print 'Found trigger! Rate: %u Hz' % dev.frequency
+			print ('Found trigger! Rate: %u Hz' % dev.frequency)
 
-		print '\t\t%u channels found:' % len(dev.channels)
+		print ('\t\t%u channels found:' % len(dev.channels))
 
 		for chn in dev.channels:
-			print '\t\t\t%s: %s (%s)' % (chn.id, chn.name or "", 'output' if chn.output else 'input')
+			print ('\t\t\t%s: %s (%s)' % (chn.id, chn.name or "", 'output' if chn.output else 'input'))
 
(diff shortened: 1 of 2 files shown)


lclausen-adi:
Hi,

Thanks for the patches, looks very good.

I looked into the very same issue last week and one of the questions that came up is whether we want to do the decoding/encoding between str and bytes inside the bindings rather than outside. The reason why we use bytes internally is because that is necessary to call into the C library which expects a 'char *'. But python3 has made unicode strings their default I think we should follow this for libiio python bindings so that the external interface provides str objects rather than bytes objects. And all the conversion is done internally when accessing the C datastructures.

One other thing, in my opinion we should use 'ascii' for the encoding. The C libiio really wouldn't be able to understand utf-8. Any chars outside the ascii range hopefully never show up in any IIO attribute names or values.

jeras:
Hi,

If there is a policy regarding unicode use in kernel interfaces, then it should probably be followed, regardless of current support in the library. Yesterday I would agree with you regarding ascii, it should cover all current use cases I could think off. It was a bit late and I would do anything to just get it done. Today I think some thought should be given future use cases. Anyway it is your decision to make.

Regardless adding Python 3 support might be more important then the decision about what encoding to use.

I was trying to get Zynq XADC to work in a Jupyter notebook, but I could find no working examples to use as reference. I will check PYNQ today.

pcercuei:
I agree that we should keep a 'str' interface rather than having to use 'bytes'.

pcercuei:
I made a 'python3' branch, with a reworked patch on iio.py, and some commits from you (unmodified).
See https://github.com/analogdevicesinc/libiio/compare/python3

Could you tell me if it works for you?
And if so, can I add your Signed-Off-By to the commits?

jeras:
Hi,

For now I will just list a few things so I do not forget them.

I was able to get a buffer frpm Zynq XADC, so now I have something to test with.
The kernel driver for XADC, has a bug, all channels have to be enabled otherwise kernel crashes on buffer request.

There were some other issues with some attributes like frequency not having their value displayed.

Then there are `*_en` and `*_type` listed as attributes:
This one could just be a bug in 0.7 already fixed in master.
```
	iio:device1: xadc
		14 channels found:
			temp0:  (input)
				scale, value: 123.040771484
				offset, value: -2219
				en, value: 0
				index, value: 8
				type, value: le:u12/16>>4
				raw, value: 2766
```

Is there a builtin way to parse `*_type`, and to handle the buffer based on it?

pcercuei:
The `*_en` and `*_type` were hidden in 0.8.
You shouldn't need to parse `*_type`, just do `chn.read(buffer)`. That will return a byte array of the samples for this channel, converted (pass raw = True if you don't want them converted).

pcercuei:
I'd like to merge my python3 branch into master, so that I can tag v0.9, can I add your Signed-Off-By on your commits?

jeras:
Since Python 3 bindings were merged into 0.9, there is no need for this pull request.
But is seems the install procedure only installs bindings for python 2.7.

I used this step to install for Python 3:
```
pip3 install bindings/python/
```

pcercuei:
So I didn't add the documentation fix. Can I add your Signed-Off-By on this one and merge it?

jeras:
The example works fine. So yes, you can add me to Signed-Off-By.

Unfortunately I did not have the time to properly tests buffer access properly, either with the previous of current bindings.

PR #91, opened 2017-02-06
travis: need to build OSX in two places (one for pkg, and one for tar)

rgetz:
osx ssh isn't compatible with the ssh on the server, so hack around it
        merge the deploy, so we don't need to maintain linux & osx separately


pcercuei:
It's hard to see what it does, there are many different fixes squashed together into one big commit...

I'd rather have one commit that fixes the deployment, one commit that fixes tarballs on OSX, one commit that adds the make_[darwin,linux] scripts, one commit that adds the missing CPACK variables in DarwinPackaging.cmake.

rgetz:
Sure - I can split things out now that it is working into the logical bits.

rgetz:
Ok - let me know how that goes?

pcercuei:
Much better :)

One last thing, about the commit message, try to keep the first line under 74 characters, then have an empty line, then the additional description. That's a good general rule and makes it easier to see the commits under 'tig' or github's web UI.

rgetz:
Thanks - will do going forward.

Looks like there is something wrong with travis OSX at the moment -- 
https://www.traviscistatus.com/

Hopefully that will get worked out - but it was tested over the weekend, and I'm pretty sure it is fine. I'm going to try to re-factor the ssh/scp to sftp, as the netstorage seems to like that better - but I need the OS-X to be working first.

-Robin

pcercuei:
Could you add https://github.com/analogdevicesinc/libiio/commit/7856a6c8eb5b5d77303cf1c636f862b420fc442c to your pull request? (the code needs to be moved since deploy_darwin is gone).

The test build https://travis-ci.org/analogdevicesinc/libiio/builds/198929427 completed, where are the uploaded tarballs? So that I can check if the Matlab files are inside.

rgetz:
Can do -- 

tar exists at:

http://swdownloads.analog.com/cse/travis_builds/libiio-0.1.1-Darwin-osx_10.10.tar.gz

File is at:
libiio-0.1.1-Darwin.tar\usr\include\iio-wrapper.h

but I think that is one of the issues - trunk isn't building the OS-X tar files properly. I will merge that into the travis-ci tree and double check.

PR #93, opened 2017-02-08
iio_buffer: add iio_buffer_get_blocking_mode() function

rlefevre:
Hi,

I'm using this new function to implement an `iio_buffer_flush()` function that does not require to pass again the blocking mode as an argument. 

The `iio_buffer_flush()` implementation is there: https://github.com/rlefevre/libiio/commit/bbac4cdc4b89566b42a2d61cd77d95f4d3be89b5 but I'm not sure if it makes sense to submit it as a PR as it can also be implemented outside of libiio.

What do you think?

Thank you

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -123,6 +123,11 @@ int iio_buffer_get_poll_fd(struct iio_buffer *buffer)
 	return iio_device_get_poll_fd(buffer->dev);
 }
 
+bool iio_buffer_get_blocking_mode(struct iio_buffer *buffer)
+{
+	return iio_device_get_blocking_mode(buffer->dev);
+}
+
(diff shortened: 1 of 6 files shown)


pcercuei:
Hi,

I don't think this would work with the network backend, the socket is never in non-blocking mode.

Besides, your application should already know whether or not the iio_buffer is in blocking mode, is this really useful to add in the library?

-Paul

lclausen-adi:
I'd rather not add new non-blocking related API. The non-blocking mode is unfortunately broken and can lead to deadlocks, using it is not recommended.

rlefevre:
Ok, I understand. I'm closing the request.

PR #99, opened 2017-05-08
Local backend: Avoid double free when create_device is aborted.

gwendalcr:
After freeing protected_attrs array, set it to NULL. Otherwise, it
could be freed a second time. It can happen in create_device() when
handle_scan_elements() returns a failure, forcing create_device() to
abort.


pcercuei:
Makes sense. Thanks!

PR #100, opened 2017-06-01
Test improvements

rgetz:
improve the output of iio_info, and add iio_attr (a util which reads/write attributes).


pcercuei:
Thanks!

rgetz:
Thanks for the reviews and feedback.

rgetz:
deleting branch that that it is in master.

PR #103, opened 2017-06-22
test/iio_attr: fix compiler warnings about printf arguments

mhei:
man 3 printf recommends to use z length modifier for ssize_t.

This fixes several warnings (wrapped line for readability):

warning: format '%li' expects argument of type 'long int',
  but argument 3 has type 'ssize_t {aka int}' [-Wformat=]

Signed-off-by: Michael Heimpold <mhei@heimpold.de>

### DIFF
--- a/tests/iio_attr.c
+++ b/tests/iio_attr.c
@@ -157,17 +157,17 @@ static void dump_device_attributes(const struct iio_device *dev,
 		printf("'%s'\n", buf);
 	} else {
 		iio_strerror(-ret, buf, sizeof(buf));
-		printf("ERROR: %s (%li)\n", buf, ret);
+		printf("ERROR: %s (%li)\n", buf, (long)ret);
 	}
 
 	if (wbuf) {
 		ret = iio_device_attr_write(dev, attr, wbuf);
 		if (ret > 0) {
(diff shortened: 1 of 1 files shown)


rgetz:
Thanks for the patch - this looks fine to me.

However - 

I think this would mean we need VS2013 or higher (e.g. _MSC_VER >= 1800). %zd was not supported before then. (I don't think that is a problem, but wanted to point it out).
https://blogs.msdn.microsoft.com/vcblog/2014/06/03/visual-studio-14-ctp/#div-comment-77743

And recent (?)  versions of mingw might also run into similar things, since they use the Microsoft runtime.
https://sourceforge.net/p/mingw-w64/bugs/377/

I will let Lars comment (as Paul is out on vacation).

-Robin

mhei:
Ah good point, I did have had these "older" VS versions on the radar anymore. Let's wait for further comments - I'm really unsure whether it's worth supporting them still...

pcercuei:
I'd rather keep support for VS 2013. The way to avoid the warning then would be to cast the variables to (long), I think.

mhei:
I've updated the PR.

pcercuei:
Looks good 👍 
Thanks!

PR #105, opened 2017-07-06
examples/ad9361-iiostream: Fix assert() usage with NDEBUG

mhennerich:
Make sure we always evaluate the expression even when NDEBUG is defined.

### DIFF
--- a/examples/ad9361-iiostream.c
+++ b/examples/ad9361-iiostream.c
@@ -19,7 +19,6 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


pcercuei:
The messages will be different when NDEBUG is set or not, I think it would be better not to use assert() at all but the same error handling function in all cases.

mhennerich:
Ok - pushed some updates.

pcercuei:
Could you just rebase your commit on top of the latest master? Then I'll merge it.

mhennerich:
Done

PR #107, opened 2017-07-17
CMake: don't use MATCHES for non-regex

gizero:
Right hand side expression will always expand to a path, never to a regex.
Prefer STREQUAL to MATCHES for comparison to avoid unexpected results if i.e.
CMAKE_CURRENT_SOURCE_DIR length exceeds max allowed regex length.

Signed-off-by: Andrea Galbusera <gizero@gmail.com>


pcercuei:
Looks good. Thanks!

PR #124, opened 2017-10-17
appveyor.yml: Create MinGW builds for Windows 32 bit and 64 bit.

AlexandraTrifan:
Provide ZIP artifacts compiled with MinGW that can be linked to any application, but don't provide an installer. 
Adding mingw to the list of supported compilers makes it easier to use libiio with several projects.


appveyor.yml
```diff
@@ -4,69 +4,132 @@ clone_depth: 1
...
+
+configuration:
+    - Release
+
 install:
+    - if "%COMPILER%"=="mingw" (set GENERATOR=Unix Makefiles)
+    - if "%COMPILER%"=="msvc" (set GENERATOR=Visual Studio 12)
     - echo "Checking out sub-modules..."
     - git submodule update --init
+    - cd ..
```
lclausen-adi: Is this cd necessary? It is only followed by a echo and then another cd to an absolute location

rgetz:
I think most people want a single zip file with the msvs and the mingw directories in them. For example:
https://github.com/libusb/libusb/releases/download/v1.0.21/libusb-1.0.21.7z

has a few top level folders; MS64, MS32, MinGW64, MinGW32, include (which is common for all), and a README.txt file (which says how to use things).

I would think we would want something similar. The README.txt file can just be put into the CI/appveyor directory, and you can just copy it into the zip file when you are making it...

Thanks
-Robin

rgetz:
Lars/Alexandra: does this need any further updates? Or are we waiting on end user feedback before a merge?

AlexandraTrifan:
I think the only thing that is missing is README.txt. I didn't add it because I am not exactly sure what we want to write in that file.

rgetz:
I can add a README.txt this week sometime

mhennerich:
The README is still missing @rgetz.
I merge this PL anyways and open Isse #140 instead.

PR #127, opened 2017-11-08
iiod: fix crash setting buffers on invalid device

abrauchli:
Fix a segfault when setting the kernel buffers by checking for a valid
device before calling iio_device_set_kernel_buffers_count.

Signed-off-by: Andreas Brauchli <andreas.brauchli@sensirion.com>


lclausen-adi:
Looks good, thanks.

PR #128, opened 2017-11-08
Update channel type and modifier lists

abrauchli:
Update the channel type and modifier lists against Linux 4.14-rc8.

Signed-off-by: Andreas Brauchli <andreas.brauchli@sensirion.com>

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -51,6 +51,11 @@ static const char * const iio_chan_type_name_spec[] = {
 	[IIO_CONCENTRATION] = "concentration",
 	[IIO_RESISTANCE] = "resistance",
 	[IIO_PH] = "ph",
+	[IIO_UVINDEX] = "uvindex",
+	[IIO_ELECTRICALCONDUCTIVITY] = "electricalconductivity",
+	[IIO_COUNT] = "count",
+	[IIO_INDEX] = "index",
+	[IIO_GRAVITY] = "gravity",
 };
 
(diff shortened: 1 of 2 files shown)


channel.c
```diff
@@ -51,6 +51,11 @@ static const char * const iio_chan_type_name_spec[] = {
 	[IIO_CONCENTRATION] = "concentration",
 	[IIO_RESISTANCE] = "resistance",
 	[IIO_PH] = "ph",
+	[IIO_UVINDEX] = "uvindex",
+	[IIO_ELECTRICALCONDUCTIVITY] = "electricalconductivity",
+	[IIO_COUNT] = "count",
+	[IIO_INDEX] = "index",
+	[IIO_GRAVITY] = "gravity",
```
lclausen-adi: Hi,

Thanks for the patch. I think we also need to update the iio_chan_type and iio_modifier enums in iio.h

abrauchli: Hi @lclausen-adi, indeed, sorry about that. I'll amend it.

abrauchli:
Thanks for double-checking @lclausen-adi, I must have looked at the implementation in industrialio-core.c. Falls into the no more pull requests in the after-hours.

Cheers

lclausen-adi:
It seems the CI has trouble building the latest changes, but I think that is an issue with the CI.

Will merge, thanks.

PR #130, opened 2017-11-12
Cmake: If the version of libserialport isn't 0.1.1; don't fail, just warn

rgetz:
This makes it easier (possible) for those who are using distributions that
	don't keep up to libserialport uptodate (like Ubuntu 16 LTS).
	You can build without this, just not automatically (it requires some command line options).


pcercuei:
Looks OK to me.

rgetz:
Thanks for having a look.

PR #131, opened 2017-11-12
Travis ci3

rgetz:
make releases/tags on travis-ci more robust, and handle more issues.
 - release OS-X files
 - don't overwrite files on release.


rgetz:
I think this is necessary before the next release.

mhennerich:
I assume this have been tested.

rgetz:
Its tested as much as it can be - there is some code only run during a tag, and doesn't get used until we make a release.

PR #133, opened 2017-11-15
Fix block allocation

lclausen-adi:


pcercuei:
I think if we get only one block and requested more, it should fail as well, because there's no way we can do streaming in that case.

PR #135, opened 2017-11-15
Expose Buffer Attributes

mfornero:
Provide a mechanism to expose buffer attributes via libiio.

Buffer attributes may include:

* Watermarks
* FIFO Controls
* Buffer fill levels (see https://www.spinics.net/lists/linux-iio/msg36130.html)

Buffer attributes are exposed as a iio_device API, as they may need to be set
*before* creating and enabling the iio_buffer


local.c
```diff
@@ -109,6 +109,12 @@ static const char * const device_attrs_blacklist[] = {
 	"uevent",
 };
 
+static const char * const buffer_attrs_reserved[] = {
+	"length",
+	"watermark",
```
mfornero: Would we want to also expose watermark? I don't see an existing mechanism to interact with this attribute in libiio

local.c
```diff
@@ -1488,6 +1494,22 @@ static int detect_and_move_global_attrs(struct iio_device *dev)
...
+	const char *name = strrchr(path, '/') + 1;
+	int i;	
+
+	for (i = 0; i < ARRAY_SIZE(buffer_attrs_reserved); i++)
+		if (!strcmp(buffer_attrs_reserved[i], name))
+			return 0;
+
+	iio_snprintf(buf, sizeof(buf), "buffer/%s", name);
+
+	return add_attr_to_device(dev, buf);
```
lclausen-adi: How do we avoid clashes if the buffer and the device have attributes with the same name?

We initially added the scan_element attributes to the channel attributes, but then eventually clashes happened and we had to undo that. I'd like to avoid that for the buffer attributes/

mfornero: The "buffer/" is retained in the name. So you would have attributes named "buffer/data_available"; this should disambiguate between any other attribute names

mfornero: For example:

```
C:\Work\libiio\build\examples\Debug>iio_info -n 10.0.0.220
Library version: 0.11 (git tag: cb2f40f)
Compiled with backends: xml ip usb serial
IIO context created with network backend.
Backend version: 0.11 (git tag: v0.11  )
Backend description string: 10.0.0.220 Linux buildroot-x86mw 4.9.0-00622-g43fb8a6-dirty #21 Tue Nov 14 20:17:29 EST 2017 x86_64
IIO context has 2 attributes:
        local,kernel: 4.9.0-00622-g43fb8a6-dirty
        ip,ip-addr: 10.0.0.220
IIO context has 2 devices:
        iio:device1: mwipcore0:mm2s0 (buffer capable)
                1 channels found:
                        data0:  (output, index: 0, format: le:U32/32>>0)
                3 device-specific attributes found:
                                attr  0: reset_ip_mode value: none
                                attr  1: reset_ip_mode_available value: none enable disable all
                                attr  2: buffer/data_available value: 0
```

mfornero: @lclausen-adi -- does the use of a "buffer/" namespace resolve your concern about collisions with other attribute names, or is there some other issue I'm missing?

pcercuei:
Would it make sense to add buffer attributes, just like we have channel and device attributes?

mfornero:
If buffer attributes were added, would we expose everything other than ```enable``` & ```length```?

pcercuei:
I guess so. I don't know about "watermark", I've never seen it before.
@lclausen-adi, what do you think?

mfornero:
I haven't used it either, but it appears to be part of the [testing] IIO ABI for some devices:

https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio

pcercuei:
Ok, that's something that could be exposed indeed.

rgetz:
Watermarks are pretty common inside some IIO device classes. The [ADXL345](http://www.analog.com/adxl345) [datasheet](http://www.analog.com/media/en/technical-documentation/data-sheets/ADXL345.pdf) describe watermarks as:

Watermark 
The watermark bit is set when the number of samples in FIFO equals the value stored in the samples bits (Register FIFO_CTL, Address 0x38). The watermark bit is cleared automatically when FIFO is read, and the content returns to a value below the value stored in the samples bits. 

and I think it is inside many other devices with a FIFO on them from other manufactures as well. The trick there is to make sure it gets exposed the same way - from a user perspective - I don't care if the FIFO watermark is DMA inside the SoC, or on a SPI and managed by a peripheral. I think @mhennerich  wrote a lot of the accelerometer drivers - he may have some comments.

-Robin

mfornero:
Given that these attributes are somewhat common, it sounds like refactoring this to iio_buffer attributes API would be a good next step.

mfornero:
One thing to consider-- do some of these buffer attributes become read-only once the buffer is enabled? if so, we may have an issue with order of operations; the constructor for the iio_buffer pointer also enables the buffer, so it may not be possible to write to some of these attributes.

mhennerich:
Devices with an integrated FIFO would use the hwfifo_watermark ABI:

sysfs-bus-iio:What:		/sys/bus/iio/devices/iio:deviceX/buffer/hwfifo_enabled
sysfs-bus-iio:What:		/sys/bus/iio/devices/iio:deviceX/buffer/hwfifo_watermark
sysfs-bus-iio:What:		/sys/bus/iio/devices/iio:deviceX/buffer/hwfifo_watermark_min
sysfs-bus-iio:What:	       /sys/bus/iio/devices/iio:deviceX/buffer/hwfifo_watermark_max
sysfs-bus-iio:What:	       /sys/bus/iio/devices/iio:deviceX/buffer/hwfifo_watermark_available

mfornero:
@lclausen-adi, @pcercuei -- refactored the changes to provide a full set of buffer attribtue APIs

mfornero:
The kernel patch for enabling "data_available" was picked up by linux-iio over the weekend
https://www.spinics.net/lists/linux-iio/msg36277.html

Any comments on the updated iio_buffer attributes APIs @lclausen-adi, @pcercuei @rgetz @mhennerich ?

PR #144, opened 2018-01-12
build: generate static library for libiio

commodo:
For some environments/systems, it's preferred to have a static
libiio library to link into the app.
Especially when the version of libiio installed on
the system/distro is older than the one that's required.

This change enables the generation of a static lib,
along side the shared lib.
Users can choose to use it [or not].

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,8 +303,9 @@ if(WITH_TESTS)
 	add_subdirectory(tests)
 endif()
 
-add_library(iio ${LIBIIO_CFILES} ${LIBIIO_HEADERS} ${LIBIIO_EXTRA_HEADERS})
-set_target_properties(iio PROPERTIES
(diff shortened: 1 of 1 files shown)


pcercuei:
We can do that already.
cmake -DBUILD_SHARED_LIBS=OFF .

commodo:
I did not know about that option.
I tried it, and it generates the static lib only.

Question now is, if it's acceptable to generate both shared & static and have them side-by-side [as this patch proposes], or keep just the shared as default.

I'm fine either way :)

pcercuei:
I'd rather leave it as it is, building both shared and static isn't needed in 99% of the cases.
However the option to build static is hidden (handled directly by cmake), we could have an option(BUILD SHARED LIBS...) to make it more visible.

commodo:
ack ;

no need to add `BUILD_SHARED_LIBS` ; was added in this commit https://github.com/analogdevicesinc/libiio/commit/f52095da2966666f465d77e84d722b8f25e07f8f

will close

thanks :)

pcercuei:
Alright then :)

Issue #94, opened 2017-02-08
Windows doesn't fill in file information (version, name, copyright year, etc).


rgetz:
![image](https://cloud.githubusercontent.com/assets/1017869/22759764/a3d25078-ee21-11e6-802f-2cd342e27b46.png)

I think the fix is:
http://stackoverflow.com/questions/6693100/how-to-generate-windows-dll-versioning-information-with-cmake

More info at:
http://www.codeofclimber.ru/2015/generate-windows-dll-version-info-resource-with-cmake/

rgetz:
I think this is done, so closing

PR #145, opened 2018-01-17
Fill in the Windows file information (version, name, copyright, year).

AlexandraTrifan:
Fill in the Windows file information (version, name, copyright, year).  Requested in issue #94.


CMakeLists.txt
```diff
@@ -4,6 +4,9 @@ project(libiio C)
 set(LIBIIO_VERSION_MAJOR 0)
 set(LIBIIO_VERSION_MINOR 12)
 set(VERSION "${LIBIIO_VERSION_MAJOR}.${LIBIIO_VERSION_MINOR}")
+if (WIN32)
+	string(TIMESTAMP BUILD_YEAR "%Y")
+endif()
```
commodo: curios: do we need this for non-Win32 builds ? or just for Win32 ?

commodo:
rest looks fine from me

[LGTM](http://cdn.343i.org/forum/fileshare/gallery/album_615/gallery_77261_615_18114.jpg)

CMakeLists.txt
pcercuei: The WIN32 variable is defined on Windows, Cygwin and MinGW, independently of the CPU architecture (stupid, I know).

AlexandraTrifan:
Done. Sorry for the delay.

pcercuei:
Unfortunately the AppVeyor build seems broken (for a different reason). The appveyor.yml does things that are not needed anymore (like forcing a re-install of GCC to get the latest version). Could you maybe take a look?

AlexandraTrifan:
I added a commit regarding this. In my opinion, the initial core update can be dropped, since we update the package list before installing certain packages. 
If this option does not look ok to you, it seems that there is another workaround using an undocumented option _--ask_ (mainly added for internal usage, though).

pcercuei:
That's OK to me, we can merge. Thanks!

PR #146, opened 2018-01-29
Allow iio.Buffer to be aware of the Context, to avoid a segmentation violation

hrjonashansen:
Added a new parameter (ctx = None) for the iio.Buffer class constructor. This parameter is stored in self._ctx as a reference to a Context variable. This reference ensures that the Context outlives the buffer, if this is not the case a segmentation violation will occur when destroying the buffer.

### DIFF
--- a/bindings/python/iio.py
+++ b/bindings/python/iio.py
@@ -519,6 +519,10 @@ class Buffer(object):
 		self._length = samples_count * device.sample_size
 		self._samples_count = samples_count
 
+		self._ctx = device.ctx() 
+		# Holds a reference to the corresponding IIO Context. This ensures that
+		# every iio.Buffer object is destroyed before its corresponding IIO Context.
+
 	def __del__(self):
 		if self._buffer is not None:
 			_buffer_destroy(self._buffer)


lclausen-adi:
Thanks, looks good. Can you squash both commits into one and add a Signed-of-by?

hrjonashansen:
The commits have been squashed and a Signed-of-by has been added.

hrjonashansen:
@lclausen-adi  The commits have been squashed and a Signed-of-by has been added.

hrjonashansen:
@pcercuei My Python script produces a segmentation violation without this reference. Given that the reference to a iio.Context that the iio.Devices class has is weak, it does not ensure that the context is destroyed after the iio.Buffer.

pcercuei:
What I meant was, maybe you can get the context from the "device" parameter.

hrjonashansen:
Ah, I see. No, It would still be a weak reference.

pcercuei:
It's not possible to get a regular reference from a weak reference?

lclausen-adi:
It is possible. By invoking the weak reference with ().

``` Python
>>> class foo(): pass
... 
>>> o = foo()
>>> o2 = weakref.ref(o)
>>> o3 = o2()
>>> del o
>>> o
NameError: name 'o' is not defined
>>> o3
<__main__.foo instance at 0x7f61cea1be60>
>>> o2
<weakref at 0x7f61cea1f418; to 'instance' at 0x7f61cea1be60>
>>> del o3
>>> o2
<weakref at 0x7f61cea1f418; dead>
>>> 
```

hrjonashansen:
Well, that makes it a lot smoother. I have implemented this instead. Thank you.

pcercuei:
Please add a small comment in the code to explain why it's there; otherwise I fear that sooner or later somebody will think it's dead code :)

hrjonashansen:
@pcercuei That does seem like a valid concern! Thus, I have added a short comment :-)

pcercuei:
I don't think you need a Doxygen comment, a regular comment is fine (it's useful for people reading the code, not for the devs who want to use the bindings). Otherwise it looks OK for me.

hrjonashansen:
@pcercuei That does seem like a valid point, I have corrected my comment.

pcercuei:
I'm going to be annoying, but could you make it a multi-line comment? :D

hrjonashansen:
@pcercuei Would that be all? ;-)

pcercuei:
Yes :) Good to merge.

PR #147, opened 2018-02-13
Improve install

afreof:
My goal is to update the openembedded/Yocto recipe for libiio. I'm using current Yocto with usrmerge flag enabled. This requires these two patches to build.
Installing udev rules to /usr/lib/udev/rules.d instead of /lib/udev/rules.d is common for systemd based Linux distributions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -399,6 +399,9 @@ if(WITH_IIOD)
 	option(WITH_SYSTEMD "Enable installation of systemd service file for iiod" OFF)
 	set(SYSTEMD_UNIT_INSTALL_DIR /lib/systemd/system CACHE PATH "default install path for systemd unit files")
 
+	option(WITH_SYSVINIT "Enable installation of SysVinit script for iiod" OFF)
+	set(SYSVINIT_INSTALL_DIR /etc/init.d CACHE PATH "default install path for SysVinit scripts")
+
 	if (NOT PTHREAD_LIBRARIES)
 		message(WARNING "IIOD requires threads support; disabling")
 		set(WITH_IIOD OFF CACHE BOOL "" FORCE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -402,6 +402,9 @@ if(WITH_IIOD)
 	option(WITH_SYSVINIT "Enable installation of SysVinit script for iiod" OFF)
 	set(SYSVINIT_INSTALL_DIR /etc/init.d CACHE PATH "default install path for SysVinit scripts")
 
+	option(WITH_UPSTART "Enable installation of upstart config file for iiod" OFF)
+	set(UPSTART_CONF_INSTALL_DIR /etc/init CACHE PATH "default install path for upstart conf files")
+
 	if (NOT PTHREAD_LIBRARIES)
 		message(WARNING "IIOD requires threads support; disabling")
 		set(WITH_IIOD OFF CACHE BOOL "" FORCE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -396,6 +396,9 @@ if(OSX_PACKAGE)
 endif()
 
 if(WITH_IIOD)
+	option(WITH_SYSTEMD "Enable installation of systemd service file for iiod" OFF)
+	set(SYSTEMD_UNIT_INSTALL_DIR /lib/systemd/system CACHE PATH "default install path for systemd unit files")
+
 	if (NOT PTHREAD_LIBRARIES)
 		message(WARNING "IIOD requires threads support; disabling")
 		set(WITH_IIOD OFF CACHE BOOL "" FORCE)
--- a/iiod/CMakeLists.txt
+++ b/iiod/CMakeLists.txt
@@ -62,3 +62,8 @@ add_definitions(-D_GNU_SOURCE=1)
 if(NOT SKIP_INSTALL_ALL)
 	install(TARGETS iiod RUNTIME DESTINATION ${CMAKE_INSTALL_SBINDIR})
 endif()
+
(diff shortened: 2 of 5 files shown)


pcercuei:
I don't want the checks for existing directories. That will break when cross-compiling when using DESTDIR. Better use option() and let the user override the paths.

afreof:
Thank you for your feedback and for considering cross compilation too.
I think the patch is fine for cross compiling. I tested it in a Yocto cross compiled environment. It works without any extra cmake arguments.

It uses an absolute path (the _FULL_ variant of CMAKE_INSTALL_FULL_LIBDIR) which is valid for cross compiled environments too. If the folder is not there the install procedure is expected to behave exactly as before, without any disadvantages.

Some lines from the install log of my environment:
```
DEBUG: Executing shell function do_install
NOTE: DESTDIR=/home/user/projects/folder/yocto-prj/build/tmp/work/cortexa7hf-neon-prj-linux-gnueabi/libiio/0.14+git999-r0/image cmake --build /home/user/projects/folder/yocto-prj/build/tmp/work/cortexa7hf-neon-prj-linux-gnueabi/libiio/0.14+git999-r0/libiio-0.14+git999/ --target install -- -j 4
...

Install the project...
/home/user/projects/folder/yocto-prj/build/tmp/work/cortexa7hf-neon-prj-linux-gnueabi/libiio/0.14+git999-r0/recipe-sysroot-native/usr/bin/cmake -P cmake_install.cmake
-- Install configuration: "RelWithDebInfo"
-- Installing: /home/user/projects/folder/yocto-prj/build/tmp/work/cortexa7hf-neon-prj-linux-gnueabi/libiio/0.14+git999-r0/image/usr/lib/pkgconfig/libiio.pc
-- Installing: /home/user/projects/folder/yocto-prj/build/tmp/work/cortexa7hf-neon-prj-linux-gnueabi/libiio/0.14+git999-r0/image/usr/lib/libiio.so.0.14
-- Installing: /home/user/projects/folder/yocto-prj/build/tmp/work/cortexa7hf-neon-prj-linux-gnueabi/libiio/0.14+git999-r0/image/usr/lib/libiio.so.0
-- Installing: /home/user/projects/folder/yocto-prj/build/tmp/work/cortexa7hf-neon-prj-linux-gnueabi/libiio/0.14+git999-r0/image/usr/lib/libiio.so
-- Installing: /home/user/projects/folder/yocto-prj/build/tmp/work/cortexa7hf-neon-prj-linux-gnueabi/libiio/0.14+git999-r0/image/usr/include/iio.h
-- Installing: /home/user/projects/folder/yocto-prj/build/tmp/work/cortexa7hf-neon-prj-linux-gnueabi/libiio/0.14+git999-r0/image/usr/lib/udev/rules.d/90-libiio.rules

```

pcercuei:
It may work on Yocto, but it will break on Buildroot...

lclausen-adi:
My concern here is about predictable builds. The order in which packages are build affects the outcome. If some random package that is build before libiio will create those folders libiio will use them, otherwise not. This means that different runs in the same build system might produce different results.

afreof:
I agree.
There is a rebased version fallowing your advice.

afreof:
I disagree.
I think the init script itself is installed into /etc/init.d folder without any kind of prefix. Theses enumerated names are assigned to links binding an init-script to a run-level. For example if iiod will be started with run level 1 we get a link /etc/rc1.d/K20iiod -> ../init.d/iiod. /etc/rc1.d/K20iiod is a link created by a distro specific tool. cmake provides /etc/init.d/iiod with a neutral name.

pcercuei:
Ok, then is it expected that we create these symlinks from the install script as well?

Another question, does the systemd script has to be "activated" in some way for systemd to pick it up? Or a system restart is enough?

afreof:
I think the most common and also the best approach is to install the scripts/config files like the three patches do it right now. cmake should not do any thing further. Additional steps such as creating sym links define a lot of policy such as the priority or the run-level iiod is started with. It is also distribution specific if the user has to start the service explicitly at run time or if the links are created at build time.

afreof:
Is this ready for merging?

lclausen-adi:
Waiting for Paul for the final word on this.

pcercuei:
Well I don't really know. I have the feeling that it's the job of package managers to install the scripts, because of the post-install actions required (symlinks with sysvinit, and systemd script activation).

For that reason I think it makes sense not to install these within libiio's CMake. But we can discuss further, if you think there's a proper way.

afreof:
Arguments to get the init/service files installed by cmake:
- cmake is aware of the installed binaries and the corresponding installation paths. A templating engine is required to generate the service file to be consistent with the binary install path (and may be other variables in the init/service file). cmake provides a template engine.
- This step is generic for all Distros (read only file systems, rw filesystems, services started by default if installed, services started manually, independent from run-levels a service is started...)
- Recommended by systemd author http://0pointer.de/public/systemd-man/daemon.html, please refer to section "Installing Systemd Service Files"
- Recommended by Fedora: https://fedoraproject.org/wiki/Packaging:Systemd, please refer to section "Unit Files"

Arguments for not adding further steps (e.g. creating sym-links) to cmake install step:
- Debian creates the sysv links at runtime by update-rc.d not at install time. [Debian Policy](https://www.debian.org/doc/debian-policy/) 9.3.3.1. Managing the links
- Systemd based Distros (Fedora, SuSE, Debian) create the links in the scripts executed at package (un)installation time. The links are not included in the package.
- This step is Distro specific

I hope these arguments motivate you to merge this pull request.

pcercuei:
Alright, thanks for clearing that up. I'll do a last quick review later and merge if I don't see anything strange.

pcercuei:
Merged. Thanks!

PR #151, opened 2018-02-26
Fix optarg() handling

lclausen-adi:
A lot of the IIO tools currently assumes that the argument for an option that
is parsed by getopt_long() is in the argv[] following the option and then
the combination of option and argument always consumes two argv[] entries.

This is not necessarily true though. getopt_long() accepts option and
argument in the same argv[] entry. For short options the argument can
directly follow the option character and for long options the argument can
follow in the same argv[] entry separated by a '=' character.

E.g. both -u ip:192.168.1.1 and -uip:192.168.1.1 as well as --uri
192.168.1.1 and --uri=192.168.1.1 are equivalent and all valid.

As a result of this the tools will show undefined behavior when
option and argument are passed in the same argv[] entry (E.g. crash with a
segmentation fault).

To address this properly use the optarg and optind variables that are
provided by the getopt() interface. optarg points to the start of the
argument for the current option and optind will point to the first non
option argv[] entry after all options have been scanned.

### DIFF
--- a/tests/iio_genxml.c
+++ b/tests/iio_genxml.c
@@ -66,8 +66,10 @@ int main(int argc, char **argv)
 {
 	char *xml;
 	struct iio_context *ctx;
-	int c, option_index = 0, arg_index = 0, xml_index = 0, ip_index = 0;
-	int uri_index = 0;
+	int c, option_index = 0;
+	const char *arg_uri = NULL;
+	const char *arg_xml = NULL;
+	const char *arg_ip = NULL;
 	enum backend backend = LOCAL;
 
 	while ((c = getopt_long(argc, argv, "+hn:x:u:",
@@ -82,8 +84,7 @@ int main(int argc, char **argv)
 				return EXIT_FAILURE;
 			}
 			backend = NETWORK;
-			arg_index += 2;
-			ip_index = arg_index;
+			arg_ip = optarg;
 			break;
 		case 'x':
 			if (backend != LOCAL) {
@@ -91,12 +92,10 @@ int main(int argc, char **argv)
 				return EXIT_FAILURE;
 			}
 			backend = XML;
-			arg_index += 2;
-			xml_index = arg_index;
+			arg_xml = optarg;
 			break;
 		case 'u':
-			arg_index += 2;
-			uri_index = arg_index;
+			arg_uri = optarg;
 			backend = AUTO;
 			break;
 		case '?':
@@ -104,18 +103,18 @@ int main(int argc, char **argv)
 		}
 	}
 
-	if (arg_index >= argc) {
+	if (optind != argc) {
 		fprintf(stderr, "Incorrect number of arguments.\n\n");
 		usage();
 		return EXIT_FAILURE;
 	}
 
 	if (backend == AUTO) 
-		ctx = iio_create_context_from_uri(argv[uri_index]);
+		ctx = iio_create_context_from_uri(arg_uri);
 	else if (backend == XML)
-		ctx = iio_create_xml_context(argv[xml_index]);
+		ctx = iio_create_xml_context(arg_xml);
 	else if (backend == NETWORK)
-		ctx = iio_create_network_context(argv[ip_index]);
+		ctx = iio_create_network_context(arg_ip);
 	else
 		ctx = iio_create_default_context();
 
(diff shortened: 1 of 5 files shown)


tests/iio_readdev.c
```diff
@@ -249,7 +249,8 @@ int main(int argc, char **argv)
 {
 	unsigned int i, nb_channels;
 	unsigned int buffer_size = SAMPLES_PER_READ;
-	int c, option_index = 0, arg_index = 0, ip_index = 0, uri_index = 0;
+	char *arg_uri = NULL, *arg_ip = NULL;
```
commodo: nitpick: `const char *` could be added here

commodo: i checked that the `iio_create_context_` functions do take `const char *` args

lclausen-adi: right, I wanted to change that, but forgot about it, thanks.

tests/iio_readdev.c
```diff
@@ -355,15 +349,15 @@ int main(int argc, char **argv)
...
+	if (argc == optind + 2) {
 		/* Enable all channels */
 		for (i = 0; i < nb_channels; i++)
 			iio_channel_enable(iio_device_get_channel(dev, i));
 	} else {
 		for (i = 0; i < nb_channels; i++) {
 			unsigned int j;
 			struct iio_channel *ch = iio_device_get_channel(dev, i);
-			for (j = arg_index + 2; j < (unsigned int) argc; j++) {
+			for (j = optind + 1; j < (unsigned int) argc; j++) {
```
commodo: curios: is this a bugfix ?

commodo: the reason for asking is that `j` starts at `optind + 1 ` where previously it was `arg_index + 2`; 
i admit i did not check the docs for `optind` yet, but in the above if-check it's `if (argc == optind + 2)`

lclausen-adi: arg_index = optind - 1

pcercuei:
You should rebase on top of master, to see if it builds on AppVeyor.

lclausen-adi:
v2: Use **const** char * for the argument pointers.

tests/iio_readdev.c
```diff
@@ -355,15 +349,15 @@ int main(int argc, char **argv)
 
 	nb_channels = iio_device_get_channels_count(dev);
 
-	if (argc == arg_index + 2) {
+	if (argc == optind + 2) {
```
commodo: in that case (https://github.com/analogdevicesinc/libiio/pull/151/commits/88ef7db62e2e3892dd8e1df1307025cce2a7cf49#r170843858) , should it be `argc == optind + 1` ?

lclausen-adi: Yes, thanks.

tests/iio_genxml.c
```diff
@@ -82,40 +84,37 @@ int main(int argc, char **argv)
...
+			arg_uri = optarg;
 			backend = AUTO;
 			break;
 		case '?':
 			return EXIT_FAILURE;
 		}
 	}
 
-	if (arg_index >= argc) {
+	if (optind != argc) {
```
commodo: curios: `optind + 1 != argc` here ?

commodo:
rest looks good
i really like the change :)

tests/iio_genxml.c
lclausen-adi: optind should be == argc if all arguments have been consumed. It think this is correct the way it is.

commodo: ack ; thanks for the info ;
i admit i got a bit confused in `optind`  stuff ;

lclausen-adi:
v3: Fix iio_readdev check for optional channel list

pcercuei:
Could you test a build with uClibc and musl? (Using BuildRoot for instance). Just to be sure we don't break anything.

lclausen-adi:
iiod uses optarg, if it works, this should work as well.

commodo:
maybe we could integrate the glibc, uClibc & musl compat matrix/stuff into an automated test-run ;

AFAIK, musl's and uClibc's compatibility level with glibc for `optag` and `getopt()` is pretty good [especially if versions are not too old];
I've written [a while back] code using `getopt()` and `getopt_long` on uClibc, then converted it to `musl` and `glibc` and the transition was seamless ;

so if it works on glibc, I would accept this change ;
and continue the effort to develop an automated test-matrix

Issue #154, opened 2018-03-07
C# Binding - IIOBuffer.fill() Missing Functionality


raceybe:
The functionality provided by the fill() method uses the Marshal.Copy command:

`Marshal.Copy(array, 0, iio_buffer_start(buf), length);`

However, this is only good for copying from the managed array to the unmanaged OUTPUT buffer.

Another function is required for input buffers (or perhaps a flag to select the appropriate Marshal.Copy overload). For an INPUT buffer, we need to copy from unmanaged buffer to managed array...something like the following:

```
public void fillFromInput(short[] array)
{
long length = (long)iio_buffer_end(buf) - (long)iio_buffer_start(buf);
if (length > array.Length)
length = array.Length;
Marshal.Copy(iio_buffer_start(buf), array, 0, array.Length);
}
```

Notice the new order of the input arguments.

pcercuei:
Indeed, we seem to lack a function here. Feel free to send a patch, and we can add it (note: I'd prefer IOBuffer.read() as the function name).

By the way, you can extract data from the IOBuffer using Channel.read(), and it gives you the samples demultiplexed for that channel. Which might not be what you want to achieve, I just thought I'd mention it.

raceybe:
I will look into the Channel.read(). Does the same thing apply to Channel.write() as well? That may just render a days work obsolete.

it appears read() and write() operate on a byte level, so the 16 bit word would still need assembly from the stream? I had modified/added my function as a short so the entire 16 bits could be read/write in one shot.

pcercuei:
The same thing applies to Channel.write() as well, yes.

Channel.read()/write() use 8-bit arrays, but you're free to reinterpret (cast) them as 16-bit arrays. This will work as long as the hardware expects 16-bit samples.

rgetz:
@AlexandraTrifan : can you have a look at this?

rgetz:
Pull request merged - so closing this.

Issue #153, opened 2018-03-07
C# Binding - IIOBuffer.fill() Overflow


raceybe:
In IIOBuffer.cs, the fill() method reports an overflow when the iio_buffer_end(buf) and iio_buffer_start(buf) Intptr is cast to int. Changing the cast to long prevents this overflow...does this depend on the bitness of the OS or the bitness the DLL is compiled for?

`int length = (int) iio_buffer_end(buf) - (int) iio_buffer_start(buf);`
becomes
`long length = (long) iio_buffer_end(buf) - (long) iio_buffer_start(buf);`

pcercuei:
I guess it should be "long", yes. Could you send a patch?

rgetz:
@AlexandraTrifan : can you have a look at this?

rgetz:
Pull request merged - so closing this issue.

PR #155, opened 2018-03-07
Change cast from int to long and add read()

raceybe:
Changed the casting and variable size from int to long on line 138.

Added read() method.


raceybe:
This pull request addresses issues #153 and #154

raceybe:
I omitted a necessary casting and need to update it.

PR #156, opened 2018-03-07
Modified casting and added read()

raceybe:
Addressing #153 and #154.

### DIFF
--- a/bindings/csharp/IOBuffer.cs
+++ b/bindings/csharp/IOBuffer.cs
@@ -132,12 +132,25 @@ namespace iio
         /// <summary>Copy the given array of samples inside the <see cref="iio.IOBuffer"/> object.</summary>
         /// <param name="array">A <c>byte</c> array containing the samples that should be written.</param>
         /// <remarks>The number of samples written will not exceed the size of the buffer.</remarks>
+        /// <remarks>Use this to fill an output buffer.</remarks>
         public void fill(byte[] array)
         {
-            int length = (int) iio_buffer_end(buf) - (int) iio_buffer_start(buf);
+            long length = (long) iio_buffer_end(buf) - (long) iio_buffer_start(buf);
             if (length > array.Length)
                 length = array.Length;
-            Marshal.Copy(array, 0, iio_buffer_start(buf), length);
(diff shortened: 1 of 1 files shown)


pcercuei:
If you want to modify your pull request, no need to close it and open a new one, just force-push your branch.

You have two different changes in one commit, please split those in two distinct commits.

Drop the remark "use this with an input/output buffer", in theory buffers can be both at the same time.

I disagree about the prototype of your function read(), it should return a byte[] array, instead of taking it as a parameter.

Finally, the comment that describes what read() does is not correct.

pcercuei:
Oh, and I forgot, please sign your commit using "git commit -s ..." or if you already commited it, "git commit -s --amend --no-edit". Then you have the copyright on your code.

raceybe:
I'm in agreement with everything you said, but I'm not sure how to go back and make those changes. I'm not using the desktop client, just the web interface.

commodo:
I just tested Github's file `Edit` feature.
I did not know about it, though probably I did notice the button, I never thought to check.
Unfortunately, there does not seem to be a `Edit` feature for git commits, or at least a `amend last commit` feature.

I will admit, that I am a bit worried about allowing the use of that feature, because it seems a bit inflexible during reviews, and it means that each PR re-spin needs to be created from scratch, and a new PR to be created.
This can cause quite a bit frustration for big PRs.
`git` allows quite a bit of flexibility when editing commits, and re-spinning branches.

I'd say that maybe for small changes/updates [like this] it would be fine to re-spin via the web-GUI, as long as it's ok with the submitter [i.e. @raceybe  ].
Unless there are other objections.

If there are no objections, then a new PR with 2 web-GUI edits could be done.
As for the commit comment template, something like this is preferred usually :)
[git commit template](https://i.imgur.com/qvbe9ERl.png)

rgetz:
@raceybe can you let us know what your intent is?

It would be best if you can add your Signed-off-by. Without doing that - no one can use your code. :(

https://signed-off-by.me/

without your explicated confirmation that you own the code - and are contributing it - we can't take it.

-Robin

rgetz:
Since this has been open for two years with no signed-off-by; I will assign the issues to @AlexandraTrifan ; and she can re-implement things without looking at this patch - so we can apply them to the repo.

PR #157, opened 2018-03-09
tests/iio_writedev: Add iio_writedev analogous to iio_readdev

mhennerich:
This test program uses the libiio library to read standard input and write
as raw stream to a device. This utility is vastly based on iio_readdev.
It adds one more parameter option to use cyclic buffers.

### DIFF
--- a/tests/iio_writedev.c
+++ b/tests/iio_writedev.c
@@ -1,0 +1,453 @@
+/*
+ * libiio - Library for interfacing industrial I/O (IIO) devices
+ *
+ * Copyright (C) 2014-2018 Analog Devices, Inc.
+ * Author: Paul Cercueil <paul.cercueil@analog.com>
+ *         that Michael Hennerich <michael.hennerich@analog.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * */
+
+#include <errno.h>
+#include <getopt.h>
+#include <iio.h>
+#include <signal.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+#ifdef _WIN32
+#include <windows.h>
+#include <fcntl.h>
+#include <io.h>
+#else
+#include <unistd.h>
+#endif
+
+#define MY_NAME "iio_writedev"
+
+#define SAMPLES_PER_READ 256
+#define DEFAULT_FREQ_HZ  100
+
+static const struct option options[] = {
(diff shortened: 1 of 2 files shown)


tests/iio_writedev.c
```diff
@@ -0,0 +1,437 @@
...
+		if (ret < 0) {
+			if (app_running) {
+				char buf[256];
+				iio_strerror(-ret, buf, sizeof(buf));
+				fprintf(stderr, "Unable to push buffer: %s\n", buf);
+			}
+			break;
+		}
+
+		while(cyclic_buffer && app_running)
```
commodo: this looks like it would hang indefinitely here, without any operation;
that is, until `app_running` is false via sig-handler

is this intended ?

commodo: i did not dig here yet;
but does the IIO lib have some internal threading that would do operation here ?

mhennerich: Yes - a cyclic buffer is only pushed once. This ensures that the buffer remains running until a signal is catched. Using sleep here is beneficial since it works cross platform equally well.

mhennerich:
There is still a problem on Windows -
iio_writedev input from a pipe - closes if the binary data appears to contain special chars like EOF.

C:\Users\foo>iio_readdev -n 192.168.2.1 -b 100000 cf-ad9361-lpc | iio_writedev  -n 192.168.2.1 -b 100000  cf-ad9361-dds-core-lpc

I tried several things:

#ifdef _WIN32
	_setmode( _fileno( stdin ), _O_BINARY );
#endif

or 

#ifdef _WIN32
	freopen(NULL, "rb", stdin);
#endif

to no avail - any comments appreciated.

BTW - also tried to change iio_readdev to write binary to stdout - same result.

commodo:
what if you check for  `feof(stdin)` after you get 0 bytes from `fread()` ?

mhennerich:
don't know how this could be used to fix the behavior ...

commodo:
so, in the case of:
```
+				size_t nb = fread(start, 1, len, stdin);
+				if (!nb)
+					goto err_destroy_buffer;
```
that code is exiting if `fread()` returned 0 bytes ;
but there is no check for any `feof()` condition ;
though, i am not sure [either] if doing a `freopen()` would help here [as you mentioned it does not work];
or how to force the stream re-open / skip [or something] ;

i am bit vague on where [in the code] the `freopen()` was used ;

mhennerich:
https://github.com/analogdevicesinc/libiio/blob/iio_writedev/tests/iio_writedev.c#L397

It's purpose was only to change the access mode of stdin to binary.

mhennerich:
my fault - I just realized after I didn't see my debug prints, that cmd.exe was not picking up the latest installations. Using the builds from the zip file work ok.

commodo:
Will merge this later today if no objections

pcercuei:
I'd rather have that bug fixed before...

mhennerich:
which bug? - In fact there was no bug - 
_setmode( _fileno( stdin ), _O_BINARY ); fixed the issue the first time. It was just my testing that flawed.
All my new win with this being fixed didn't succeed. So I was constantly testing my first version.

pcercuei:
Oh, alright. I didn't understand that.

commodo:
any objections left ? :)

commodo:
will merge this later [unless objections]

Issue #158, opened 2018-03-09
LIBIIO access via network on Windows Subsystem for Linux


jaredd:
I've been trying to get my ADALM-Pluto to work over the network context on windows subsystem for linux on Win 10 Pro.  The PlutoSDR is attached through a powered USB hub.  I setup the device according to the analog devices tutorials.  The device operation was verified using the ADI Oscilloscope on windows.  The windows iio_X utilities also work.  

I would like to use Windows Subsystem for Linux for its ease of GNU Radio development.  The libiio library and utilities were built using native (linux) tools on WSL.  The [libiio_git_diff.txt](https://github.com/analogdevicesinc/libiio/files/1798242/libiio_git_diff.txt) patch seems sufficient to get proper operation on WSL.  I'll describe what symptoms indicated a problem and how the patch seems to fix them:

1.  iio_info -n 192.168.2.1 would hang (not exit or error) despite being able to both ping and ssh to the device from WSL.  I'd expect, if there was a problem, that a problem would be indicated and the utility would exit.  In the patch, the lines negating the network_should_retry call seem to result in the right behavior.  It's hard to tell whether this is the right behavior in general, it seems to me if "network_should_retry" returns False then you should return and not retry.  In the repo the behavior is the opposite.

2.  After fix 1, the context could not be created and the utility would exit.  It seems that windows doesn't have good support for MSG_TRUNC through WSL in the network_read_line function.  I simply added an additional __WSL__ define and forced the non linux code to be used instead.  I'm not sure how a good way to auto-detect WSL in CMake, though "Microsoft" is in the uname line below.  Cmake seems like the right place to detect and set defines.

More info.  Development was from a fresh clone of libiio.

$ git remote -v
origin  https://github.com/analogdevicesinc/libiio (fetch)
origin  https://github.com/analogdevicesinc/libiio (push)

$ git describe
v0.7

$ uname -a
Linux WPL-T470 4.4.0-43-Microsoft #1-Microsoft Wed Dec 31 14:42:53 PST 2014 x86_64 x86_64 x86_64 GNU/Linux

pcercuei:
Interesting, I think you are the first one to try libiio on WSL.

For your issue 1, please try the latest libiio from master. The code you patched has changed since then.

For issue 2, we can just hardcode the define, I think, because we know the exact configuration where it works. If you want credit, send a PR with a fix :)

lclausen-adi:
Fixed with #171.

PR #159, opened 2018-03-11
Add check for Microsoft Windows Subsystem for Linux

jaredd:
These changes resolve issue #158 .  The current master branch had already solved libiio hanging due to receive errors (bug 1 mentioned in the referenced issue) while this PR handles disabling the Linux specific code for WSL which seems to have a problem with the MSG_TRUNC flag.

### DIFF
--- a/network.c
+++ b/network.c
@@ -86,6 +86,7 @@ struct iio_context_pdata {
 	struct addrinfo *addrinfo;
 	struct iio_mutex *lock;
 	struct iiod_client *iiod_client;
+	bool msg_trunc_supported;
 };
 
 struct iio_device_pdata {
@@ -1349,7 +1350,10 @@ static ssize_t network_read_line(struct iio_context_pdata *pdata,
 
 		/* Advance the read offset to the byte following the \n if
 		 * found, or after the last charater read otherwise */
-		ret = network_recv(io_ctx, NULL, to_trunc, MSG_TRUNC);
+		if (pdata->msg_trunc_supported)
+			ret = network_recv(io_ctx, NULL, to_trunc, MSG_TRUNC);
+		else
+			ret = network_recv(io_ctx, dst - ret, to_trunc, 0);
 		if (ret < 0)
 			return ret;
 
@@ -1385,6 +1389,12 @@ static const struct iiod_client_ops network_iiod_client_ops = {
 	.read = network_read_data,
 	.read_line = network_read_line,
 };
+
+static bool msg_trunc_supported(struct iio_network_io_context *io_ctx)
+{
+	int ret = network_recv(io_ctx, NULL, 0, MSG_TRUNC);
+	return ret != -EFAULT && ret != -EINVAL;
+}
 
 struct iio_context * network_create_context(const char *host)
 {
@@ -1460,6 +1470,7 @@ struct iio_context * network_create_context(const char *host)
 	pdata->io_ctx.fd = fd;
 	pdata->addrinfo = res;
 	pdata->io_ctx.timeout_ms = DEFAULT_TIMEOUT_MS;
+	pdata->msg_trunc_supported = true;
 
 	pdata->lock = iio_mutex_create();
 	if (!pdata->lock) {
@@ -1468,7 +1479,15 @@ struct iio_context * network_create_context(const char *host)
 	}
 
 	pdata->iiod_client = iiod_client_new(pdata, pdata->lock,
-			&network_iiod_client_ops);
+                       &network_iiod_client_ops);
+
+	if (msg_trunc_supported (&pdata->io_ctx)) {
+		DEBUG("MSG_TRUNC is supported\n");
+	} else {
+		DEBUG("MSG_TRUNC is NOT supported\n");
+		pdata->msg_trunc_supported = false;
+	}
+
 	if (!pdata->iiod_client)
 		goto err_destroy_mutex;
 


commodo:
Change looks good.

Can you add a description in the commit comment ?
And a `Signed-off-by` tag ?

pcercuei:
I don't think the CMake part is necesary, if running on WSL the __WSL__ macro is already automatically defined, this seems redundant.

lclausen-adi:
I wonder do we need a way to detect this at runtime? Can native Linux binaries build on Linux run under WSL?

lclausen-adi:
I guess my question is what happens when you use MSG_TRUNC on WSL? Does it return an error or does it just result in undefined behavior?

rgetz:
My understanding is that it fails/returns error code. 

Chrome is doing similar things:
https://gist.github.com/therealkenc/94e0757c80113bd89a8bc22410d308be

(Hacks in code to make up for incompatibilities). 

https://github.com/Microsoft/WSL/issues/2232

Indicates : "MSG_TRUNC is supported as far as I know. "

I dont have a WSL machine to test things on. Now that appveour supports Windows Server 2016 - we should be able to increase the build matrix to include that (I think). 
https://www.appveyor.com/docs/build-environment/

jaredd:
@lclausen-adi From memory I believe the MSG_TRUNC was returning an error, EINVAL I believe.  
@pcercuei can you elaborate on the WSL macro being predefined?  I tried a short test program:

```c++
#include <iostream>

int main()
{
#ifdef WSL
  std::cout << "WSL defined as " << WSL << std::endl;
#else
  std::cout << "WSL NOT defined" << std::endl;
#endif
  return 0;
}
```
```
$ make test_wsl && ./test_wsl
g++     test_wsl.cc   -o test_wsl
WSL NOT defined
```

I'd be happy to just use a pre-defined macro and not mod the CMakeList.txt.

pcercuei:
@jaredd: I meant ```__WSL__``` but github ate my underscores :p

jaredd:
Am I using the wrong compiler?  Or maybe I have the wrong WSL build?  

$ ver
Microsoft Windows [Version 10.0.16299.248]

$ cat test_wsl.cc
```c++
#include <iostream>

int main()
{
#ifdef __WSL__
  std::cout << "WSL defined as " << __WSL__ << std::endl;
#else
  std::cout << "WSL NOT defined" << std::endl;
#endif
  return 0;
}
```
$ make test_wsl && ./test_wsl
g++     test_wsl.cc   -o test_wsl
WSL NOT defined

pcercuei:
@jaredd: well in the issue report you linked to a patch that used that macro, so I assumed it was available...

Now I agree with Lars, we should handle this at runtime, because from what I understood WSL can install whole distributions like Ubuntu and Fedora, with pre-compiled packages.

lclausen-adi:
Maybe they just check that buf is != NULL and if it is return NULL. @jaredd can you see what happens when you use a dummy buffer for the MSG_TRUNC call?

lclausen-adi:
Or something like:
```
ret = network_recv(io_ctx, NULL, to_trunc, MSG_TRUNC);
if (ret == -EINVAL)
		ret = network_recv(io_ctx, dst, to_trunc, 0);

**Comment by jaredd** [comment]
You guys probably know better, but could this be something simple to check with a toy test program in cmake and set a macro?  Seems a waste to try the call again and again at runtime since if it fails once, it will always fail.

**Comment by lclausen-adi** [comment]
The problem is that we do not know if we run on WSL or not when we compile the library. You can just apt-get install standard Linux packages.

**Comment by jaredd** [comment]
I see, you guys want to be able to install binaries through apt w/o worrying about whether it's WSL or not.  Yeah, I try out the suggested fix.

**Comment by lclausen-adi** [comment]
I've setup a WSL environment here. recv() yields a EFAULT when the buffer is NULL, even though MSG_TRUNC is set. Since recv() will never return EFAULT when MSG_TRUNC is set in a compliant implementation I think we should only retry if the return value is EFAULT.

If we want to avoid having to call recv() twice each time we could also add a flag to the io_context struct to indicate that we need to use the WSL workaround.

**Comment by pcercuei** [comment]
If we pass the internal buffer instead of NULL, does it work with MSG_TRUNC? And does it discard the bytes from the input queue without copying them again?

**Comment by lclausen-adi** [comment]
On WSL it returns EINVAL if MSG_TRUNC is set.

**Comment by jaredd** [comment]
You guys already have a level of indirection with the iiod_client_ops, perhaps rather than defining explicitly you could have it set via some call that accounts for WSL being detected.  Original stays the same, the nework_read_line_wsl gets set when some WSL flag is true (determined at startup).

**Comment by lclausen-adi** [comment]
That could work. I guess calling recv() with flags = MSG_TRUNC and len = 0 on the newly connected socket should do nothing on Linux, but would allow us to detect the missing feature on WSL.

I'd rather not explicitly check for WSL (e.g. by reading /proc/version) to make sure MSG_TRUNC is used once it is available in WSL.

File: network.c [hunk 1]
```diff
@@ -1389,6 +1449,18 @@ static const struct iiod_client_ops network_iiod_client_ops = {
 	.read_line = network_read_line,
 };
 
+static const struct iiod_client_ops network_iiod_client_ops_no_trunc = {
+	.write = network_write_data,
+	.read = network_read_data,
+	.read_line = network_read_line_no_trunc,
+};
+
+bool msg_trunc_supported(struct iio_network_io_context *io_ctx)
```
lclausen-adi: Should be static. The function is not used outside this compilation unit.

network.c
```diff
@@ -1470,8 +1542,17 @@ struct iio_context * network_create_context(const char *host)
...
 	}
 
-	pdata->iiod_client = iiod_client_new(pdata, pdata->lock,
-			&network_iiod_client_ops);
+	if (msg_trunc_supported (&pdata->io_ctx)) {
+		DEBUG("MSG_TRUNC is supported\n");
+		pdata->iiod_client = iiod_client_new(pdata, pdata->lock,
+					&network_iiod_client_ops);
+	}
+	else {
```
lclausen-adi: else should be on the same line as }

network.c
```diff
@@ -1389,6 +1449,18 @@ static const struct iiod_client_ops network_iiod_client_ops = {
...
 
+static const struct iiod_client_ops network_iiod_client_ops_no_trunc = {
+	.write = network_write_data,
+	.read = network_read_data,
+	.read_line = network_read_line_no_trunc,
+};
+
+bool msg_trunc_supported(struct iio_network_io_context *io_ctx)
+{
+	int ret = network_recv(io_ctx, NULL, 0, MSG_TRUNC);
```
lclausen-adi: While MSG_TRUNC is defined on other platforms I think only Linux supports MSG_TRUNC as a flag to recv(). Maye we should add a 
```
#ifdef __linux__
  ...
#elif
  return false;
#endif
```
around this.

network.c
```diff
@@ -1323,6 +1323,69 @@ static ssize_t network_read_line(struct iio_context_pdata *pdata,
...
+	if (!found || i == len - 1)
+		return -EIO;
+
+	return (ssize_t) i + 1;
+#endif
+}
+
+/* use for systems that don't support MSG_TRUNC (e.g. Windows Subsystem for Linux) */
+static ssize_t network_read_line_no_trunc (struct iio_context_pdata *pdata,
+		void *io_data, char *dst, size_t len)
```
lclausen-adi: This is a bit too much copy&paste for my taste. If we update this code in the future we have to update the same code in two places. Maybe you can factor this out into a common function that takes a extra do_trunc boolean flag that is set to either true or false. And then let the compiler take care of optimizing this as it seems fit.

pcercuei:
Could you try to do it without network_read_line_no_trunc()? Because that's a lot of duplicated code...

You could for instance store a ```bool trunc_supported``` in the context_pdata structure, initialized to true. Change it to false if the read() fails.

jaredd:
I'll take a look.

jaredd:
Any comments on the latest update?  I've tested this on both WSL and a OpenSUSE linux box.  Seems to work.

lclausen-adi:
Looks good to me.

pcercuei:
Please squash all the commits :)

From what I see the code still does the recv with MSG_PEEK. Is this working? If it is then the data is read twice (in the case MSG_TRUNC isn't supported) which is not optimal.

lclausen-adi:
reading a large buffer twice is probably still faster than the alternative of reading the data char by char.

pcercuei:
Yes, and reading a large buffer once is even faster :)

You could just skip the first read if trunc_supported is false, I guess.

pcercuei:
Also, now that I think of it, this cannot work. This function is used to read a line and not more than that. If you read too much data after the CR character, the data that comes after will be lost. That will completely disrupt the communication with the server. That's why the windows code does read char-by-char, for instance.

pcercuei:
(So that it's clear, my "this cannot work" was for my idea of a single buffer. With MSG_PEEK we can detect the CR char and know exactly how many bytes to read on the second recv call)

lclausen-adi:
@jaredd OK, rebase, squash and then we are good to go.

Thanks for working on this.

lclausen-adi:
Hi @jaredd,

I've tested your changes and notices two issues.
* MSG_TRUNC is not defined on Windows (at least not with the includes that we use), so this doesn't build on native Windows anymore. If fixed this by adding a #ifdef around the function.
* The msg_trunc_supported() adds a initial startup delay since the recv function waits for the timeout. I've fixed this by adding MSG_DONTWAIT.

I've pushed the fixed commit to the https://github.com/analogdevicesinc/libiio/tree/wsl-workaround branch. Let me know if you are OK with those changes and whether things still work properly under WSL.

Thanks.

jaredd:
Changes seem reasonable.  I've tested the wsl-workaround branch on both my wsl install and a Linux box and it seems to work.

lclausen-adi:
Closing since #171 got merged

PR #162, opened 2018-04-06
Prevent out-of-bounds access in read_each_attr()/iio_channel_attr_read_all()

lclausen-adi:
read_each_attr()/iio_channel_attr_read_all() read all device or channel
attributes at once. The way this is implemented the server sends back a
packet with all the attribute values. For each value it also stores the
size of the value.

The current implementation blindly trusts this value that is store in the
received buffer and increments its pointer position accordingly. But the
data is untrusted data from the network and could contain random garbage.

If the length stored in the buffer is large enough to move the read pointer
beyond the buffer this will cause undefined behavior.

To prevent this make sure that the attribute length stored in the received
buffer is properly validated and read_each_attr()/iio_channel_attr_read_all()
do not access their receive data buffer beyond the amount of data that has
been received.

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -744,7 +744,7 @@ int iio_channel_attr_read_all(struct iio_channel *chn,
 			const char *attr, const char *val, size_t len, void *d),
 		void *data)
 {
-	int ret;
+	int ret, buf_size;
 	char *buf, *ptr;
 	unsigned int i;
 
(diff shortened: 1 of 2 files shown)


commodo:
i guess we can ignore the appveyor build;
will be handled via #163

pcercuei:
Looks OK to me.

Issue #98, opened 2017-05-02
man pages for applications in ./tests


rgetz:
should we build man pages for iio_info, iio_readdev, and iio_reg?

or just put things in the wiki? (and the man pages point to the wiki?)

-Robin

pcercuei:
It would be great to have man pages, yes.

rgetz:
What should be the master format?

http://asciidoc.org/a2x.1.html
seems to be the way many other cmake projects do it.

(rather than just creating actual troff files).

?

pcercuei:
a2x uses asciidoc, which is itself a 1.0 GiB download size / 1.5 GiB install size on my Debian if I try to apt-get it...
I'm more in favour of not adding dependencies, so I'd prefer actual TROFF files.

rgetz:
OK - closing this, then, expect something tomorrow.

PR #174, opened 2018-05-21
Doc update

rgetz:
This will just start the discussion on man pages, which have been missing from the start. Feel free to comment.

The libiio.3 man page is created mostly from a script which parses the header. Doing it this way, rather than  since the discussion #98 suggested to do it this way.

I understand that this will be up for review/comments for awhile.

-Robin


lclausen-adi:
Can you split this into two separate pull requests. One documentation updates, which we should do first, then as a second step adding the man pages.

rgetz:
@commodo : please don't delete branches that aren't yours. I'm still working (slowly) on this.

commodo:
Sorry about that.
I misinterpreted the status of that branch.

I typically go through merged branches and branches that seem abandoned to unclutter the repos.

It happens quite often that these branches pile up, making the repos look dirty, and hard to find things (after a certain branch count).


Maybe, we could switch to a forking method of working with Github.
That way, each person is responsible for (un)cluttering his/her own repo.

The alternative (to keeping branches in-check) is to have someone police-ing merged or (seemingly) abandoned branches, which doesn’t look like a robust way of doing things.

PR #175, opened 2018-05-24
local: Sort devices, channels and attributes

rgetz:
when creating internal structures, sort them. In this way, other iio utilities
	will output information in an easy to read/compare way.

	This has no functionality difference, except to add a small
	negative performance, but it makes the output of iio_attr and
	iio_info, much easier to track by hand.

### DIFF
--- a/local.c
+++ b/local.c
@@ -18,6 +18,7 @@
 
 #include "debug.h"
 #include "iio-private.h"
+#include "sort.h"
 
 #include <dirent.h>
 #include <errno.h>
@@ -1359,7 +1360,9 @@ static int add_channel_to_device(struct iio_device *dev,
 
 	channels[dev->nb_channels++] = chn;
 	dev->channels = channels;
-	DEBUG("Added channel \'%s\' to device \'%s\'\n", chn->id, dev->id);
+	DEBUG("Added %s channel \'%s\' to device \'%s\'\n",
+		chn->is_output ? "output" : "input", chn->id, dev->id);
+
 	return 0;
 }
 
@@ -1708,6 +1711,9 @@ static int add_buffer_attributes(struct iio_device *dev, const char *devpath)
 		int ret = foreach_in_dir(dev, buf, false, add_buffer_attr);
 		if (ret < 0)
 			return ret;
+
+		qsort(dev->buffer_attrs, dev->nb_buffer_attrs, sizeof(char *),
+			qsort_iio_buffer_attr);
 	}
 
 	return 0;
@@ -1761,11 +1767,17 @@ static int create_device(void *d, const char *path)
 		free_protected_attrs(chn);
 		if (ret < 0)
 			goto err_free_scan_elements;
+
+		qsort(chn->attrs,  chn->nb_attrs, sizeof(struct iio_channel_attr),
+			qsort_iio_channel_attr);
 	}
 
 	ret = detect_and_move_global_attrs(dev);
 	if (ret < 0)
 		goto err_free_device;
+
+	qsort(dev->attrs,  dev->nb_attrs, sizeof(char *),
+		qsort_iio_device_attr);
 
 	dev->words = (dev->nb_channels + 31) / 32;
 	if (dev->words) {
@@ -1989,6 +2001,9 @@ struct iio_context * local_create_context(void)
 	ret = foreach_in_dir(ctx, "/sys/bus/iio/devices", true, create_device);
 	if (ret < 0)
 		goto err_context_destroy;
+
+	qsort(ctx->devices, ctx->nb_devices, sizeof(struct iio_device *),
+		qsort_iio_device);
 
 	foreach_in_dir(ctx, "/sys/kernel/debug/iio", true, add_debug);
 
--- a/sort.c
+++ b/sort.c
@@ -1,0 +1,87 @@
+/*
+ * libiio - Library for interfacing industrial I/O (IIO) devices
+ *
+ * Copyright (C) 2018 Analog Devices, Inc.
+ * Author: Robin Getz <robin.getz@analog.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * */
+
+#include "iio-private.h"
+#include <string.h>
+
+/* These are a few functions to do sorting for various
+ * iio structures. For more info, see the qsort(3) man page.
+ * If the structures are updated, the sort functions may
+ * need to be updated.
+ *
+ * The actual arguments to these function are "pointers to
+ * pointers to char", but strcmp(3) arguments are "pointers
+ * to char", hence the cast plus dereference
+ */
+
+int qsort_iio_channel(const void *p1, const void *p2)
+{
+	const struct iio_channel *tmp1 = *(struct iio_channel **)p1;
+	const struct iio_channel *tmp2 = *(struct iio_channel **)p2;
+
+	/* make sure buffer enabled channels are first */
+	if (iio_channel_is_scan_element(tmp1) && !iio_channel_is_scan_element(tmp2))
+		return -1;
+	if (!iio_channel_is_scan_element(tmp1) && iio_channel_is_scan_element(tmp2))
+		return 1;
+	/* and sort them by index */
+	if (iio_channel_is_scan_element(tmp1) && iio_channel_is_scan_element(tmp2)){
+		if (iio_channel_get_index(tmp1) > iio_channel_get_index(tmp2))
+			return 1;
+		return -1;
+	}
+	/* otherwise, if the ID is the same, input channels first */
+	if (strcmp(tmp1->id, tmp2->id) == 0)
+		return !iio_channel_is_output(tmp1);
+
+	/* finally by ID */
+	return strcmp(tmp1->id, tmp2->id);
+}
+
+int qsort_iio_channel_attr(const void *p1, const void *p2)
+{
+	const struct iio_channel_attr *tmp1 = (struct iio_channel_attr *)p1;
+	const struct iio_channel_attr *tmp2 = (struct iio_channel_attr *)p2;
+	/* qsort channel attributes by name */
+	return strcmp(tmp1->name, tmp2->name);
+}
+
+int qsort_iio_device(const void *p1, const void *p2)
+{
+	const struct iio_device *tmp1 = *(struct iio_device **)p1;
+	const struct iio_device *tmp2 = *(struct iio_device **)p2;
+	/* qsort devices by ID */
+	return strcmp(tmp1->id, tmp2->id);
+}
+
+int qsort_iio_device_attr(const void *p1, const void *p2)
+{
+	const char *tmp1 = *(const char **)p1;
+	const char *tmp2 = *(const char **)p2;
+	/* qsort device attributes by name */
+	return strcmp(tmp1, tmp2);
+}
+
+int qsort_iio_buffer_attr(const void *p1, const void *p2)
+{
+	const char *tmp1 = *(const char **)p1;
+	const char *tmp2 = *(const char **)p2;
+	/* qsort buffer attributes by name */
+	return strcmp(tmp1, tmp2);
+}
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ if (ENABLE_IPV6)
 	endif()
 endif()
 
-set(LIBIIO_CFILES backend.c channel.c device.c context.c buffer.c utilities.c scan.c)
+set(LIBIIO_CFILES backend.c channel.c device.c context.c buffer.c utilities.c scan.c sort.c)
 set(LIBIIO_HEADERS iio.h)
 
 add_definitions(-D_POSIX_C_SOURCE=200809L -D__XSI_VISIBLE=500 -DLIBIIO_EXPORTS=1)
--- a/context.c
+++ b/context.c
@@ -19,6 +19,7 @@
 #include "debug.h"
 #include "iio-config.h"
 #include "iio-private.h"
+#include "sort.h"
 
 #include <errno.h>
 #include <string.h>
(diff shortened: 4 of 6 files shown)


local.c
```diff
@@ -1183,7 +1183,7 @@ static int read_device_name(struct iio_device *dev)
```
commodo: nitpick: comment title is a bit long; it could be trimmed to `local: Sort devices, channels and attributes when adding them`

git-style comments prefer an empty between title & description

local.c
```diff
@@ -1340,6 +1351,18 @@ static int add_attr_to_channel(struct iio_channel *chn,
 	attrs[chn->nb_attrs++].name = name;
 	chn->attrs = attrs;
 	DEBUG("Added attr \'%s\' to channel \'%s\'\n", name, chn->id);
+
+	for (i = chn->nb_attrs - 1; i > 0;  i--) {
+		if (strcmp(attrs[i].name, attrs[i-1].name) < 0) {
+			tmp1 = attrs[i].filename;
```
commodo: for this case particularly, I'd use a `struct iio_channel_attr tmp;` and memcpy() stuff onto it and back with `sizeof(tmp)`
it's one variable, and can handle future cases [new members on the `struct iio_channel_attr ` ] ;

commodo: as a continuation here [and while thinking about it a bit], an idea would be to do a search for the index/position where the new element could be added;
then after the `realloc` memmove() everything +1 position to the right;

so, something like:
```
int insert_idx = 0;
for (i = 0; i < chn->nb_attrs - 1;  i--) {
	if (strcmp(attrs[i].name, attrs[i+1].name) >= 0)
		break
}
insert_idx = i;  /* or just use `i` directly */

attrs = realloc(chn->attrs, (1 + chn->nb_attrs) * sizeof(struct iio_channel_attr));

memmove(attrs[i+1], attrs[i], chn->nb_attrs - i);
chn->nb_attrs++;
```

i don't have a strong preference about doing it either way;

local.c
commodo: ah; and the description shouldn't be indented ;

pcercuei:
That would break the ABI, kind of - for applications that don't care about names and hardcode opening device #1 channel #2. Although if they do that they probably deserve it ;)

rgetz:
@pcercuei - yes - we have actually seen this happen when adding features to drivers, or adding drivers to a platform - things get exposed in a different order, and anything that is hard coded breaks. I don't see this as any different. If you want me to add something to the doc that says explicitly that order is not guaranteed - I can do that.

lclausen-adi:
How about using [qsort](https://linux.die.net/man/3/qsort)?

commodo:
I was also thinking about qsort ; I'm curious if qsort is available outside of glibc [GNU context]

lclausen-adi:
"The qsort() function conforms to SVr4, 4.3BSD, C89, C99."

commodo:
Cool :) 
Thanks for the info

rgetz:
OK - I will update things to use qsort. (I checked, and like Lars indicated it is in glibc, uClibc, and bionic). It's missing from many embedded C libraries, but I don't think that is a problem - people using that should be using libtinyiiod (at https://github.com/analogdevicesinc/libtinyiiod )

Thanks for feedback
-Robin

rgetz:
Ok - better late than never - I wanted to have this reviewed before we get together next week, since it helps point out some issues I was looking at/puzzling about..

Updated to use qsort, let me know if there are other issues.

local.c
```diff
@@ -1181,6 +1181,14 @@ static int read_device_name(struct iio_device *dev)
 		return 0;
 }
 
+static int cmp_iio_device_attr(const void *p1, const void *p2)
+{
+	const char *tmp1 = *(const char **)p1;
+	const char *tmp2 = *(const char **)p2;
+
+	return strcmp(tmp1, tmp2);
```
commodo: is there a MAX value/definition for these so that strncmp() can be used ?

rgetz: I don't think so - this is the max path that the kernel knows about. @larsclausen or @pcercuei would know for sure.

pcercuei:
Channel re-ordering is already done here: https://github.com/analogdevicesinc/libiio/blob/master/context.c#L233
Your code with qsort is much better, but could you move it there? iio_context_init( ) is called independently of the backend used, so it would sort the channels/devices/attrs even if using the network backend on an old IIOD, for instance.

local.c
pcercuei: These strings will always be NULL-terminated, so strcmp() is fine.

pcercuei:
Note: the current reorder_channels( ) function sorts the channels by their index when applicable, I think that should take precedence over their ID.

mhennerich:
Given Paul's feedback, I don't think this is ready yet.

local.c
```diff
@@ -1575,6 +1618,14 @@ static int detect_and_move_global_attrs(struct iio_device *dev)
 	return 0;
 }
 
+static int cmp_iio_buffer_attr(const void *p1, const void *p2)
```
lclausen-adi: Would be good to re-use the same function for all which just do strcmp().

local.c
```diff
@@ -1181,6 +1181,14 @@ static int read_device_name(struct iio_device *dev)
 		return 0;
 }
 
+static int cmp_iio_device_attr(const void *p1, const void *p2)
```
lclausen-adi: Most functions in this file are prefixed local_, would be good to continue this pattern.

E.g. local_compare_device_attr().

commodo:
ah ; sorry for the rush
will wait

local.c
rgetz: I wasn't sure if I should do that - in case the structures changed later...

-Robin

rgetz:
I think I have all @larsclausen  requests in. 

@pcercuei : still need to move things to the context.c - wanted to chat about how to sort channels first. when sorting by ID, you get:

root@analog:/usr/local/src/libiio-test/build# iio_attr -a -c .
Using auto-detected IIO context at URI "local:"
dev 'ad7291', channel 'temp0' (input), found 3 channel-specific attributes
dev 'ad7291', channel 'voltage0' (input), found 2 channel-specific attributes
dev 'ad7291', channel 'voltage1' (input), found 2 channel-specific attributes
dev 'ad7291', channel 'voltage2' (input), found 2 channel-specific attributes
dev 'ad7291', channel 'voltage3' (input), found 2 channel-specific attributes
dev 'ad7291', channel 'voltage4' (input), found 2 channel-specific attributes
dev 'ad7291', channel 'voltage5' (input), found 2 channel-specific attributes
dev 'ad7291', channel 'voltage6' (input), found 2 channel-specific attributes
dev 'ad7291', channel 'voltage7' (input), found 2 channel-specific attributes

isn't that better than random voltagexxx numbers? (which is what happens today, when you sort it by index).

?

pcercuei:
It looks better, maybe. But I'm pretty sure some other part of libiio or iiod expect the channels to be sorted by index, so we have to be very cautious here.

rgetz:
Ok - I had a brief look, but didn't see anything - but you know the code base better than I - I will change it /update it tomorrow. Any other comments?

rgetz:
And any objections if I move the qsort callback/comparison functions to iio-private.h? It might be more obvious to the reader what is going on if they were defined/included with the structures, and it was just the calls to qsort that were added to local.c

?

rgetz:
Ok - update 3, I think this has Paul's and Lar's requests in.

-Robin

pcercuei:
One function that will break if the order of channels changes, is iio_buffer_foreach_sample( ), it exits the loop as soon as (chn->index < 0) is reached. I guess it would be enough to replace the "break" with a "continue" there.

pcercuei:
Same in iio_buffer_first( ). The function expects the channels to be ordered by index.

rgetz:
I didn't think there should be any channel changes anymore - it should be the same as before. It's exactly the same sorting logic.

inline is only really needed to make sure that you don't get a complication error / unused function (since -Wall turns on -Wunused-function).

-Robin

pcercuei:
Ok, I think I missed an important point, the sorting problem happens when the channels are non-buffer-capable. When they are buffer capable, will the ID always match the index? Is that enforced somewhere?

pcercuei:
Use iio_channel_is_scan_element.

lclausen-adi:
Maybe a stupid question, but why is there be a warning about unused functions if the functions are used?

rgetz:
@lclausen-adi : when I asked about putting them in the header/by the structure a few weeks ago, there were no objections (or acknowledgment) of either way. I just thought from a long term maintenance standpoint - it would be easier to keep the sorting order by the structures. I'm happy to put them any  place, including a separate c file. just let me know what is the preference.

rgetz:
Current Master version of sort:

analog@analog:~$ iio_attr -a -c cf-ad9361-dds-core-lpc
Using auto-detected IIO context at URI "local:"
dev 'cf-ad9361-dds-core-lpc', channel 'voltage0' (output, index: 0, format: le:S16/16>>0), found 3 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'voltage1' (output, index: 1, format: le:S16/16>>0), found 3 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'voltage2' (output, index: 2, format: le:S16/16>>0), found 3 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'voltage3' (output, index: 3, format: le:S16/16>>0), found 3 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage3', id 'TX1_Q_F2' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage1', id 'TX1_I_F2' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage0', id 'TX1_I_F1' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage7', id 'TX2_Q_F2' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage6', id 'TX2_Q_F1' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage5', id 'TX2_I_F2' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage2', id 'TX1_Q_F1' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage4', id 'TX2_I_F1' (output), found 5 channel-specific attributes

new code, based on this branch:

analog@analog:~$ iio_attr -a -c cf-ad9361-dds-core-lpc
Using auto-detected IIO context at URI "local:"
dev 'cf-ad9361-dds-core-lpc', channel 'voltage0' (output, index: 0, format: le:S16/16>>0), found 3 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'voltage1' (output, index: 1, format: le:S16/16>>0), found 3 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'voltage2' (output, index: 2, format: le:S16/16>>0), found 3 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'voltage3' (output, index: 3, format: le:S16/16>>0), found 3 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage0', id 'TX1_I_F1' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage1', id 'TX1_I_F2' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage2', id 'TX1_Q_F1' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage3', id 'TX1_Q_F2' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage4', id 'TX2_I_F1' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage5', id 'TX2_I_F2' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage6', id 'TX2_Q_F1' (output), found 5 channel-specific attributes
dev 'cf-ad9361-dds-core-lpc', channel 'altvoltage7', id 'TX2_Q_F2' (output), found 5 channel-specific attributes

which I think is OK? (buffers first, index matches the element).

pcercuei:
Looks good.

Could you move the qsort* functions to a qsort.c file, and their prototypes to qsort.h?

rgetz:
Ok - done. and refactored the channel sorting to ensure it's done by index when buffer capable.

Let me know if there are any other tweaks necessary.

rgetz:
@pcercuei : sorry for the copy/paste error on your name. I took it out, since like you said - your email isn't valid (which is something else - did you want me to set those to something valid now? - there are 46 instances of the old email).

@larsclausen : any other comments/fixes?

So, this is Ok to merge now? 

-Robin

lclausen-adi:
I'd call the file `sort.c`. The fact that qsort is used as the sorting function is a implementation detail. Any other function could do the same thing.

rgetz:
updated.

pcercuei:
Looks good to me.

mhennerich:
LGTM let's merge things now.

mhennerich:
LGTM let's merge things now.

mhennerich:
LGTM let's merge things now

mhennerich:
LGTM let's merge things now

rgetz:
@mhennerich : Did you want to revert the commit, and have me change the function names like Lars suggested?

lclausen-adi:
@rgetz  Follow up patch please. But wait a day or two. github is a bit broken today, I was not able to review things in full.

rgetz:
Ok - will do. I also just noticed an issue I will fix at the same time. (channel attributes aren't always sorted properly for some reason).

rgetz:
@lclausen-adi : Any other comments before I send the next patch?

local.c
```diff
@@ -1761,12 +1767,18 @@ static int create_device(void *d, const char *path)
 		free_protected_attrs(chn);
 		if (ret < 0)
 			goto err_free_scan_elements;
+
+		qsort(chn->attrs,  chn->nb_attrs, sizeof(struct iio_channel_attr),
```
lclausen-adi: Extra space here after `chan->attrs,`

local.c
```diff
@@ -1761,12 +1767,18 @@ static int create_device(void *d, const char *path)
...
+			qsort_iio_channel_attr);
 	}
 
 	ret = detect_and_move_global_attrs(dev);
 	if (ret < 0)
 		goto err_free_device;
 
+	qsort(dev->attrs,  dev->nb_attrs, sizeof(char *),
```
lclausen-adi: Here as well

sort.c
```diff
@@ -0,0 +1,87 @@
...
+	/* make sure buffer enabled channels are first */
+	if (iio_channel_is_scan_element(tmp1) && !iio_channel_is_scan_element(tmp2))
+		return -1;
+	if (!iio_channel_is_scan_element(tmp1) && iio_channel_is_scan_element(tmp2))
+		return 1;
+	/* and sort them by index */
+	if (iio_channel_is_scan_element(tmp1) && iio_channel_is_scan_element(tmp2)){
+		if (iio_channel_get_index(tmp1) > iio_channel_get_index(tmp2))
+			return 1;
+		return -1;
```
lclausen-adi: I think we want to fallback to id if the index is the same.

PR #183, opened 2018-07-25
appveyor.yml: change versioning to '{branch}.{build}' format

commodo:
The old '2015.1.{build}' format is a bit confusing, because it gives the
impression that something is stuck in the past.

Change this to '{branch}.{build}'.
That way we don't need to care about date/time, since Appveyor will track
that anyway.


commodo:
if no objections will merge this tomorrow

PR #184, opened 2018-07-27
.travis.yml: add host-key algo ssh-dss for xenial deploy

commodo:
The version of OpenSSH included in 16.04 disables ssh-dss.
We need to enable it explicitly for this to work.


commodo:

PR #192, opened 2018-09-19
cmake: Add LD_LIBRARY_PATH to udev rule

jpendlum:
With local installs, libiio.so will not be found when the udev rule is ran due to not having the local lib install dir in the lib path.

### DIFF
--- a/libiio.rules.cmakein
+++ b/libiio.rules.cmakein
@@ -1,1 +1,1 @@
-SUBSYSTEM=="usb", PROGRAM=="/bin/sh -c '@CMAKE_INSTALL_FULL_BINDIR@/iio_info -s | grep %s{idVendor}:%s{idProduct}'", RESULT!="", MODE="666"
+SUBSYSTEM=="usb", PROGRAM=="/bin/sh -c 'LD_LIBRARY_PATH=@CMAKE_INSTALL_FULL_LIBDIR@ @CMAKE_INSTALL_FULL_BINDIR@/iio_info -s | grep %s{idVendor}:%s{idProduct}'", RESULT!="", MODE="666"


lclausen-adi:
I'm not so sure about this. Is there any precedence for using LD_LIBRARY_PATH in udev rules?

commodo:
I don't really know; but if it looks too weird, we could replace this with a shell script [at some point] and wrap some of this logic.

I was a bit unsure about using LD_LIBRARY_PATH here as well.

The way I see it, is that `@CMAKE_INSTALL_FULL_BINDIR@` is already being used as a prefix for `iio_info`, so using `@CMAKE_INSTALL_FULL_LIBDIR@` to explicitly point to libs, shouldn't be a problem [ from my point of view ].

One use-case is when you install libiio in `/usr/local/lib` ; which is not always in the default system library path [though it sounds like it could be, in certain setups].

Also, LD_LIBRARY_PATH is not too hard [from what I can remember]. 
i.e. if there is no library at that path, it will trickle down to other default system library paths.

pcercuei:
Shouldn't the rpath be hardcoded inside the binary?

rgetz:
to provide some context, (and @jpendlum should comment more if i mis-describe anything) - at GR Conference, he was trying to standbox a libiio install in Linux, and the udev rule didn't run (so he didn't have permissions to access the device), since iio_info would error, since it could not find libiio.so

-Robin

rgetz:
I found some info, I think we are missing something like:

set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)

As suggested in the solution:
https://stackoverflow.com/questions/30398238/cmake-rpath-not-working-could-not-find-shared-object-file

This - as Paul suggests, should set the internal rpath, and allow the application to run. 

I need to set up a VM and try installing things into a sandbox to test. Comments appreciated if folks understand cmake in more detail...

rgetz:
Ok - had a look at this today...

When I add the patch:
==========
rgetz@brain:~/github/libiio/build$ git diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index cf61c93..c1e42a4 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,9 @@ if (NOT WIN32 AND CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
        set(CMAKE_INSTALL_PREFIX "/usr" CACHE PATH "default install path" FORCE)
 endif()
 
+set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+
 set(CMAKE_INSTALL_DOCDIR "" CACHE PATH "documentation root (DATAROOTDIR/doc/${PROJECT_NAME}${LIBIIO_VERSION_MAJOR}-doc)")
 include(GNUInstallDirs)
 set(CMAKE_INSTALL_DOCDIR "${CMAKE_INSTALL_DATAROOTDIR}/doc/${PROJECT_NAME}${LIBIIO_VERSION_MAJOR}-doc")
=============

It seems to work...

rgetz@brain:~/github/libiio/build$ cmake -DCMAKE_INSTALL_PREFIX=/home/rgetz/foo
rgetz@brain:~/github/libiio/build$ make && sudo make install

rgetz@brain:~/foo/bin$ ./iio_info -s
Library version: 0.15 (git tag: 72feae0)
No contexts found.

and strace says it is loading the right one:

openat(AT_FDCWD, "/home/rgetz/foo/lib/libiio.so.0", O_RDONLY|O_CLOEXEC) = 3

I can open a new pull request, unless someone has any objections.

-Robin

pcercuei:
```set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_LIBDIR})``` would be better, or ```${CMAKE_INSTALL_FULL_LIBDIR}``` if the full path is indeed required.

jpendlum:
@rgetz Yea that seems like the better approach. I'll close this PR.

PR #195, opened 2018-09-27
CI/travis: make shell scripts more pendantic

commodo:
As scripts grow in size, it's a good idea to exit & fail them on the first non-zero exit code.

That also enforces us to re-think the code, and not ignore any non-successful parts of the CI scripts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 obj-*
 *.deb
 
+*.swp
+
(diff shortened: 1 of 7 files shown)


commodo:
Will merge this later today if no objections.

commodo:
my bad;
i deleted the branch on my fork
re-opening

PR #196, opened 2018-09-28
build,CI/travis: setup CentOS testing

commodo:
Adapted from blog-post:
     https://derekweitzel.com/2016/05/03/building-centos-packages-on-travisci/
    
This essentially uses CentOS docker images to run the libiio build and
create RPM packages, which is a bit closer to the real-world case of RPM
usage.
    
We're not doing any deployments of these files just yet.
This just builds them and confirms that we can build and install on
CentOS, and offers the procedure to do it.
A later change-set will add deployment of CentOS RPMs, since these seem to
be interesting for people.

### DIFF
--- a/cmake/LinuxPackaging.cmake
+++ b/cmake/LinuxPackaging.cmake
@@ -100,7 +100,7 @@ if(DEB_DETECT_DEPENDENCIES AND DPKG_CMD AND DPKGQ_CMD)
 		${CPACK_DEBIAN_PACKAGE_DEPENDS})
 else()
 	# assume everything is turned on, and running on a modern OS
-	set(CPACK_DEBIAN_PACKAGE_DEPENDS "libaio-dev (>= 0.3.109), libavahi-client-dev (>= 0.6.31), libavahi-common-dev (>= 0.6.31), libc6-dev (>= 2.19), libusb-1.0-0-dev (>= 2:1.0.17), libxml2-dev (>= 2.9.1)")
+	set(CPACK_DEBIAN_PACKAGE_DEPENDS "libaio (>= 0.3.109), libavahi-client (>= 0.6.31), libavahi-common (>= 0.6.31), libc6 (>= 2.19), libusb-1.0-0 (>= 2:1.0.17), libxml2 (>= 2.9.1)")
 	message(STATUS "Using default dependencies for packaging")
 endif()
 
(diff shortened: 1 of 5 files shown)


cmake/LinuxPackaging.cmake
```diff
@@ -100,7 +100,7 @@ if(DEB_DETECT_DEPENDENCIES AND DPKG_CMD AND DPKGQ_CMD)
 		${CPACK_DEBIAN_PACKAGE_DEPENDS})
 else()
 	# assume everything is turned on, and running on a modern OS
-	set(CPACK_DEBIAN_PACKAGE_DEPENDS "libaio-dev (>= 0.3.109), libavahi-client-dev (>= 0.6.31), libavahi-common-dev (>= 0.6.31), libc6-dev (>= 2.19), libusb-1.0-0-dev (>= 2:1.0.17), libxml2-dev (>= 2.9.1)")
+	set(CPACK_DEBIAN_PACKAGE_DEPENDS "libaio (>= 0.3.109), libavahi-client (>= 0.6.31), libavahi-common (>= 0.6.31), libc6 (>= 2.19), libusb-1.0-0 (>= 2:1.0.17), libxml2 (>= 2.9.1)")
```
lclausen-adi: This is obviously a follow up change, but can this be adopted so that there is only a dependency on those packages for which features were enabled? E.g. if libiio is build without usb backend no libusb dependency.

lclausen-adi:
Does this run for every commit or just releases?

commodo:
every commit

cmake/LinuxPackaging.cmake
commodo: so, `CPACK_DEBIAN_PACKAGE_DEPENDS` is already being dynamically created on the `if()` branch ; this is the `else()` branch where things are added as fall-back;

for RPMs (and CentOS RPMs specifically), we surely could also do this; i just decided to go with a static list for now, and take a package list from CentOS 6, since those are currently the oldest versions that are being used for build;

lclausen-adi: Hm, but we should always know which backends are enabled.

commodo: That's true.
tbh, I don't know why this has been moved under a `DEB_DETECT_DEPENDENCIES` external variable (with default OFF).
We could remove it and always have that behavior.

No opinion

lclausen-adi: Yes

rgetz: The DEB_DETECT_DEPENDENCIES is there to detect dependencies when others build and release binaries. We are not the only ones releasing debs.

commodo: There may be more context-info that I am missing here about the `DEB_DETECT_DEPENDENCIES` part. So, I propose to defer this discussion on a PR that deals with this specifically.

commodo:
If no objections, will merge this today or tomorrow.

If there are topics to discuss (other enhancements, or ways to approach dependency detection), let's handle them in follow-up series.

PR #198, opened 2018-09-28
Fix bug in mask bit tests

dpfrey:
Previously, the index field of struct iio_channel was being used to test
against the buffer field of struct iio_device and struct iio_buffer.
This is incorrect because the index of the first channel may be non-zero
and the indexes may not be contiguous. For further evidence, look at
iio_channel_enable/disable() which use number rather than index.

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -223,14 +223,14 @@ ssize_t iio_buffer_foreach_sample(struct iio_buffer *buffer,
 				break;
 
 			/* Test if the buffer has samples for this channel */
-			if (!TEST_BIT(buffer->mask, chn->index))
+			if (!TEST_BIT(buffer->mask, chn->number))
 				continue;
 
 			if ((ptr - start) % length)
 				ptr += length - ((ptr - start) % length);
 
 			/* Test if the client wants samples from this channel */
-			if (TEST_BIT(dev->mask, chn->index)) {
+			if (TEST_BIT(dev->mask, chn->number)) {
 				ssize_t ret = callback(chn,
 						(void *) ptr, length, d);
 				if (ret < 0)
@@ -272,7 +272,7 @@ void * iio_buffer_first(const struct iio_buffer *buffer,
 			break;
 
 		/* Test if the buffer has samples for this channel */
-		if (!TEST_BIT(buffer->mask, cur->index))
+		if (!TEST_BIT(buffer->mask, cur->number))
 			continue;
 
 		/* Two channels with the same index use the same samples */


pcercuei:
The channels are sorted by indexes so the first one should be zero and they should be contiguous, unless I'm missing something. What's your setup?

dpfrey:
If you look at `drivers/iio/imu/bmi160/bmi160_core.c`, you will see that it declares `enum bmi160_scan_axis` and values 0 through 3 are occupied by definitions which correspond to a bmm150 magnetometer downstream of the bmi160.  The driver doesn't currently support the bmm150.  So when the driver declares an array of type `struct iio_chan_spec`, the first element in the array has `scan_index == BMI160_SCAN_ACCEL_X`.  `BMI160_SCAN_ACCEL_X` has a value of 4.  As for why the indicies would be discontiguous... I don't know.  I just can't find any documentation in the kernel that indicates that they *must* be contiguous.

This is the only documentation I can find:

> Here scan_index defines the order in which the enabled channels are placed inside the buffer. Channels with a lower scan_index will be placed before channels with a higher index. Each channel needs to have a unique scan_index.
> 
> Setting scan_index to -1 can be used to indicate that the specific channel does not support buffered capture. In this case no entries will be created for the channel in the scan_elements directory

lclausen-adi:
@pcercuei https://github.com/analogdevicesinc/libiio/commit/4792fa3a005165597afeed3700831302c256d1d6

lclausen-adi:
Do we need the same fix in iiod/ops.c when server side demuxing is used?

dpfrey:
@lclausen-adi: Are you referring to the `send_sample()` and `receive_sample()` functions?  They definitely look suspicious.  I didn't do an exhaustive search through the codebase for similar defects.  I made the change in `iio_buffer_foreach_sample()` because I was blocked from reading a sensor properly by that issue.  I made the change in `iio_buffer_first()` because I happened to notice it while making the first change.

lclausen-adi:
@dpfrey Yes. I think those also need to be fixed. But I'd like @pcercuei to confirm this.

Either way your patch looks good. Thanks for fixing this issue.

pcercuei:
I *think* to make it work properly, the channel mask within the iio_buffer should be a bitmask of the indices, and the mask within the iio_device should be a bitmask of the channel numbers.

lclausen-adi:
I don't see why the code in iio_buffer_foreach_sample() should match the code in iio_device_get_sample_size_mask(). Both in the end compute the same thing, whereas iio_buffer_foreach_sample() just does a few intermediary stops.

lclausen-adi:
Which reveals another bug in iio_buffer_foreach_sample(). It does not take the repeat property of the channel into account.

We should factor that code into a common helper function.

pcercuei:
Speaking about iio_device_get_sample_size_mask, it's the only blocker that prevents us to compile IIOD with the public libiio API. It would be nice to decouple them.

pcercuei:
@dpfrey: Could you fix it also in iio_buffer_foreach_sample, and (in a separate commit) in IIOD's send_sample/receive_sample? Thanks!

mhennerich:
please see #211 - this is untested

commodo:
Will merge this later today if no objections together with #211

commodo:
[LGTM](https://i.pinimg.com/564x/df/4b/83/df4b8303fe2b0c3256bf941faeec5ca4.jpg)

pcercuei:
There is an objection :)
Please add the same fix to iio_buffer_foreach_sample.

mhennerich:
> There is an objection :)
> Please add the same fix to iio_buffer_foreach_sample.

I added a fix for this to this PR:
https://github.com/analogdevicesinc/libiio/pull/211

PR #201, opened 2018-10-12
ML bindings FIR load fix

tfcollins:
This change to the MATLAB bindings actually enables the FIR after loading it. This is how the current examples expect the functionality of the writeFirData function to work.

Signed-off-by: Travis Collins <travis.collins@analog.com>

### DIFF
--- a/bindings/matlab/iio_sys_obj_matlab.m
+++ b/bindings/matlab/iio_sys_obj_matlab.m
@@ -333,7 +333,15 @@ classdef iio_sys_obj_matlab
         
         function ret = writeFirData(obj, fir_data_file)
             fir_data_str = fileread(fir_data_file);
-            ret = writeAttributeString(obj.libiio_ctrl_dev, 'filter_fir_config', fir_data_str);
+            % Force sample rate to a value not requiring a FIR to prevent loading errors,
+            % FIR profile will update the desired sample rate once enabled
+            r0 = writeAttributeString(obj.libiio_ctrl_dev, 'in_voltage_sampling_frequency', '3000000');
+            r1 = writeAttributeString(obj.libiio_ctrl_dev, 'in_out_voltage_filter_fir_en', '0');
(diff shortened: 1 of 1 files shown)


commodo:
[LGTM](https://i.pinimg.com/originals/95/3d/73/953d73b50424e60868684017435f365f.jpg)

rgetz:
@mhennerich  - can you have a look at this - is this the way the FIR enable attribute works? (it will automatically set the data rates?)

tfcollins:
This is borrowed from what we do in libad9361.

mhennerich:
@tfcollins @rgetz -
This is not necessary in case the filter file has the rates sets!
The workaround we implemented here: https://github.com/analogdevicesinc/libad9361-iio/commit/b2e88caf562bfac2daa45f9040f7e9f861f5e99a

Was only necessary since the filter config did NOT include the rates AND in this case the driver will keep the current baseband rate, which might NOT necessarily work for the loaded filter.

tfcollins:
So Michael should we assume the filter files have those rates? There are some old filters that don't do this (https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/filters/GSM.ftr), which I would argue we should probably not ship anymore anyway.

But in regards to this change, you would just want to remove the explicit sample rate change?

mhennerich:
The files without the rates are legacy. They have been around long before we started the Filter Wizard.
I think we can easily add the rates to them.
The workaround was initially implemented for the auto rate functions. If we add the rates there we can drop the workaround. Whatever seems easier.

-Michael

tfcollins:
Sounds reasonable. I think we can extrapolate the rate setting from the related ini's and dec/int in the existing files.

I'll remove that sample rate update call on this PR.

tfcollins:
Removed that sample rate update as requested. @commodo please squash and merge if @mhennerich approves.

PR #202, opened 2018-10-18
appveyor.yml: Downgrade curl in order to fix the appveyor builds.

AlexandraTrifan:
It seems that the latest version provided by appveyor (7.61.1-3) breaks some dependencies. Use the last
version which was known to work (7.61.1-2) until things get fixed.


rgetz:
Thanks - this looks good to me.

"Until things get fixed" - do we have a note, saying to check things before the next release, or when\how will that happen?

AlexandraTrifan:
I didn't find anything in the changelog that might break the package.  Might be some dependencies that are not compatible or linked correctly. So I think we need to check for appveyor/msys updates. The next curl release is scheduled for 31st of October, so the appveyor/msys packages might update again and we can test whether the issue still appears or not.

commodo:
To add here: the 2 curl packages seem to be the same content-wise, meaning that some library/dependency underneath changed.
[Generally speaking] It's not worth digging too deep into it, since the issue is not our fault.

[LGTM](https://static.boredpanda.com/blog/wp-content/uploads/2018/08/funny-coworkers-69-5b7ec12961df5__605.jpg)

commodo:
Will merge later today if no objections

PR #203, opened 2018-10-24
minor sorting cleanups

rgetz:
- change functions names to more descriptive iio_*_compare
- move attribute sorting into context.c
- when sorting iio_channel, if index is the same, use ID
- fix whitespace issues that were accidently committed


rgetz:
This is a followup to #175

rgetz:
Ok - will split things.

rgetz:
split. Let me know if there are other things.

PR #205, opened 2018-10-25
usb: Increase ctrl pipe timeout

lclausen-adi:
Experimental testing shows that with the current timeout if the remote system
is under high load the timeout is triggered occasionally.

To mitigate this issue increase the timeout.

### DIFF
--- a/usb.c
+++ b/usb.c
@@ -150,7 +150,7 @@ static unsigned int usb_calculate_remote_timeout(unsigned int timeout)
 	return timeout / 2;
 }
 
-#define USB_PIPE_CTRL_TIMEOUT 200 /* These should not take long */
+#define USB_PIPE_CTRL_TIMEOUT 1000 /* These should not take long */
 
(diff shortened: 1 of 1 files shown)


pcercuei:
That doesn't really fix the problem, does it? How about giving iiod real-time rights?

lclausen-adi:
It fixes the problem. Higher priory for IIOD would be a orthogonal solution.

pcercuei:
It does not fix it, it only makes it less likely...

lclausen-adi:
So does running iiod with RT priority ;)

The patch makes it so unlikely that it will only happen once in 100 years instead of once in 10 minutes. That's good enough for me.

commodo:
[LGTM](http://sixgun.org/files/patchin.jpg)

commodo:
I'm unsure here if the discussion has been resolved.
If not, please notify me.

If it has been resolved, then I'll merge this tomorrow

PR #206, opened 2018-10-25
Ensure iio_info can find a locally installed libiio.so

rgetz:
When doing a sandbox/local install:
  cmake -DCMAKE_INSTALL_PREFIX=/home/rgetz/foo
the iio_info instance that is executed as part of the udev rule will not be able to find the libiio, and it will fail to change permissions properly. Hard coding the RPATH ensures that the library can be found.

before this change:
rgetz@brain:~/foo/bin$ ./iio_info -s
./iio_info: error while loading shared libraries: libiio.so.0: cannot open shared object file: No such file or directory

after this change:
./iio_info -s
Library version: 0.15 (git tag: 72feae0)
Available contexts:
	0: 0456:b673 (Analog Devices Inc. PlutoSDR (ADALM-PLUTO)), serial=10440004278a0007f2ff1400ed95ed41cd [usb:1.6.5]

and strace indicates the proper library is being used.
openat(AT_FDCWD, "/home/rgetz/foo/lib/libiio.so.0", O_RDONLY|O_CLOEXEC) = 3


rgetz:
This is a followup to #192

rgetz:
yes - it works for all the tools, it just that iio_info is the one used in the udev rule, and the one that was causing the issue...

Also note - that if you have a standard install, and then install a test in a sandbox, the default search order is to look in the standard LD paths before the hard coded one in the binary, so other than testing the user space apps, it will not run the locally installed library code.

-Robin

PR #207, opened 2018-10-25
usb: add libusb version to context attributes

rgetz:
When debugging, it's always nice to know the libusb version. This makes it easy to find.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,6 +158,7 @@ endif()
 find_library(LIBUSB_LIBRARIES usb-1.0)
 find_path(LIBUSB_INCLUDE_DIR libusb-1.0/libusb.h)
 if (LIBUSB_LIBRARIES AND LIBUSB_INCLUDE_DIR)
+	message(STATUS "Looking for libusb-1.0 : Found")
 	option(WITH_USB_BACKEND "Enable the libusb backend" ON)
 
 	if(WITH_USB_BACKEND)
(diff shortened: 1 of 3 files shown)


rgetz:
Looks like old/ancient versions of libusb don't support this, so I will need to ifdef this based on the libusb version. (It was added 2012-04-17: v1.0.10)

/libiio/usb.c: In function 'usb_populate_context_attrs':
/libiio/usb.c:743: warning: implicit declaration of function 'libusb_get_version'

So I will need to test with a #ifdef LIBUSB_API_VERSION (that was added Jan 8, 2014, replacing the older LIBUSBX_API_VERSION)
https://github.com/libusb/libusb/commit/1eff220474f63d7ea7f8f99bef2a3da9da5324eb
if we want to support the older distributions.

Thoughts?

commodo:
I'd use the same mechanism that's around line 840:
```
#if defined(LIBUSB_API_VERSION) && (LIBUSB_API_VERSION >= 0x01000016)
    libusb_set_auto_detach_kernel_driver(hdl, true);
#endif
```

If we can find the libusb API version that introduced this, it would be neat, and a bit more robust.
Other than that, this looks good.

pcercuei:
You could also check whether the function symbol exists, CMake supports that and IIRC we already do this for another function.

rgetz:
LIBUSB_API_VERSION was added after libusb_get_version. If the define is there - the function exists. 

- Robin

commodo:
Paul's suggestion also sounds nice.

lclausen-adi:
Sounds nice, but is more work to get the same result.

commodo:
Hmm, ok ; no strong opinion from me.
Let's have this as is.

If there aren't any strong opinions for other options, will merge this tomorrow

rgetz:
Ok - I added @pcercuei  suggestion before I saw @lclausen-adi comment - let me know which version I should leave.

-Robin

pcercuei:
Using LIBUSB_API_VERSION is fine with me.

mhennerich:
Either way seems fine to me - let's merge this

rgetz:
thanks

PR #209, opened 2018-10-27
travis : Add the LDIST for centos

rgetz:
Currenty - when travis-ci builds for centos, the LDIST is set to trusty, so things get deployed with the trusty namespace. Fix that.

Signed-off-by: Robin Getz <Robin.Getz@analog.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,10 +54,12 @@ matrix:
       env:
         - OS_TYPE=centos
         - OS_VERSION=6
+        - LDIST=-centos6
     - os: linux
       env:
         - OS_TYPE=centos
         - OS_VERSION=7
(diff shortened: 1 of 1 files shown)


rgetz:
If this works - We need to update the readme.md to point to the new file.

rgetz:
This seems to work, but http://swdownloads.analog.com/cse/travis_builds/master_latest_libiio-centos6.rpm will not show up until this is merged to master, so once this is merged, I will send a new patch for the readme.md, which points to the new artifacts. (so that I can actually test it, which I can't do now).

commodo:
[LGTM](https://i.pinimg.com/564x/9f/48/16/9f48166bb13942291f9bd53392a5220f.jpg)

PR #210, opened 2018-10-29
Update README.md with newest centos packages

rgetz:


commodo:
Should we remove `deb` packages from CentOS  and `rpm` packages from Ubuntu ?
Or leave them ? I admit that supporting these rare combination doesn't hurt, but I'm not sure if it's worth the support cases ; or having people confused and ask questions.

Otherwise looks good.

commodo:
I can take a look [later] at trimming/adapting the build to build `deb` only on Ubuntu and `rpm` only on CentOS.
We could also delegate this; it's not a big task.

rgetz:
I would just leave it for now, We can add something to the README if people ask. I think it's pretty clear (the host OS the package is built on vs the package type). I will see if I can find a "deb" icon, not a "debian" one - I think that is all that is necessary.

How about this one:
http://www.softicons.com/system-icons/hycons-icon-theme-by-gomez-hyuuga/mimetypes-application-x-deb-icon

-Robin

commodo:
I don’t mind whether it’s one icon or the other.

I guess whatever is preferred is fine from my side.

I would be fine with merging this as is.
But if there’s a preference to change the icons we can wait.

rgetz:
The icons are managed in the wiki separately. The readme just points to them.

just clone https://github.com/analogdevicesinc/libiio.wiki.git - and update the files in the img directory.

```
rgetz@brain:~/github/libiio.wiki/img$ file *
deb.png:       PNG image data, 48 x 48, 8-bit/color RGBA, non-interlaced
linux_box.png: PNG image data, 48 x 48, 8-bit/color RGBA, non-interlaced
osx_box.png:   PNG image data, 48 x 48, 8-bit/color RGBA, non-interlaced
rpm.png:       PNG image data, 48 x 48, 8-bit/color RGBA, non-interlaced
win_box.png:   PNG image data, 48 x 48, 8-bit/color RGBA, non-interlaced
```
the README will be updated with the new icons.

-Robin

commodo:
Ah, I did not notice.
Thanks for the info

PR #211, opened 2018-10-30
IIOD: Fix bug in mask bit tests (continued)

mhennerich:
From: #198 

Previously, the index field of struct iio_channel was being used to test
against the buffer field of struct iio_device and struct iio_buffer.
This is incorrect because the index of the first channel may be non-zero
and the indexes may not be contiguous. For further evidence, look at
iio_channel_enable/disable() which use number rather than index.

### DIFF
--- a/iiod/ops.c
+++ b/iiod/ops.c
@@ -362,7 +362,7 @@ static ssize_t send_sample(const struct iio_channel *chn,
 		void *src, size_t length, void *d)
 {
 	struct sample_cb_info *info = d;
-	if (chn->index < 0 || !TEST_BIT(info->mask, chn->index))
+	if (chn->index < 0 || !TEST_BIT(info->mask, chn->number))
 		return 0;
 	if (info->nb_bytes < length)
 		return 0;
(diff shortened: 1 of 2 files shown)


iiod/ops.c
```diff
@@ -560,7 +560,10 @@ static void rw_thd(struct thread_pool *pool, void *d)
 				struct iio_channel *chn = dev->channels[i];
 				long index = chn->index;
 
-				if (index >= 0 && TEST_BIT(entry->mask, i))
+				if (index < 0)
+					continue;
+
+				if (TEST_BIT(entry->mask, i))
```
commodo: curios: should this be `chn->number` instead of `i`?

mhennerich: The assumption is that both are the same! 
But Lars also thinks that chn->number looks more robust.
I'll fix this.

commodo:
[LGTM](https://i.pinimg.com/564x/c7/89/d8/c789d8bbbf8b1ec2821369f5c8f0a58e.jpg)

pcercuei:
Please split the libiio and IIOD changes into two commits.

mhennerich:
> Please split the libiio and IIOD changes into two commits.

ok - Done!

Issue #185, opened 2018-07-27
Wiki: Discuss the assignment of triggers when creating buffers


allsey87:
I suggest that https://analogdevicesinc.github.io/libiio/#create_buffer be updated to mention that a trigger (at least in the case of a triggered buffer) needs to be assigned *before* the buffer can be created. Otherwise the call to `local_write_dev_attr(dev, "buffer/enable", "1", 2, false);` in local.c fails with -22 (EINVAL).

I would also modify `static ssize_t local_enable_buffer(const struct iio_device *dev)` to print an warning/error when local_write_dev_attr fails. For the moment, no error is reported which made this issue difficult to find.

pcercuei:
You're right, the error is ignored and that's not normal.

rgetz:
@allsey87 : I have a patch for master #213 - can you try it out, and see if that would have helped in debugging things?

rgetz:
patch was applied, so closing this.

PR #213, opened 2018-10-30
local: pass errors up the stack

rgetz:
When a enabling a buffer fails (in the case of triggered buffer not having a trigger set up), the error is currently ignored.

Ensure that errors are passed back up the stack for end user debugging.

Signed-off-by: Robin Getz <Robin.Getz@analog.com>

### DIFF
--- a/local.c
+++ b/local.c
@@ -425,7 +425,7 @@ static ssize_t local_enable_buffer(const struct iio_device *dev)
 			pdata->buffer_enabled = true;
 	}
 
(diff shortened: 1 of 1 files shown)


rgetz:
This should fix/resolve #185 

-Robin

commodo:
[LGTM](https://i.pinimg.com/564x/50/23/5d/50235d2dff0d5cca432d3213e3df0341.jpg)

Issue #199, opened 2018-09-30
Ubuntu 12.04 LTS (Precise Pangolin) reached its regular End of Life on April 28, 2017. 


rgetz:
should we just remove it from travis-ci (and it's mostly failing), and add something more modern?

Today Travis-ci supports:
Ubuntu 14.04.5 LTS (Trusty Tahr) : EOL April 2019, which we build now, but April is coming soon.
Ubuntu 12.04.5 LTS (Precise Pangolin) : reached its regular End of Life on April 28, 2017. 

and supports these from a docker environment:
Ubuntu 18.04.1 LTS (Bionic Beaver) : EOL April 2023. 
Ubuntu 16.04.5 LTS (Xenial Xerus) : EOL April 2021

Adding this issue, so we start looking at the docker build environment, and eventually get 16.x & 18.x building.

?

commodo:
libiio already builds for Ubuntu 16.04 
See https://travis-ci.org/analogdevicesinc/libiio/jobs/435412264

For bionic, we would need to add a docker build, since Travis-CI doesn't have bionic in their inventory yet.

Currently, it's in-progress to add CentOS 6 & 7, since there seem to be some questions & interest about building libiio on CentOS. See PR #196

commodo:
Regarding 12.04, I think we could leave it until Travis CI drops it. They do show a banner about it being in the process of being decommissioned, but that looks like it's been there for a few months.

The builds on 12.04 haven't been failing in the last 22 days or so (https://travis-ci.org/analogdevicesinc/libiio/builds ) . They are failing intermittently, but I don't know if we can control that.
Ubuntu 12.04 seems to be one of those old distros that seems to have stuck around for a long time; I guess the best moment to discontinue builds for it, is when Travis-CI no longer supports it.

AlexandraTrifan:
I'll start looking into a docker build for Bionic, since Xenial is already up and running.

rgetz:
Thanks - I didn't see that Xenial was already building, since it wasn't in the releases (I never checked in the README), but that is since we haven't done a release with it yet.

As for the intermittent failures - dropping old things is just part of maintenance. We shouldn't enable bad practices (using not supported OSes), that the vendor isn't providing security updates for. It has been a year plus already...

-Robin

rgetz:
I also noticed that we are seeing:
```
Running builds with Xcode 6.4 in Travis CI is deprecated and will be removed in January 2019. 
[...]

Issue #148, opened 2018-02-14
Add more architectures to downloads


markusr:
It would be great to have some more architectures as Downloads (.deb and .rpm). I think the most important one are for ARM platforms because of the many SBCs (raspberrypi, orangepi) 

- linux armhf
- linux arm64
- linux armel

Also a hint that the current packages for linux are all amd64 would be good.

rgetz:
@markusr - that is a great idea. 

The current packages are built on travis-ci or appveour. Do you have time to make a patch to do this? (Get travis to Install cross compile tools, and build?)

Thanks

markusr:
@rgetz  I will have a look what I can do. Not sure if I will succeed.

Which cross compile tools should be used? Linaro? / Debian?

mhennerich:
If you use a Debian based distro for ARM you can already get Debian packages:

https://packages.debian.org/sid/libiio0

markusr:
@mhennerich 
Yes, I know. The idea was more of getting the latest release.

I made some test but I encountered that cross compilation does not work, see https://github.com/analogdevicesinc/libiio/issues/27 .

We will first need to solve this issue.

rgetz:
There are armhf now, and upstream debian builds alot more...

@markusr : unless you have more that we can add to the CI, I will close this, and you can open a new one.

PR #216, opened 2018-11-02
travis-ci: don't hard code distributions anymore

rgetz:
try to make the system find out, as this is more robust

### DIFF
--- a/CI/travis/before_deploy
+++ b/CI/travis/before_deploy
@@ -18,8 +18,12 @@ else
 fi
 pwd
 
+if [ -z "${LDIST}" -a -f "build/.LDIST" ] ; then
+	LDIST="-$(cat build/.LDIST)"
+fi
 if [ -z "${LDIST}" ] ; then
-	LDIST=-$(lsb_release -c | awk '{print $NF}')
+	. CI/travis/get_ldist
+	LDIST="-$(get_ldist)"
 fi
 
 check_file()
@@ -52,7 +56,7 @@ if [ -n "${temp}" ] ; then
 	deploy=$(expr ${deploy} + 1)
 	if [ -z "${TARGET_DEB}" ] ; then
 		export TARGET_DEB=$(echo ${temp} | \
-			sed -e 's:^./.*/::' -e 's:.deb$::')${LDIST}.deb
+			sed -e 's:^./.*/::' -e 's:-Linux::' -e 's:.deb$::')${LDIST}.deb
 	fi
 	echo "deploying ${temp} to nightly $TARGET_DEB"
 	if [ -z "${RELEASE_PKG_FILE_DEB}" ] ; then
@@ -72,7 +76,7 @@ if [ -n "${temp}" ] ; then
 	deploy=$(expr ${deploy} + 1)
 	if [ -z "${TARGET_RPM}" ] ; then
 		export TARGET_RPM=$(echo ${temp} | \
-			sed -e 's:^./.*/::' -e 's:.rpm$::')${LDIST}.rpm
+			sed -e 's:^./.*/::' -e 's:-Linux::' -e 's:.rpm$::')${LDIST}.rpm
 	fi
 	echo "deploying ${temp} to nightly $TARGET_RPM"
 	if [ -z "${RELEASE_PKG_FILE_RPM}" ] ; then
@@ -137,7 +141,7 @@ if [  -n "${temp}"  ] ; then
 		)
 
 		export TARGET_TGZ=$(echo ${temp} | \
-			sed -e 's:^./.*/::' -e 's:.tar.gz$::')${LDIST}.tar.gz;
+			sed -e 's:^./.*/::' -e 's:-Linux::' -e 's:-Darwin::' -e 's:.tar.gz$::')${LDIST}.tar.gz;
 	fi
 	echo "deploying ${temp} to $TARGET_TGZ"
 	if [ -z "${RELEASE_PKG_FILE_TGZ}" ] ; then
--- a/CI/travis/get_ldist
+++ b/CI/travis/get_ldist
@@ -1,0 +1,35 @@
+get_ldist() {
+	case "$(uname)" in
+	Linux*)
+		if [ ! -f /etc/os-release ] ; then
+			if [ -f /etc/centos-release ] ; then
+				echo "centos-$(sed -e 's/CentOS release //' -e 's/(.*)$//' \
+					-e 's/ //g' /etc/centos-release)-$(uname -m)"
+				return 0
+			fi
+			ls /etc/*elease
+			[ -z "${OSTYPE}" ] || {
+				echo "${OSTYPE}-unknown"
+				return 0
+			}
+			echo "linux-unknown"
+			return 0
+		fi
+		. /etc/os-release
+		command dpkg --version >/dev/null 2>&1
+		if [ "$?" -ne 0 ] ; then
+			echo $ID-$VERSION_ID-$(uname -m)
+		else
+			echo $ID-$VERSION_ID-$(dpkg --print-architecture)
+		fi
+		;;
+	Darwin*)
+		echo "darwin-$(sw_vers -productVersion)"
+		;;
+	*)
+		echo "$(uname)-unknown"
+		;;
+	esac
+	return 0
+}
+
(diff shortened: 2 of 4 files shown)


rgetz:
This is for test purposes only - and the easiest thing to do to kick off a travis-ci is to make a pull request. Once this is ready to be reviewed - I will let folks know.

-Robin

rgetz:
Ok - this should be fine for review - it helps with #148 - which adds -amd64 to the release package names, or x86_64 or -armhf (if we get to that).

-robin

CI/travis/before_deploy
```diff
@@ -18,8 +18,41 @@ else
...
 pwd
 
+if [ -z "${LDIST}" -a -f "build/.LDIST" ] ; then
+	LDIST=$(cat build/.LDIST)
+fi
 if [ -z "${LDIST}" ] ; then
-	LDIST=-$(lsb_release -c | awk '{print $NF}')
+	# provide something like "-Debian-buster-amd64" or
+	# "-Ubuntu-xenial-amd64" or "-CentOS-7.5-x86_64"
+	case "$(uname)" in
```
commodo: while i like the idea, i think something a bit more elegant would work for Linux ;
maybe, break it down into a separate function (or functions) ; something like:
```
get_ldist() {
   if [ ! -f /etc/os-release ] ; then
        return 1
   fi
   source /etc/os-release
   [ -z "$UBUNTU_CODENAME" ] || {
         echo "$UBUNTU_CODENAME"
         return
   }

    [ -z "REDHAT_SUPPORT_PRODUCT" ] || {
         echo "${REDHAT_SUPPORT_PRODUCT}${REDHAT_SUPPORT_PRODUCT_VERSION}"
          return 0
    }
    return 1
}

LDIST=$(get_ldist)
```

some notes:
* this is untested but overall it should work
* for CentOS, there are other variables to choose from, in order to build the LDIST variable ( e.g. ID for `centos`, VERSION_ID or CENTOS_MANTISBT_PROJECT_VERSION for CentOS version ) ; I chose the REDHAT_ ones because they seemed handy
* the `ID_LIKE` variable can be used to organize distros together ; 
  * for example `ID_LIKE=debian` is common for Ubuntu & Raspbian  ; Debian doesn't have any ID_LIKE, it just has `ID=debian`
  * for CentOS, ID_LIKE seems to be `ID_LIKE="rhel fedora` ; I'd assume that ID_LIKE is a bit different for Fedora & Redhat, but it could still be useful

CI/travis/before_deploy
```diff
@@ -18,8 +18,41 @@ else
...
+	Linux*)
+		LDIST=$(awk -F= '/^NAME/{print $2}' /etc/os-release | \
+			sed -e 's| GNU/Linux||' -e 's/ Linux//' -e 's/"//g' \
+			-e 's/ /-/g' -e 's|/|_|g')
+		case "${LDIST}" in
+		Debian|Raspbian|Linaro|Ubuntu)
+			LDIST=-${LDIST}-$(lsb_release -c | awk '{print $NF}')
+			LDIST=${LDIST}-$(dpkg --print-architecture)
```
commodo: this means download links should change, right ?
System architecture wasn't part of LDIST before

rgetz:
Yes, urls will change. I will send a follow on pull request for README, updating the urls. And update other projects pulling from master-latest.

commodo:
The `.` should be pretty portable [vs `source`].

rgetz:
I will update things based on this, but based on https://www.freedesktop.org/software/systemd/man/os-release.html . What actually exists on all the different distribution that we want to build against - I think it a little more complicated...

If there aren't objections, I will use $VERSION_ID ("16.04" or "7.3") rather than codename.

so, what we would end up with is: -Debian-9-amd64

?

CI/travis/before_deploy
rgetz: This is a better idea - and I refactored things like that, but didn't use the vendor specific names, sticking with ones in the standard http://0pointer.de/blog/projects/os-release.html

rgetz:
Ok - deployed files names are now:

libiio-0.15.gb9e17fc-ubuntu-12.04-amd64.deb
libiio-0.15.gb9e17fc-ubuntu-14.04-amd64.deb
libiio-0.15.gb9e17fc-ubuntu-16.04-amd64.deb
libiio-0.15.gb9e17fc-centos-6.10-x86_64.rpm
libiio-0.15.gb9e17fc-centos-7-x86_64.rpm
libiio-0.15.gb9e17fc-darwin-10.10.5.pkg
libiio-0.15.gb9e17fc-darwin-10.11.6.pkg
libiio-0.15.gb9e17fc-darwin-10.13.3.pkg

does that work?

commodo:
Works for me [ the file names ]

CI/travis/get_ldist
```diff
@@ -0,0 +1,35 @@
+get_ldist() {
+	case "$(uname)" in
+	Linux*)
+		if [ ! -f /etc/os-release ] ; then
```
commodo: why this if block ?
is there a case where this happens ?
I admit I don't have a CentOS 6 VM to check ; I only have a CentOS 7 around

CI/travis/get_ldist
```diff
@@ -0,0 +1,35 @@
...
+			[ -z "${OSTYPE}" ] || {
+				echo "${OSTYPE}-unknown"
+				return 0
+			}
+			echo "linux-unknown"
+			return 0
+		fi
+		. /etc/os-release
+		command dpkg --version >/dev/null 2>&1
+		if [ "$?" -ne 0 ] ; then
```
commodo: nitpick: this could be compacted a bit like
```
if command dpkg --version &> dev/null ; then
     echo $ID-$VERSION_ID-$(dpkg --print-architecture)
else
     echo $ID-$VERSION_ID-$(uname -m)
fi
```

CI/travis/inside_centos_docker.sh
```diff
@@ -44,4 +44,7 @@ make package
 
 if [ "$TRAVIS_CI" == "travis-ci" ] ; then
 	yum -y install /libiio/build/libiio-*.rpm
+	# need to find this out inside the container
+	. /libiio/CI/travis/get_ldist
+	echo "$(get_ldist)" > /libiio/build/.LDIST
```
commodo: note/nitpick: I think `get_ldist > /libiio/build/.LDIST`  would also work
also fine to leave this as is

commodo:
one general note is that maybe `get_ldist` should add a dash prefix, since LDIST was initially conceived to work like that
or, we could check all occurrences of ${LDIST} expansions and add it there ;

CI/travis/get_ldist
rgetz: Darwin doesn't support /etc/os-release - that's a Linux only thing.

rgetz: centos 6 doesn't include /etc/os-release, they only added that in centos 7

CI/travis/get_ldist
rgetz: that will error on older versions of dash, and cause the script to error out and fail

rgetz:
there is a dash - it's included in LDIST="-$(cat build/.LDIST)" and LDIST="-$(get_ldist)" in CI/travis/before_deploy

rgetz:
@commodo :  Thanks for feedback.

commodo:
rest looks good ;

 any idea what this build failure means ?
```
+ssh -oHostKeyAlgorithms=+ssh-dss [secure]@[secure] rm [secure]/rgetz-test-donotmerge_libiio-0.15.g9f7c011-Linuxubuntu-14.04-amd64.deb
+ssh -oHostKeyAlgorithms=+ssh-dss [secure]@[secure] rm [secure]/rgetz-test-donotmerge_libiio-0.15.geaed20c-Darwin-osx-10.11.6.tar.gz
rm: [secure]/rgetz-test-donotmerge_libiio-0.15.geaed20c-Darwin-osx-10.11.6.tar.gz: No such file or directory
Script failed with status 1
```

commodo:
The branch name says `donotmerge`.
Will wait for the ok to merge this from @rgetz

rgetz:
that failure - is a build issue...

in order to keep the number of artifacts/files to a reasonable limit, we only keep the last 200 files. Since multiple build threads kick off, multiple vm can do an ls at the same time, and then only one is going to actually do the rm properly.

It doesn't happen often, so i haven't thought about a solution yet. Most likely - pick one VM, and only have that do the pruning. I will fix that at the same time as #199

rgetz:
this is OK to merge from me.

commodo:

[

PR #217, opened 2018-11-05
sorting: ensure sorting happens after global attributes are added

rgetz:
If global attributes are added after sorting, they are not included
in the sort, which defeats the purpose of things. This puts things
in the proper order (add all the attributes, then sort them).


commodo:
[LGTM](https://i.pinimg.com/564x/a1/0e/5b/a10e5bacb795e15dc756736615d8bc43.jpg)

PR #220, opened 2018-11-07
travis-ci: add build for Ubuntu 18.04 and drop Ubuntu 12.04

AlexandraTrifan:
- Add a docker build for Ubuntu 18.04.1 LTS (Bionic Beaver) : EOL April 2023.
- Ubuntu 12.04.5 LTS reached its regular End of Life on April 28, 2017 so we should remove the corresponding Travis CI build. 
- Builds with Xcode 6.4 in Travis CI are deprecated and will be removed in January 2019, so we can remove the Xcode 6.4 builds from Travis CI.

Signed-off-by: Alexandra.Trifan <Alexandra.Trifan@analog.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,10 @@ matrix:
       env:
         - OS_TYPE=centos
         - OS_VERSION=7
+    - os: linux
+      env:
+        - OS_TYPE=ubuntu_docker
+        - OS_VERSION=bionic
     - compiler: "gcc"
       os: osx
       osx_image: xcode6.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,9 +38,6 @@ matrix:
       env: LDIST=DO_NOT_DEPLOY
     - compiler: "gcc"
       os: linux
-      dist: precise
-    - compiler: "gcc"
-      os: linux
       dist: trusty
     - compiler: "gcc"
       os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,9 +58,6 @@ matrix:
         - OS_VERSION=bionic
     - compiler: "gcc"
       os: osx
-      osx_image: xcode6.4
-    - compiler: "gcc"
-      os: osx
       osx_image: xcode7.3
     - compiler: "gcc"
       os: osx
--- a/CI/travis/before_install_linux
+++ b/CI/travis/before_install_linux
@@ -11,9 +11,16 @@ handle_centos() {
 	sudo docker pull centos:centos${OS_VERSION}
 }
 
+handle_ubuntu_docker() {
+	sudo apt-get -qq update
+	sudo service docker restart
+	sleep 5
+	sudo docker pull ubuntu:${OS_VERSION}
+}
+
 handle_default() {
 	sudo apt-get -qq update
-	sudo apt-get install -y cmake doxygen libaio-dev libavahi-client-dev libavahi-common-dev libusb-1.0-0-dev libxml2-dev rpm tar bzip2 gzip
+	sudo apt-get install -y cmake doxygen libaio-dev libavahi-client-dev libavahi-common-dev libusb-1.0-0-dev libxml2-dev rpm tar bzip2 gzip flex bison git
 	if [ `sudo apt-cache search libserialport-dev | wc -l` -gt 0 ] ; then
 		sudo apt-get install -y libserialport-dev
 	fi
--- a/CI/travis/inside_bionic_docker.sh
+++ b/CI/travis/inside_bionic_docker.sh
@@ -1,0 +1,27 @@
+#!/bin/sh -xe
+
+TRAVIS_CI="$1"
+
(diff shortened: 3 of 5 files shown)


.travis.yml
```diff
@@ -36,9 +36,6 @@ matrix:
       os: linux
       dist: trusty
       env: LDIST=DO_NOT_DEPLOY
-    - compiler: "gcc"
-      os: linux
-      dist: precise
```
commodo: I think you could also remove `osx_image: xcode6.4`
you could do it in this commit, or make a new one ; doesn't matter much

CI/travis/before_install_linux
```diff
@@ -11,6 +11,13 @@ handle_centos() {
 	sudo docker pull centos:centos${OS_VERSION}
 }
 
+handle_bionic() {
```
commodo: you could call this `handle_ubuntu_docker` and it should only run for builds that define `OS_TYPE=ubuntu_docker`
and then you can do `sudo docker pull ubuntu:$OS_VERSION` and you wouldn't need the changes in the `handle_${OS_TYPE}` logic

cmake/LinuxPackaging.cmake
```diff
@@ -48,6 +48,9 @@ if(DEB_DETECT_DEPENDENCIES AND DPKG_CMD AND DPKGQ_CMD)
 	if(WITH_XML_BACKEND)
 		set(PACKAGES "${PACKAGES} libxml2")
 	endif()
+	if(WITH_SERIAL_BACKEND)
```
commodo: these packaging changes look like they could go into a separate commit that handles only Linux packaging

CI/travis/inside_bionic_docker.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh -xe
+
+TRAVIS_CI="$1"
+
+apt-get -qq update
+apt-get install -y cmake doxygen libaio-dev libavahi-client-dev libavahi-common-dev libusb-1.0-0-dev libxml2-dev git
+apt-get install -y tar bzip2 gzip flex bison
```
commodo: you could include the `before_install_linux` script here ;
for Ubuntu this is do-able, since it's common code with the main code that runs in Travis-CI

CI/travis/make_linux
```diff
@@ -16,8 +16,19 @@ handle_centos() {
 		/bin/bash -xe /libiio/CI/travis/inside_centos_docker.sh ${OS_VERSION} travis-ci
 }
 
+handle_bionic() {
```
commodo: same here with `handle_ubuntu_docker` and using `$OS_VERSION` instead of `bionic`

this helps us extend for future Ubuntu versions (if we want to)

commodo:
Codacy has one small complaint.
I admit that this code was adapted from an existing one; we could fix the old one as well (not necessarily in this PR).

.travis.yml
AlexandraTrifan: Added a new commit to remove the xcode 6.4 image

commodo:
The Codacy complaint went away after `before_install_linux` was included;

.travis.yml
```diff
@@ -36,9 +36,6 @@ matrix:
       os: linux
       dist: trusty
       env: LDIST=DO_NOT_DEPLOY
-    - compiler: "gcc"
```
commodo: for this change it would be a good idea to add a small comment about why it's being removed (i.e. 12.04 has been deprecated, and will be dropped by Travis-CI in the future, etc, etc, etc)

.travis.yml
```diff
@@ -56,9 +56,6 @@ matrix:
       env:
         - OS_TYPE=ubuntu_docker
         - OS_VERSION=bionic
-    - compiler: "gcc"
```
commodo: similarly:
for this change it would be a good idea to add a small comment about why it's being removed (i.e. Xcode 6.4 has been deprecated, and will be dropped by Travis-CI in the future, etc, etc, etc)

commodo:
[LGTM](http://i.imgur.com/YuL9Yns.jpg)

.travis.yml
AlexandraTrifan: Added some details in the commit message and the PR description.

.travis.yml
AlexandraTrifan: the same: added some details in the commit message and the PR description.

tfcollins:
Those three look fine to me.  Ideally, we would have 8.2 but that isn't available. Drop the oldest if we want to reduce the number of builds.

AlexandraTrifan:
The Xcode 8.1 build is automatically routed to the Xcode 8.3 image.
I will make the required changes to use those three versions and I will create a new pull request for that, if that's ok.

commodo:
Let's do in a subsequent PR.
It allows us to move forward and iterate over this and other takes.

Issue #218, opened 2018-11-05
adc scale attributes include extra newline


rgetz:
I don't know if this is a bug in libiio, or in the kernel.

in some kernel drivers, the attributes (accidently?) puts a newline at the end of the field, which gets added, and sent to the entire system...

```
dev 'ad7291', channel 'voltage0' (input), attr 'raw', value '3330'
dev 'ad7291', channel 'voltage0' (input), attr 'scale', value '0.610351562
'
dev 'ad7291', channel 'voltage1' (input), attr 'raw', value '3364'
dev 'ad7291', channel 'voltage1' (input), attr 'scale', value '0.610351562
'
dev 'ad7291', channel 'voltage2' (input), attr 'raw', value '3367'
dev 'ad7291', channel 'voltage2' (input), attr 'scale', value '0.610351562
'
```
rather than

```
dev 'ad7291', channel 'voltage0' (input), attr 'raw', value '3378'
dev 'ad7291', channel 'voltage0' (input), attr 'scale', value '0.610351562'
dev 'ad7291', channel 'voltage1' (input), attr 'raw', value '3364'
dev 'ad7291', channel 'voltage1' (input), attr 'scale', value '0.610351562'
dev 'ad7291', channel 'voltage2' (input), attr 'raw', value '3368'
dev 'ad7291', channel 'voltage2' (input), attr 'scale', value '0.610351562'
dev 'ad7291', channel 'voltage3' (input), attr 'raw', value '1307'
dev 'ad7291', channel 'voltage3' (input), attr 'scale', value '0.610351562'
dev 'ad7291', channel 'voltage4' (input), attr 'raw', value '2510'
dev 'ad7291', channel 'voltage4' (input), attr 'scale', value '0.610351562'
dev 'ad7291', channel 'voltage5' (input), attr 'raw', value '1303'
dev 'ad7291', channel 'voltage5' (input), attr 'scale', value '0.610351562'
dev 'ad7291', channel 'voltage6' (input), attr 'raw', value '1'
dev 'ad7291', channel 'voltage6' (input), attr 'scale', value '0.610351562'
dev 'ad7291', channel 'voltage7' (input), attr 'raw', value '1'
dev 'ad7291', channel 'voltage7' (input), attr 'scale', value '0.610351562'
```

if we want to eat newlines at the end of the line, this works, but I wasn't sure how deep to put things (or change it in the kernel driver).

```
root@analog:/usr/local/src/libiio# git diff
[...]

PR #223, opened 2018-11-12
In some older kernels, the kernel accidently returns two newlines

rgetz:
This ensures that libiio eats the additional newlines, which can
cause some userspace issues.

analog@analog:~$ hexdump -c /sys/bus/iio/devices/iio:device0/in_voltage_scale
0000000   0   .   6   1   0   3   5   1   5   6   2  \n  \n

now prints out as:

dev 'ad7291', channel 'voltage0' (input), attr 'scale', value '0.610351562'

before it had a newline in the single quotes.


rgetz:
This will fix #218

pcercuei:
For attributes that contain raw data (e.g. fir_filter), the last bytes of the actual data can be the code for the ASCII \n, this would strip it away...

mhennerich:
> @mhennerich would need to chime in - I don't think that fir_filter returns anything - it's all ASCII into the filter. I thought all attributes were always ASCII.

fir_filter, profiles and gain_tables configs are all ASCII. The only inkernel driver (staging) is the ad5933 which uses bin format attributes. However it's scheduled for lifting out of staging, this usage will be subject for revision.

pcercuei:
The double line-ending character bug was in mainline too? Or just ADI's kernel?

Binary format attributes are deprecated now?

mhennerich:
> The double line-ending character bug was in mainline too? Or just ADI's kernel?
> 
> Binary format attributes are deprecated now?

I think it was only in the ADI kernel, due to a merge conflict. At some point __iio_format_value() was adding the '/n', then it was moved to iio_format_value().
https://github.com/analogdevicesinc/linux/commit/b58e9efe2da6c39906c2ae6f009b810eff05d66e

I think binary attributes are deprecated. And in fact the ad5593 driver implements it's own real kfifo based buffer now.

pcercuei:
If the bug is not in mainline, and it was fixed in ADI's kernel, I think we don't need the fix in libiio. But it's up to you.

PR #225, opened 2018-11-13
fix snprintf warnings from gcc 8

rgetz:
In recent versions of gcc, there are a few places where gcc warns:
iio_info.c: In function ‘main’:
iio_info.c:333:41: warning: ‘%u’ directive output may be truncated writing between 1 and 10 bytes into a region of size 7 [-Wformat-truncation=]
      snprintf(repeat, sizeof(repeat), "X%u",
                                         ^~
This fixes the warn by increasing the size of the buffer to 12 from 8.
No functional differences.

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -190,7 +190,7 @@ static char *get_attr_xml(struct iio_channel_attr *attr, size_t *length)
 
 static char * get_scan_element(const struct iio_channel *chn, size_t *length)
 {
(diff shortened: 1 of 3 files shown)


commodo:
Will merge later today if no objections

[LGTM](https://i0.wp.com/www.geeksaresexy.net/wp-content/uploads/2013/09/batman1.jpg?resize=600%2C628)

PR #228, opened 2018-11-15
doxygen: If dot exists, use it.

rgetz:
Add some cmake logic to use graphviz/dot to include call graphs in
the generated docs. This helps users understand the details of the
library.


pcercuei:
Some changes are unrelated to graphviz/dot, could you split them into another commit?

rgetz:
Sure - updated

rgetz:
Fixed the Windows issue - have cmake check the case insensitivity of the file system, and report that to Doxygen (as a separate commit).

rgetz:
in travis-ci, graphviz wasn't installed - only install it on the ubuntu images (to keep the time down).

CI/travis/before_install_linux
```diff
@@ -20,7 +20,7 @@ handle_ubuntu_docker() {
 
 handle_default() {
 	sudo apt-get -qq update
-	sudo apt-get install -y cmake doxygen libaio-dev libavahi-client-dev libavahi-common-dev libusb-1.0-0-dev libxml2-dev rpm tar bzip2 gzip flex bison git
+	sudo apt-get install -y cmake doxygen graphviz libaio-dev libavahi-client-dev libavahi-common-dev libusb-1.0-0-dev libxml2-dev rpm tar bzip2 gzip flex bison git
```
commodo: do we also need to do the same for CentOS here https://github.com/analogdevicesinc/libiio/blob/master/CI/travis/inside_centos_docker.sh#L8 ?

rgetz: No - the doxygen html isn't actually deployed anywhere (yet), and it just takes extra time.

commodo: ack

PR #231, opened 2018-11-20
build: trigger libad9361-iio build

commodo:
Let's start a "pipeline", where the libiio build triggers a
`libad9361-iio`, which will trigger other builds as well.

We're getting to a point where the lack of automation is starting to hinder
us a bit and changing things in `libiio` are not visible along the way,
until it gets a bit late.

Note that this does not trigger appveyor yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,9 @@ matrix:
       osx_image: xcode10.1
       env:
         - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
+    - os: linux
+      env:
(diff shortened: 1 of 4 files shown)


commodo:
Yep.

This is the start of the pipeline.
Next will be libad9361-iio, then iio-osciloscope, gnuradio, gr-iio and scopy

commodo:
Will merge this once libad9361-iio is ready.

commodo:
Will re-open this from another branch

PR #232, opened 2018-11-20
build,travis: parametrize library name

commodo:
The intent is to re-use some of these scripts for libad9361-iio, which has
quite a bit of the same build rules.


commodo:

PR #233, opened 2018-11-21
CI/travis/get_ldist: move dpkg check in if check

commodo:
This is simpler, and works well when enabling the `-e` (errexit) shell
option. Otherwise the shell exits when `command dpkg` returns a non-zero
exit code.


commodo:

PR #234, opened 2018-11-21
 CI/travis/deploy: generalize garbage cleanup

commodo:
There are 4222 build artifacts stored on the server at the time of this
writing.
Here's a count:
   https://api.travis-ci.org/v3/job/457853889/log.txt

We should limit this a bit more. So now the idea it to limit this per
file-type (deb, rpm, tgz). About the oldest 30 per file-type sounds about
reasonable.


rgetz:
I don't like this - I use these images all the time.

I will need to review in more detail, but don't have time today.

commodo:
Ok.
I thought maybe there would be some usage of these files, but in any case, I will argue that some procedure for keeping the list down is needed.

So, now with this changeset, things get deployed only for branches `master` & branches `20xx_Ry`. I can understand that there may be a need to deploy PR branches (for testing), and I was thinking about this since yesterday, and maybe one solution [which takes some work, but not a whole lot] would be:
1. put back the branch filter in the GLOB var
2. let's limit the number of branches with this filter to 5-10 [per PR/branch/etc] ; I don't think we will ever need to keep the last 100 files for each branch
3. when deploying on the `master` branch, we can add a rule/script that retrieves all the branches for this repo, and deletes all the files that don't have that branch prefix ; we would need to exclude from this list:
   * filenames prefixed with pipeline branches (master & branches `20xx_Ry`)
   * filenames prefixed with `libiio-` ; those seem historical ; I am bit surprised those are still used though, since I guess people have mostly forgot about them, and use the files from the Github releases page: https://github.com/analogdevicesinc/libiio/releases

If we do get this done, maybe we would need a new repo for this script, or we could just put it here and re-use it in other repos by downloading it.

Unless I'm missing something here, I guess that would cover most aspects I can think of.
But, input/suggestions are welcome.

rgetz:
There something that keeps the list down - it just doesn't keep it down as much as you want. :) (Which I agree with you - it needs some work).

See https://github.com/analogdevicesinc/libiio/blob/master/CI/travis/deploy#L54

commodo:
I know that part of the script ; I updated a few months ago to also cleanup non-Precise deb packages.
That still leaves un-handled rpm, tar.gz and other stuff.

Will update in a new series.

PR #239, opened 2018-11-23
CI/travis: reorganize scripts (part 1)

commodo:
The aim is to create a generic `CI/travis/lib.sh` that gets re-used in other builds as well.
Otherwise a lot of the deploy & build logic gets duplicated in other repos along the way.


commodo:
Will merge this later today if no objections.

rgetz:
This is a bad idea. It was implemented in the yaml on purpose for Travis-ci security issues. 

The yaml doesn't expose the secret keys to the log like a shell does. Please revert this. I now need to revoke all the existing keys, and update them to new ones. 

Making a pull request and committing it on the same day, on days that people are out -  is sub optimal. :(

commodo:
My bad.
To me it looks like Travis-CI does a reasonably good job not displaying env-vars in the log files, but there may be stuff I am not aware.

Regarding the haste: I'll assume the blame. Things keep piling up and frustration-levels go up [when they're not getting done].

PR #240, opened 2018-11-23
CI/travis/deploy: lib-ify the upload function

commodo:
The `send()` function is 90-99% re-usable for uploading artifacts, and cleaning them.
The cleaning up part, will need some work, but the first step is to make it a library-type function.


rgetz:
The remote commands are super fragile - akamai doesn't use a standard sftp or ssh server, so they only do a fraction of the actual commands. It took a long time to make this reliable.

commodo:
ack;
we'll probably need to test things, but the idea of the `CI/travis/lib.sh: rework sftp script generation` is to just group the commands ;

at least in this form, they're easier to extend/work-with if we need to do it [later on]

PR #242, opened 2018-11-26
.travis.yml,CI/travis/after_deploy: trigger other builds after libiio

commodo:
Let's start a "pipeline", where the libiio build triggers a
`libad9361-iio`, which will trigger other builds as well.

We're getting to a point where the lack of automation is starting to hinder
us a bit and changing things in `libiio` are not visible along the way,
until it gets a bit late.

There's a Python script that each job would run, to check if it's the last
job (in the parallel CI run), and the last job, will trigger the next
build.
There's no better way at the moment to do this, without re-factoring the
entire `.travis.yml` file to work with build-stages (a new feature that's a
bit of work to get working properly with our current setup).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -139,3 +139,6 @@ deploy:
     on:
       condition: "$TRAVIS_OS_NAME = osx"
       all_branches: true
+
+after_deploy:
+  - ${TRAVIS_BUILD_DIR}/CI/travis/after_deploy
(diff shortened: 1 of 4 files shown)


commodo:
Did a few tests with a hard-coded `master` branch in the `b2` branch.
Raw build log for the build that triggered the next build: https://api.travis-ci.org/v3/job/460152961/log.txt

The entire build status: https://travis-ci.org/analogdevicesinc/libiio/builds/460152950
The last build triggers the next build.

commodo:
If no objections, will merge this in 2 days.

PR #246, opened 2018-12-11
CI/travis: lib-ify docker logic

commodo:
We want to re-use the docker logic in other repos (e.g. libad9361-iio).
We need to lib-ify that logic, since we won't download `make_linux` and
`before_install_linux` and similar files.


commodo:
Will merge this next week, if no objections.
Should be enough time for others to take a look.

PR #247, opened 2018-12-12
CI/travis/lib.sh: fix dangling quote in symlink command

commodo:
This seems to have slipped through after doing a few rebases.
I've fixed it a few times on local branches.

The SFTP script did not error out, so that's why it wasn't caught there.


commodo:

PR #248, opened 2018-12-17
update lib.sh deployment logic

commodo:
* check if script is present in build folder so it doesn't re-download stuff everything it gets included
* ensure wget is present on the machine (particularly docker)


commodo:

PR #249, opened 2018-12-19
CI/travis/lib.sh: shell parsing errors

commodo:
These errors only show up when the script gets deployed via wget.
They don't interfere with libiio's build.


commodo:

PR #250, opened 2018-12-21
build,travis-ci: fix CentOS 7 artifacts & unify docker builds

commodo:
Turns out that using `CPACK_RPM_EXCLUDE_FROM_AUTO_FILELIST_ADDITION` does work for handling the issues with installing CentOS 7 RPMs.
I don't know what goof I did initially that made me think it doesn't work.

That means that the `inside_*_docker.sh` scripts can be unified, since the CentOS 7 hack is no longer needed.


commodo:
Will merge this on the 8th of January 2019 if no objections

PR #252, opened 2019-01-08
ci/travis: updates for docker builds & ci/travis lib path

commodo:
added INSIDE_DOCKER env var: 
* This can be used by scripts running inside docker environments to do
special behavior when running inside this type of env.

make `get_script_path()` look into `ci/travis/`: some repos are starting to use `ci/travis/` vs `CI/travis` ; so update the function to look into those folders as well


commodo:

PR #253, opened 2019-01-09
travis-ci: enable raspberry on qemu inside travis-ci 

rgetz:
This adds a two more distributions to the build matrix : raspbian jessie and  raspbian stretch.

This should also provide a basic framework for those non-native travis-ci architectures in the future that we may want to support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,18 @@ services:
 matrix:
   include:
     - compiler: "gcc"
+      dist: xenial
+      env:
+        - ARCH=arm
+        - OS_VERSION=jessie
+        - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
+    - compiler: "gcc"
+      dist: xenial
+      env:
+        - ARCH=arm
+        - OS_VERSION=stretch
+        - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
+    - compiler: "gcc"
       os: linux
       env:
         - PLATFORM=linux BITS=64 HOST=x86_64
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -15,6 +15,18 @@ services:
 
 matrix:
   include:
+    - compiler: "gcc"
+      dist: xenial
+      env:
+        - ARCH=arm
```
commodo: if you change this to `OS_TYPE=qemu_arm` and define a `handle_qemu_arm()` function in `./CI/travis/before_install_linux` and another one in `./CI/travis/make_linux`

moving the code from ` ./CI/travis/setup_qemu_for_arm.sh` to `./CI/travis/before_install_linux is a bit much; so `handle_qemu_arm()`  could just call or include `./CI/travis/setup_qemu_for_arm.sh`

CI/travis/setup_qemu_for_arm.sh
```diff
@@ -0,0 +1,87 @@
...
+# Host dependencies
+sudo apt-get update
+sudo apt-get install -qq -y qemu-user-static binfmt-support sbuild wget debian-archive-keyring ubuntu-keyring gnupg libudev1 libudev-dev liblzma-dev libglib2.0-dev libfdt-dev libpixman-1-dev zlib1g-dev
+# need a newer version of qemu :(
+QEMU_DIR=/tmp/qemu
+sudo mkdir -p ${QEMU_DIR}
+sudo chown ${USER} /tmp/qemu
+cd ${QEMU_DIR}
+wget https://download.qemu.org/qemu-3.1.0.tar.xz
+tar xf qemu-3.1.0.tar.xz
```
commodo: curios: would the default QEMU work ?

commodo: actually, there may be some good news here [no guarantees that it actually works though, but it would be worth taking a closer look] ;
you could use Travis-CI's integration with Debian Sid packages to get a new QEMU [3.1 or something]

QEMU in Sid seems to be 3.1 [now] : https://packages.debian.org/sid/qemu

What you typically would do, is:
```
addons:
  apt:
    sources:
    - debian-sid
    packages:
    - qemu
```

I "think" Travis-CI keeps a list of whitelisted packages; they don't seem to just allow anything from Sid. I don't remember exactly.

I did do this [Debian-Sid usage] a while back on some other project, and it's a mix of experiences ; some worked great, but if you mix-in other stuff [or APT sources] then using `debian-sid` can confuse other parts of the build.

Still [if possible] I would try it, mostly to reduce the build logic here, but especially to try to reduce the build-time.

rgetz: > curios: would the default QEMU work ?

No - it's too old, and gives errors during make package. it doesn't support some ioctls which cmake uses.

rgetz: > Still [if possible] I would try it, mostly to reduce the build logic here, but especially to try to reduce the build-time.

building qemu doesn't take the time (thats 2-3 min our of 20) - it's downloading the entire dist for ARM that takes the time...

.travis.yml
rgetz: Thanks for review - I can try this later tonight.

CI/travis/setup_qemu_for_arm.sh
rgetz: > QEMU in Sid seems to be 3.1 [now] : https://packages.debian.org/sid/qemu

We actually need qemu-user-static to run in the chroot; so it would be https://packages.debian.org/sid/qemu-user-static (which is still there).

rgetz:
I thought I answered all the questions — can this get merged? We did a release without ...

mhennerich:
> I thought I answered all the questions — can this get merged? We did a release without ...

Yes this can get merged - in fact I don't know why this hasn't been already. I approved this PR 4 months ago. Please follow up if things don't move...

mhennerich:
Since this PR was only CI related and didn't change any library source files. I moved to the v0.18 release tag to include this. A new build is currently triggered.

PR #254, opened 2019-01-09
Revert "sort: when sorting iio_channels, if the index is the same, us…

mhennerich:
…e ID"

This commit broke interop with previous versions of libiio, in certain scenarios.
The interoperability issue is only a problem in case all channels use the same index,
which is typically only the case with the bit mask channels used for logic analyzers.
This means v0.16 can’t interop with previous versions,
since channels are now sorted differently on the local and remote backend.

This reverts commit 67a994c973e68ddcf48c64ebaa0001c69966f6e8.


commodo:
[LGTM](https://i.pinimg.com/564x/75/6a/b0/756ab09b1e4bce4208072b9cb84252b5.jpg)

pcercuei:
You could very well do the sorting within iio_info.

PR #256, opened 2019-01-22
CI/travis/lib.sh: rework artifact removal via SFTP

commodo:
We can avoid downloading artifacts to see if they exist or now. This saves
a bit of time and some bandwidth. If we just do a `rm` command via SFTP
and allow it to fail, this should be just as good.

The `sftp` command is also wrapped into it's own function, to unclutter
some of the SFTP logic.


commodo:
This looks good now; artifacts seem to upload.
If no objections, will merge this on Friday or next week

PR #258, opened 2019-02-07
CI/travis/lib.sh: make the build dir configurable

commodo:
This allows other scripts to specify a different build directory where
the scripts (that get downloaded) will be placed.


commodo:
Changelog v1 -> v2:
* rebased on newer master after PR #256

PR #261, opened 2019-02-13
CI/travis/jobs_running_cnt.py: count `created` states as well

commodo:
Previously we were just counting `started` states, but it seems that
Travis-CI has a `created` state for jobs, which is before the `started`
state.
This means some builds were triggering before the entire job group would
finish.


commodo:

PR #269, opened 2019-04-25
sorting: Use old sorting algorithm for channels

adisuciu:
This algorithm should be backwards compatible

Signed-off-by: Adrian Suciu <adrian.suciu@analog.com>

### DIFF
--- a/context.c
+++ b/context.c
@@ -202,11 +202,30 @@ struct iio_device * iio_context_find_device(const struct iio_context *ctx,
 
 static void reorder_channels(struct iio_device *dev)
 {
+	bool found;
 	unsigned int i;
 
 	/* Reorder channels by index */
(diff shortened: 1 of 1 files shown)


pcercuei:
I think it is OK as it is now, the apps should not rely on the sorting of the channels or attributes, but rather use iio_device_find_channel.

mhennerich:
No - it's still broken -
The issue is that remote via IIOD and local share a channel enable mask. If channel sorting on both ends is done differently things fail, since the bits set doesn’t correspond to the same channels. 
If we want to maintain backwards compatibility I see no way around reverting the channel sorting completely.

mhennerich:
Hi @pcercuei,

Do you see a way of fixing this while maintaining backward compatibility?

pcercuei:
Ah, right. I don't think there's an easy way to fix this then. The revert would be the proper solution.

PR #270, opened 2019-05-04
local: Improve support for channels with extended names

egranata:
On Chrome OS devices, there is a driver named 'cros_ec_sensors_ring',
which defines a channel of type IIO_ACCEL with an extended name of "flag".

When trying to parse the 'en' attribute of this channel, libiio parses it
as a 'flag_en' attribute of the 'accel' channel instead, in turn causing
the context creation to fail due to an unrecognized attribute name.

This patch works around this issue by locating the last underscore as
the channel name marker, instead of the first. This, in turn, yields a
correct channel name of 'accel_flag' and a valid attribute name 'en'.

To the best of my understanding, this should not change the behavior of
libiio for channels that do not have an extended name.

Signed-off-by: Enrico Granata <granata.enrico@gmail.com>

### DIFF
--- a/local.c
+++ b/local.c
@@ -1135,7 +1135,7 @@ static char * get_channel_id(const char *attr)
 	size_t len;
 
 	attr = strchr(attr, '_') + 1;
-	ptr = strchr(attr, '_');
+	ptr = strrchr(attr, '_');
 	if (find_channel_modifier(ptr + 1, &len) != IIO_NO_MOD)
 		ptr += len + 1;
 


mhennerich:
Maybe it fixes your use case - but I can tell for sure it causes all kinds of undesired behavior everywhere else where extended names are used. The extended name thing as is was a pretty bad decision, because it's not parse-able. Maybe we should propose an additional attribute that just returns the extended name for a given channel. This additional attribute would be added by the IIO core. Alternatively we could introduce a naming convention that extended names must be enclosed by brackes.  
Example: in_voltageX_[extended_name]_raw.

egranata:
> Maybe it fixes your use case - but I can tell for sure it causes all kinds of undesired behavior everywhere else where extended names are used. The extended name thing as is was a pretty bad decision, because it's not parse-able.

Fair enough, and yeah duly noted. Thanks for pointing it out.

> Maybe we should propose an additional attribute that just returns the extended name for a given channel. This additional attribute would be added by the IIO core. Alternatively we could introduce a naming convention that extended names must be enclosed by brackets.

This seems like a good idea. It would be nice if we could resolve this parsing ambiguity at the core. Would the linux-iio mailing list be the proper venue to propose such a change?

> Example: in_voltageX_[extended_name]_raw.

pcercuei:
The attribute format cannot be changed, it's ABI now...
You face this problem because you only have one attribute for your "flag" channel. If you add a second (dummy) one libiio should parse them properly, as it will properly detect "flag" being the extended name.

matejk:
During my work with I encountered the same issue.

`extend_name` as it is used currently, does not really define the channel name and it is parsed with some heuristics in libiio that does not work for channels with a single attribute, which makes it inconsistent.

IMHO the cleanest way to resolve this would be an additional sysfs entry for channel extend name (alias channel name) that would be used in libiio. For example:

- `extend_voltage0` or
- `out_voltage0_extend`

```
# ls -1 /sys/bus/iio/devices/iio\:device21/
dev
name
of_node
extend_voltage0
out_voltage0_relay-spi-0_offset
out_voltage0_relay-spi-0_raw
power
subsystem
uevent
# cat /sys/bus/iio/devices/iio\:device21/extend_voltage0
relay-spi-0
```

`extend_*` would be an empty string if not defined but present anyway. For backward compatibility, libiio can still fall back to current behaviour if `extend_*` is not present.

dNechita:
Closing this as the proposed changes will break things.
This particular use case, where a channel with extended name and only one attribute gets parsed incorrectly, remains a known issue but solvable as per @pcercuei suggestion.
It is most likely that a cleaner solution will be provided with the next generation of this library.

Issue #25, opened 2016-03-04
Handling channels that have no modifier or are not indexed


gburca:
iio_channel_is_scan_element() returns false for channels without a modifier and that are not indexed, even though the kernel creates an entry in the scan_elements directory for them. Is that correct?

For example, a temperature sensor would have a single channel (so there's no need to index it), and it wouldn't need a modifier either. It would show up as "in_temp_raw" in the device directory and have "in_temp_{en|index|type}" files in the scan_elements directory.

lclausen-adi:
Hi,

It's not intentional. Please send a patch.

Thanks,
- Lars

abrauchli:
Potentially fixed by https://github.com/analogdevicesinc/libiio/pull/139

gwendalcr:
#139  has been reverted by fb5fa69.  #273 addresses the problem, but how do I ensure it does not add a regression?

gwendalcr:
#366 fixes the issue.

rgetz:
Ok - since it's fixed, I will close. Thanks

gwendalcr:
I was wrong, it is not fixed.
For simple attribute, like in_illuminance, found on driver like acpi_als, is_channel() returns false and iio_info does not report in_illuminance as a channel, iio_readdev can not read it.

With a trivial path (forcing is_channel("in_illuminance_..", true) to return true, similar to commit 4c6baf5e ("Local backend: Properly detect "timestamp" as a channel"), iio_genxml present acpi_als illuminance as channel:

from:
  <device id="iio:device0" name="acpi-als">
    <channel id="illuminance" type="input">
      <attribute name="raw" filename="in_illuminance_raw"/>
    </channel>
...

to:
  <device id="iio:device0" name="acpi-als">
    <channel id="illuminance" type="input">
      <scan-element index="0" format="le:S32/32&gt;&gt;0"/>
      <attribute name="raw" filename="in_illuminance_raw"/>
    </channel>
 
Since #273 is too generic, should we look at scan_elements/ directory to identify channels instead of the attributes at the root of iio:deviceX?

PR #273, opened 2019-05-10
Hanlde channels that have no modifier or are not indexed

gwendalcr:
Another attempts to fix https://github.com/analogdevicesinc/libiio/issues/25

The idea is to change is_channel() from looking at index and modifier to a different format:
- must start with in_ or out_
- next, it should be a channel type
- must have an extra '_'.

Test: Using iio_info: before:
          iio:device0: cros-ec-light (buffer capable)
                  2 channels found:
                          timestamp:  (input, index: 1, format: le:S64/64>>0)
                          illuminance:  (input)
                          3 channel-specific attributes found:
                                  attr  0: calibbias value: 0
                                  attr  1: calibscale value: 56.360700
                                  attr  2: raw value: 55

after:
         iio:device0: cros-ec-light (buffer capable)
                  2 channels found:
                          illuminance:  (input, index: 0, format: le:U16/16>>0)
                          3 channel-specific attributes found:
                                  attr  0: calibbias value: 0
                                  attr  1: calibscale value: 56.360700
                                  attr  2: raw value: 52
                          timestamp:  (input, index: 1, format: le:S64/64>>0)

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -113,7 +113,7 @@ enum iio_modifier find_channel_modifier(const char *s, size_t *len_p)
 			continue;
 		len = strlen(modifier_names[i]);
 		if (strncmp(s, modifier_names[i], len) == 0 &&
-				(s[len] == '\0' || s[len] == '_')) {
+				(s[len] == '\0' || s[len] == '_' )) {
 			if (len_p)
 				*len_p = len;
 			return i;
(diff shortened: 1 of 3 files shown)


channel.c
```diff
@@ -124,34 +124,42 @@ enum iio_modifier find_channel_modifier(const char *s, size_t *len_p)
...
+
 		/* Type must be followed by either a '_' or a digit */
-		if (chn->id[len] != '_' && (chn->id[len] < '0' || chn->id[len] > '9'))
+		if (name[len] != '_' && (name[len] < '0' || name[len] > '9'))
 			continue;
 
-		chn->type = (enum iio_chan_type) i;
+		break;
 	}
+	return (enum iio_chan_type) i;
```
commodo: maybe return `IIO_CHAN_TYPE_UNKNOWN` here;

channel.c
```diff
@@ -124,34 +124,42 @@ enum iio_modifier find_channel_modifier(const char *s, size_t *len_p)
...
-		if (strncmp(iio_chan_type_name_spec[i], chn->id, len) != 0)
+		if (strncmp(iio_chan_type_name_spec[i], name, len) != 0)
 			continue;
+
 		/* Type must be followed by either a '_' or a digit */
-		if (chn->id[len] != '_' && (chn->id[len] < '0' || chn->id[len] > '9'))
+		if (name[len] != '_' && (name[len] < '0' || name[len] > '9'))
 			continue;
 
-		chn->type = (enum iio_chan_type) i;
```
commodo: and maybe `return (enum iio_chan_type) i;`  here
if you break here, you will never return `IIO_CHAN_TYPE_UNKNOWN`

channel.c
```diff
@@ -152,17 +152,7 @@ void iio_channel_init_finalize(struct iio_channel *chn)
 		return;
 
 	mod++;
-
-	for (i = 0; i < ARRAY_SIZE(modifier_names); i++) {
-		if (!modifier_names[i])
-			continue;
-		len = strlen(modifier_names[i]);
-		if (strncmp(modifier_names[i], mod, len) != 0)
```
commodo: the only thing here [and where we'd need to take care] is the fact that `find_channel_modifier()` also checks if `(s[len] == '\0' || s[len] == '_')`
that [in itself] could be fine for this change; though i am not completely sure;

gwendalcr: You're right, We can find numbers as well behind the modifier.

gwendalcr: Actually, there is not number behind the modifier: If I understand correctly, the channel name is:
[in|out]_[type]{channel number}_[modifier]{_extra_modifier}

gwendalcr:
Opening another pull request with modified patches.

PR #278, opened 2019-05-26
avoid some null pointer derefs

ckuethe:
`SoapyUtil --probe` can trigger various null pointer derefences while trying to probe a PlutoSDR (https://github.com/pothosware/SoapyPlutoSDR/wiki)

Signed-off-by: Chris Kuethe <chris.kuethe@gmail.com>

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -351,6 +351,9 @@ const char * iio_channel_find_attr(const struct iio_channel *chn,
 ssize_t iio_channel_attr_read(const struct iio_channel *chn,
 		const char *attr, char *dst, size_t len)
 {
+	if (!chn)
+		return -ENOSYS;
+
 	if (chn->dev->ctx->ops->read_channel_attr)
 		return chn->dev->ctx->ops->read_channel_attr(chn,
 				attr, dst, len);
@@ -361,6 +364,9 @@ ssize_t iio_channel_attr_read(const struct iio_channel *chn,
 ssize_t iio_channel_attr_write_raw(const struct iio_channel *chn,
 		const char *attr, const void *src, size_t len)
 {
+	if (!chn)
+		return -ENOSYS;
+
 	if (chn->dev->ctx->ops->write_channel_attr)
 		return chn->dev->ctx->ops->write_channel_attr(chn,
 				attr, src, len);
--- a/device.c
+++ b/device.c
@@ -250,6 +250,9 @@ struct iio_channel * iio_device_find_channel(const struct iio_device *dev,
 		const char *name, bool output)
 {
 	unsigned int i;
+	if (!dev)
+		return NULL;
+
 	for (i = 0; i < dev->nb_channels; i++) {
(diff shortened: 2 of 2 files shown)


pcercuei:
I think that should be a fix for Soapy then; it should ensure that it passes valid pointers to libiio.

ckuethe:
I'll look into that too, but a null pointer check seems like cheap insurance.

commodo:
> I'll look into that too, but a null pointer check seems like cheap insurance.

I'm also in favor of hardening the API with some null-checks.
People generally expect libs to be tolerant to these kind of faults.

Regarding this PR specifically, I think -EINVAL should be returned for these null-checks.
It's also missing a Signed-off-by tag, some description to the commit.

And also (if doing this) I would try to do a complete audit of the API and see where we'd need to add more checks.
This only seems to fix/handle the cases which were found during some testing.

channel.c
```diff
@@ -351,6 +351,9 @@ const char * iio_channel_find_attr(const struct iio_channel *chn,
 ssize_t iio_channel_attr_read(const struct iio_channel *chn,
 		const char *attr, char *dst, size_t len)
 {
+	if (!chn)
+		return -ENOSYS;
```
commodo: -EINVAL would be more appropriate here

pcercuei:
Maybe add an error message in there then?

commodo:
> Maybe add an error message in there then?

I think just returning an err-code should be fine in the lib-part.
Apps using libiio should print error messages.

pcercuei:
Yes, just like apps using libiio should control that they don't pass null pointers :)

rgetz:
This has come up multiple times in the past. 

It was always decided (by Lars and Paul and Michael) that crashing was the best answer. It’s a bug in the application. Fix the application.   

People who write buggy code, don’t check error codes either. So silently masking things only makes the code run, but not do what they want to the hardware. 

To me, it makes more sense to add doc - when libiio crashes, fix xxxx in your application. 

My 2 cents.

commodo:
> It was always decided (by Lars and Paul and Michael) that crashing was the best answer. It’s a bug in the application. Fix the application.

I was not aware of this.
Generally speaking, "crashing the app" or "returning error codes" on faulty lib usage is a never-ending discussion of preferences.
People decide over one or the other, since none is [considerably] better than another.
If this was decided once, I don't want to change that.

Thanks & sorry for the noise.

commodo:
> > It was always decided (by Lars and Paul and Michael) that crashing was the best answer. It’s a bug in the application. Fix the application.
> 
> I was not aware of this.
> Generally speaking, "crashing the app" or "returning error codes" on faulty lib usage is a never-ending discussion of preferences.
> People decide over one or the other, since none is [considerably] better than another.
> If this was decided once, I don't want to change that.
> 
> Thanks & sorry for the noise.

Given that this has already been decided about libiio, maybe let's close this.

PR #283, opened 2019-06-23
fix install of binaries with a static only library

texierp:
Define CMAKE_INSTALL_RPATH only if ENABLE_SHARED is set otherwise the
following issue occured:

CMake Error at iiod/cmake_install.cmake:45 (file):
  file RPATH_CHANGE could not write new RPATH:

    /usr/lib

  to the file:

    /home/pjtexier/<path>/usr/sbin/iiod

  No valid ELF RPATH or RUNPATH entry exists in the file; File does not have
  a valid ELF identification.
  cmake_install.cmake:50 (include)

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>


CMakeLists.txt
```diff
@@ -15,9 +15,11 @@ endif()
 
 set(CMAKE_INSTALL_DOCDIR "" CACHE PATH "documentation root (DATAROOTDIR/doc/${PROJECT_NAME}${LIBIIO_VERSION_MAJOR}-doc)")
 include(GNUInstallDirs)
-if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-	set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_FULL_LIBDIR}")
-	set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+if(ENABLE_SHARED)
+	if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
```
pcercuei: You can make the patch smaller by just writing ```if(ENABLE_SHARED AND ${CMAKE_SYSTEM_NAME} MATCHES "Linux")```

texierp: Indeed, you're right.
Just updated, thanks !

PR #284, opened 2019-06-27
Handle channels that have no modifier or are not indexed

### LINKED ISSUE #25: Handling channels that have no modifier or are not indexed

gwendalcr:
Another attempts to fix #25

The idea is to change is_channel() from looking at index and modifier to a different format:

must start with in_ or out_
next, it should be a channel type
must have an extra '_'.
Test: Using iio_info: before:
iio:device0: cros-ec-light (buffer capable)
2 channels found:
timestamp: (input, index: 1, format: le:S64/64>>0)
illuminance: (input)
3 channel-specific attributes found:
attr 0: calibbias value: 0
attr 1: calibscale value: 56.360700
attr 2: raw value: 55

after:
iio:device0: cros-ec-light (buffer capable)
2 channels found:
illuminance: (input, index: 0, format: le:U16/16>>0)
3 channel-specific attributes found:
attr 0: calibbias value: 0
attr 1: calibscale value: 56.360700
attr 2: raw value: 52
timestamp: (input, index: 1, format: le:S64/64>>0)

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -113,7 +113,7 @@ enum iio_modifier find_channel_modifier(const char *s, size_t *len_p)
 			continue;
 		len = strlen(modifier_names[i]);
 		if (strncmp(s, modifier_names[i], len) == 0 &&
-				(s[len] == '\0' || s[len] == '_')) {
+				(s[len] == '\0' || s[len] == '_' )) {
 			if (len_p)
 				*len_p = len;
 			return i;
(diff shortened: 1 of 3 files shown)


channel.c
```diff
@@ -113,7 +113,7 @@ unsigned int find_channel_modifier(const char *s, size_t *len_p)
 			continue;
 		len = strlen(modifier_names[i]);
 		if (strncmp(s, modifier_names[i], len) == 0 &&
-				(s[len] == '\0' || s[len] == '_')) {
+				(s[len] == '\0' || s[len] == '_' )) {
```
pcercuei: Remove that added space.

channel.c
```diff
@@ -124,45 +124,43 @@ unsigned int find_channel_modifier(const char *s, size_t *len_p)
...
 			continue;
+
 		/* Type must be followed by either a '_' or a digit */
-		if (chn->id[len] != '_' && (chn->id[len] < '0' || chn->id[len] > '9'))
+		if (name[len] != '_' && (name[len] < '0' || name[len] > '9'))
 			continue;
 
-		chn->type = (enum iio_chan_type) i;
+		return (enum iio_chan_type) i;
 	}
```
pcercuei: Add an empty line here, between the closing backet and the following return statement.

local.c
```diff
@@ -1110,21 +1110,17 @@ static int local_set_trigger(const struct iio_device *dev,
...
 		return false;
 	if (!strict)
 		return true;
-	if (*(ptr - 1) >= '0' && *(ptr - 1) <= '9')
-		return true;
-
-	if (find_channel_modifier(ptr + 1, NULL) != IIO_NO_MOD)
+	if (find_channel_type(attr + skip_prefix) != IIO_CHAN_TYPE_UNKNOWN)
 		return true;
 	return false;
```
pcercuei: Maybe simplify with ``` return find_channel_type(attr + skip_prefix) != IIO_CHAN_TYPE_UNKNOWN;```

pcercuei:
I fear that this will break the shared attributes. If you have a sysfs file "in_voltage_scale_available" and a sysfs file "in_voltage0_scale", then the former is an attribute that's shared among all in_voltageX channels. Libiio then adds a "scale_available" attribute to all the voltageX input channels.

With this code, the "in_voltage_scale_available" would be improperly detected as a channel instead.

rgetz:
shared attributes are used in a lot of devices. @gwendalcr - did you test this patch with a complex device?

rgetz:
issue #25, which this patch is trying fix, was fixed with #366; so I don't think this is needed anymore. Since there have not been updates in 6 months, I will close. If you think it's still needed, please let us know.

PR #289, opened 2019-07-12
iio_attr : Add a feature which auto-generates C code

rgetz:
A few people have said that they need an on-ramp for libiio,
so make it easy to generated little examples that show how to
read/write attributes.


rgetz:
Now doing things like: 
```
./tests/iio_attr -a -q -g /tmp/foo.c -d ad9361-phy xo_correction 40000003
```
will generate code in /tmp/foo.c

```
#include <stdio.h>
#include <iio.h>

int main(int argc, char **argv)
{
        struct iio_context *ctx;
        struct iio_device *dev;
        struct iio_channel *ch;
        const char* val_str;
        ssize_t ret;
        char buf[256];

        /* Create IIO Context */
        ctx = iio_create_context_from_uri("usb:1.3.5");

        /* Find IIO device in current context */
        dev = iio_context_find_device(ctx, "ad9361-phy");

        /* Write null terminated string to device attribute: */
        ret = iio_device_attr_write(dev, "xo_correction", "40000003");
        printf("Wrote %li bytes\n", ret);

        /* Read IIO device Attribute, and put result in string */
        ret = iio_device_attr_read(dev, "xo_correction", buf, sizeof(buf)); 
        printf("xo_correction : %s\n", buf);

        iio_context_destroy(ctx);
        return EXIT_SUCCESS;
}
```

which will compile ( gcc /tmp/foo.c -o foo -liio ) and run.

Hopefully this is a easy on ramp for people creating their own applications.

tests/iio_attr.c
```diff
@@ -126,6 +146,8 @@ static struct iio_context * autodetect_context(void)
 		printf("Using auto-detected IIO context at URI \"%s\"\n",
 				iio_context_info_get_uri(info[0]));
 		ctx = iio_create_context_from_uri(iio_context_info_get_uri(info[0]));
+		if(gen_code)
```
commodo: it's a good idea to keep the code style in-sync with the rest of the file;
for this file there are spaces betweeb `if` and parentheses;
`if(gencode)` -> `if (gencode)`

tests/iio_attr.c
```diff
@@ -367,26 +471,27 @@ static void usage(void)
 				options_descriptions[i]);
 		if (i == 3)
 			printf("Optional qualifiers:\n");
-		if (i == 6)
+		if (i == 8)
```
commodo: the `6` -> `8` change looks a bit fuzzy;
i mean, i could probably find out what it means, but it takes some more digging through the file

tests/iio_attr.c
```diff
@@ -503,6 +627,12 @@ int main(int argc, char **argv)
 			fprintf(stderr, "Too many options for searching for buffer attributes\n");
 			return EXIT_FAILURE;
 		}
+		if (gen_code && !attr_index) {
+			printf("When generating code for Buffer Attributes, must include specific attribute\n"
+					"-B [IIO_device] [IIO_buffer_attribute] [value]\n");
+return EXIT_FAILURE;
```
commodo: this looks un-indented?

tests/iio_attr.c
```diff
@@ -528,21 +663,33 @@ int main(int argc, char **argv)
...
 					 strchr(argv[channel_index], '*'))) ||
 		     (attr_index && (!strcmp(".", argv[attr_index])  ||
 				      strchr(argv[attr_index], '*'))))) {
-		printf("can't write value with wildcard match\n");
+		printf("can't %s with wildcard match\n",
+				gen_code ? "generate code" : "write value");
 		return EXIT_FAILURE;
 	}
 
+
```
commodo: nitpick: extra space?

tests/iio_attr.c
```diff
@@ -528,21 +663,33 @@ int main(int argc, char **argv)
...
+	if (gen_code) {
+		gen_fd = fopen (gen_file, "w");
+		if (!gen_fd) {
+			errnum = errno;
```
commodo: `errnum` is only used in this block
it does not seem to be needed, since it only gets `errno` assigned and then this block exits;
it could be removed;

tests/iio_attr.c
```diff
@@ -367,26 +471,27 @@ static void usage(void)
...
 }
 
 int main(int argc, char **argv)
 {
 	struct iio_context *ctx;
-	int c, option_index = 0;
+	int c, option_index = 0, errnum;
 	int device_index = 0, channel_index = 0, attr_index = 0;
-	const char *arg_uri = NULL;
+	const char *arg_uri = NULL, *gen_file = NULL;
```
commodo: `arg_uri` is always NULL

tests/iio_attr.c
```diff
@@ -528,21 +663,33 @@ int main(int argc, char **argv)
...
+	}
+
 	if (detect_context)
-		ctx = autodetect_context();
-	else if (backend == AUTO)
+		ctx = autodetect_context(gen_code, gen_fd);
+	else if (backend == AUTO) {
 		ctx = iio_create_context_from_uri(arg_uri);
-	else
+		gen_start_uri(arg_uri, gen_fd);
```
commodo: is there a point where the `(backend == AUTO)` && `gen_code` are synchronized?
would we need to make this 
```
if (gen_code)
  gen_start_uri(arg_uri, gen_fd);
```
?

commodo:
the overall idea is nice;

my only concern is that we would need a more active/dedicated maintainer for libiio to go forward;
it seems to me that [at this point in time] this is the most important part for libiio;
we keep pushing stuff here, but there is no clear DRI [Directly Responsible Individual] allocated [here] to keep things in order;

mhennerich:
@adisuciu ?

tests/iio_attr.c
```diff
@@ -154,6 +176,16 @@ static void dump_device_attributes(const struct iio_device *dev,
 		if (!quiet)
 			printf("dev '%s', attr '%s', value :",
 					iio_device_get_name(dev), attr);
+		if(fd) {
+			fprintf(fd, "\t/* Read IIO device Attribute, and put result in string */\n"
```
commodo: generally: i am thinking that it may make sense to have a generic helper (something like `iio_cgen_line()` or whatever name) that is `fprintf()` wrapper;

there's a few things that this could address:
1. the `\n` char at the end of each line
2. indentation levels, which would print multiple `\t` at the beginning of each line;

this helper can be a inline function or a precompiler macro; i guess it does not matter much what it is;
but removing some elements that are prone to minor errors [or styling] when generating code, would be an idea;

examples:
```
iio_cgen_line(fd, 1, "/* Read IIO device Attribute, and put result in string */")   // one indentation level
iio_cgen_line(fd, 2, "/* Read IIO device Attribute, and put result in string */")   // 2 indentation levels
// or
iio_cgen_line1(fd, "/* Read IIO device Attribute, and put result in string */")   // one indentation level
iio_cgen_line2(fd, "/* Read IIO device Attribute, and put result in string */")   // 2 indentation levels
```

tests/iio_attr.c
```diff
@@ -166,6 +198,16 @@ static void dump_device_attributes(const struct iio_device *dev,
 		}
 	}
 	if (wbuf) {
+		if (fd) {
+			fprintf(fd, "\t/* Write null terminated string to device attribute: */\n");
```
commodo: following comment https://github.com/analogdevicesinc/libiio/pull/289/commits/4acfcd1d6408d5ef41d363323cbc1c2e90e83504#r303320445  it looks like some of these blocks could be grouped into a macro/helper function as well;

when diff-ing the code blocks, what differs is quite often some names, and a few variables;
so, whatever the case is, I guess [for a final version] it would be nice to have some sort of

```
static void iio_cgen_dev_attr(fd, const char *op) {
        if (!fd)
               return;
	fprintf(fd, "\t/* %s IIO buffer Attribute, and put result in string */\n"
		    "\tret = iio_device_buffer_attr_%s(dev, \"%s\", buf, sizeof(buf)); \n", op, attr);
	fprintf(fd, "\t/* For other types, use:\n"
		    "\t *   ret = iio_device_buffer_attr_%s_bool(dev, \"%s\", &v_bool);\n", op, attr);
	fprintf(fd, "\t *   ret = iio_device_buffer_attr_%s_double(dev, \"%s\", &v_double);\n", op, attr);
	fprintf(fd, "\t *   ret = iio_device_buffer_attr_%s_longlong(dev, \"%s\", &v_ll);\n", op, attr);
}
```
in this example, `op == "read"` ; but i'm guessing that `op == "write"` would be fairly similar & maybe some parametrization could help

the main argument to maybe doing this [aside from code de-duplication] is reducing copy+paste errors [i've personally done enough of those to hate them enough]

tests/iio_attr.c
rgetz: when autoscanning for a context (over USB), the uri is never passed back, and isn't part of the actual context - so to keep things simple, the call togen_start_uri(arg_uri, gen_fd);  is added inside the autodetect_context function.

tests/iio_attr.c
rgetz: fixed

tests/iio_attr.c
rgetz: This was done on the initial implementation - I added a new comment.

tests/iio_attr.c
rgetz: fixed

tests/iio_attr.c
rgetz: fixed

tests/iio_attr.c
rgetz: Just keeping with the rest of the coding style.

rgetz:
> the overall idea is nice;
> 
> my only concern is that we would need a more active/dedicated maintainer for libiio to go forward;
> it seems to me that [at this point in time] this is the most important part for libiio;
> we keep pushing stuff here, but there is no clear DRI [Directly Responsible Individual] allocated [here] to keep things in order;

yes - but that is outside the scope of this patchset/pull request (isn't it?) That's just a general comment about the project in whole - isn't it?

tests/iio_attr.c
rgetz: personally - one line macro functions that just replace printf don't make it easier to read/understand the code - not sure this is a good idea or not.

There isn't any indentations but none/one. The code that it generates is all flat. There are no loops/if structures (and should never be).

tests/iio_attr.c
rgetz: helper functions - sure - i wasn't sure what I was going to end up with when I started, and it is all mostly the same.

Will do. Part assumes that the api always is pretty similar (which it is today).  As long as we all like the output (for now).

rgetz:
The one thing I don't like - is there is no return/error code checking -- so it's a pretty bad example, and I don't want people to copy/paste too much of this (without error checking) into their applications.

Thought would be to put an IIO_ASSERT as a macro, similar to what is in some of the examples/ad9371-iiostream.c examples.

Thoughts?

tfcollins:
Generally having this I think is really useful since the transceiver examples are the only really "basic" ones we have. I regard to error handling, I would just add a quick macro. The errors users are likely to come across are with changing USB uri's and with the default code the error can be cryptic.

rgetz:
Ok - I will put in the asserts too. 

I guess it also makes sense to add similar functionality to iio_read and iio_write? Or you think this Ian more of an issues for iio_attr?

?

pcercuei:
Why not move this into a new program, e.g. iio_gen_code?
Then eventually it could also generate Python, C# or Matlab code.

commodo:
> > the overall idea is nice;
> > my only concern is that we would need a more active/dedicated maintainer for libiio to go forward;
> > it seems to me that [at this point in time] this is the most important part for libiio;
> > we keep pushing stuff here, but there is no clear DRI [Directly Responsible Individual] allocated [here] to keep things in order;
> 
> yes - but that is outside the scope of this patchset/pull request (isn't it?) That's just a general comment about the project in whole - isn't it?

yes & no;
yes - it is a general comment about the project as a whole;
no - it is also related about this code: it gets merged, there's fallout [which is natural and nobody wants it], but nobody steps up to fix it

stuff like this has happened before [with the code that I submitted too here], and we can ignore it for this PR, but if we do ignore [it for this PR], we will likely ignore it for the next, and so on ;
so the question becomes: where is the cut-off point where we stop ignoring this?
[ mind you: I am not a/the person to decide any of this for libiio, I am just raising a few points that I feel would be worth raising ; feel free to disregard whatever I am saying ]

we are trying to setup something like this [DRI, but we try to use "maintainer" as a word] for Linux & no-OS, and we will see how this goes; which is also [one reason] why I brought it up here;

tests/iio_attr.c
commodo: ack;
if stuff like looks/if structures become present, we can add these type of helpers then;
`iio_cgen_line(fd, indent_level_N,)`  type of funcs seemed useful when I was writing code that outputs pretty-formatted JSON [for whatever userland app I was writing], and I recalled this stuff when I was thinking about formatting this C code gen stuff;

tests/iio_attr.c
commodo: ack

rgetz:
> we are trying to setup something like this [DRI, but we try to use "maintainer" as a word] 

yeah - I'm still missing the point. If a software project doesn't have an identified maintainer (internal or external) - that's up to Michael; you should talked to him about it.

project maintainers (IMHO) aren't "picked" - people who enjoy it, do it, and eventually become a maintainer. If you don't want to do that for libiio since you are busy on other software projects you enjoy more - that's Ok too. (but I do truly appreciate your feedback - it always makes the code better/more long term maintainable).

rgetz:
> Why not move this into a new program, e.g. iio_gen_code?
> Then eventually it could also generate Python, C# or Matlab code.

Yeah, the way things are structured now, is it would be easier to put most of the adds into a separate C file, and include that when necessary into the iio_attr or iio_read or iio_write - there is just so much other logic in those places to make sure that everything is "correct", that I didn't want to duplicate in its own utility.

Plus the "normal" use case is that people figure things out interactivity, build up a small shell script (with the standard utils), and then move to C or python.

A Nack on the entire concept is OK - rather than making something that can actually compile/run - we could just print out the equivalent C/python. That would also help what most people need.

adisuciu:
I think creating a similar feature for pushing/refilling buffers would make sense.

rgetz:
Ok - I updated things, adding support for python as well. Comments welcome on the newer structure - I think it will help adding it to other of the iio-utils.

comments appreciated.

rgetz:
failed, since I was using a function that doesn't exist on Windows - just do it without...

rgetz:
fix nitpik from someone - make the generated C code, not exceed 80 chars

tests/gen_code.c
```diff
@@ -0,0 +1,285 @@
...
+void gen_ch(const struct iio_channel *ch)
+{
+	if (!fd)
+		return;
+
+	if (lang == C_LANG) {
+		fprintf(fd, "\t/* Find the IIO %s channel in the current device */\n",
+			iio_channel_is_output(ch) ? "Output" : "Input");
+		fprintf(fd, "\tIIO_ASSERT(ch = iio_device_find_channel(dev, \"%s\", %s));\n\n",
+			iio_channel_get_name(ch), iio_channel_is_output(ch) ? "true" : "false");
```
tfcollins: Might want to use iio_channel_get_id, not sure if all channels have names.

tests/gen_code.c
```diff
@@ -0,0 +1,285 @@
...
+	} else if (lang == PYTHON_LANG) {
+		fprintf(fd, "    #Find the IIO %s channel in the current device */\n",
+			iio_channel_is_output(ch) ? "Output" : "Input");
+		fprintf(fd, "    ch = dev.find_channel('%s', %s)\n",
+			iio_channel_get_name(ch), iio_channel_is_output(ch) ? "True" : "False");
```
tfcollins: Might want to use iio_channel_get_id, not sure if all channels have names.

tests/gen_code.c
```diff
@@ -0,0 +1,285 @@
...
+	if (lang == C_LANG) {
+		fprintf(fd, "\t/* Create IIO Context */\n"
+		    "\tIIO_ASSERT(ctx = iio_create_context_from_uri(\"%s\"));\n\n", uri);
+	} else if (lang == PYTHON_LANG) {
+		fprintf(fd, "    # Create IIO Context\n"
+			    "    try:\n"
+			    "        ctx = iio.Context(\"%s\")\n", uri);
+		fprintf(fd, "    except OSError as e:\n"
+			    "        print('Unable to open context %s')\n", uri);
+		fprintf(fd, "        texit()\n\n");
```
tfcollins: exit not texit

rgetz: fixed

tests/gen_code.c
rgetz: I think this does the same as iio_attr; the pattern matching happens based on get_name, and it prints out _id...

rgetz:
updated the python output, to make things more pythonic...

tests/gen_code.c
tfcollins: I get these weird nulls sometimes with get_name
```
    # Find IIO device in current context
    dev = ctx.find_device("ad9361-phy")

    # Find the IIO Output channel in the current device
    ch = dev.find_channel("(null)", True)
    # Write string to channel attribute:
    ch.attrs["rf_port_select"].value = str("A")
    print("wrote A into rf_port_select")
```
Command: `iio_attr -o -u ip:analog -q -g /tmp/foo.py -c ad9361-phy voltage0 rf_port_select 'A'`

rgetz:
If folks are mostly OK with the state of things, I will add support to iio_readdev() and iio_writedev() this week.

-Robin

tests/gen_code.c
rgetz: OK - I will have a look - thanks for the test case.

rgetz: I think this is fixed in the latest force push

PR #291, opened 2019-07-16
travis-ci: recent upgrades in the travis-ci infrastructure caused som…

rgetz:
…e breakages in deployments

Requires adding -oHostKeyAlgorithms=+ssh-dss on a few more distributions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,14 +59,17 @@ matrix:
       env:
         - OS_TYPE=centos_docker
         - OS_VERSION=6
+        - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
     - os: linux
       env:
         - OS_TYPE=centos_docker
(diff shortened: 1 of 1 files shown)


rgetz:
If this fixes the travis failures (and I think it should), this should be merged.

pcercuei:
Maybe it would be time to update the SSH key? The DSS algorithm is rather insecure and has been marked as deprecated in OpenSSL for some time now.

rgetz:
It’s not the key - it’s the broken sftp server that akaimai wrote themselves to be more “secure”. We don’t control anything on this end. They only support dss. If there was any other way - we would.

rgetz:
This is verified to fix the Akamai deployment issues.

pcercuei:
Sounds like the beginning of a bad joke. Oh well.

PR #293, opened 2019-07-18
.travis.yml: remove/retire Trusty builds

commodo:
These have been failing sporadically in the last couple of days with:
```
W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
W: Failed to fetch https://packagecloud.io/github/git-lfs/ubuntu/dists/trusty/InRelease  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
E: Failed to fetch http://dl.hhvm.com/ubuntu/dists/trusty/main/binary-amd64/Packages  Hash Sum mismatch
W: Some index files failed to download. They have been ignored, or old ones used instead.
```

This can be worked around until Travis-CI fixes their Trusty docker images.
But maybe it's also a good idea to draw a line a remove them.

Trusty is 5 years old now.
And since this year, Ubuntu Xenial 16.04 is the default in Travis-CI:
 https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment


rgetz:
According to https://wiki.ubuntu.com/Releases

Ubuntu 14.04.6 LTS (Trusty Tahr) was end of support April 2019, and will be end of life April 2022. 

I don’t have a problem - but do we want to keep it for one more release?

I get that it has been failing - and looking up all these false failures are a pain.

commodo:
> I think this also removes clang, which caught some errors early on ...

yeah, i was also un-certain there;
i figured GCC is new enough to cover some of these errors & reduce the build matrix


> According to https://wiki.ubuntu.com/Releases
> 
> Ubuntu 14.04.6 LTS (Trusty Tahr) was end of support April 2019, and will be end of life April 2022.
> 
> I don’t have a problem - but do we want to keep it for one more release?
> 
No strong opinion from my side.

> I get that it has been failing - and looking up all these false failures are a pain.

I don't mind fixing failures at-this-point-in-time [especially if I can get around to it]. They don't seem to occur too often yet/here.
I'm trying to think of a small-enough build & deploy matrix that would cover all/most cases.

commodo:
Changelog v1 -> v2:
* added `clang` build back

Issue #272, opened 2019-05-08
Add macOS 10.14 (Mojave) to built releases


tfcollins:

mhennerich:
Now that travis-ci supports macOS 10.14 add it to the build maxtrix  and possibly drop 10.12

rgetz:
I think this is done now?

https://github.com/analogdevicesinc/libiio/blob/master/.travis.yml#L69

We are building xcode9.2 and xcode10.1

https://docs.travis-ci.com/user/reference/osx/#macos-version

says we are missing xcode11 or  xcode10.1 (macOS 10.14)

@tfcollins - any preference - I don't think it is necessary to build both.

tfcollins:
I would say xcode11 since we have builds in xcode10.X

rgetz:
I think this is fixed with the existing pull request #294- so closing.

PR #294, opened 2019-07-22
build,.travis.yml: add support for Xcode 11

commodo:
According to Travis-CI docs:
  https://docs.travis-ci.com/user/reference/osx/#macos-version

The latest is Xcode11 and it runs on MacOS 10.14 (Mojave).

Fixes https://github.com/analogdevicesinc/libiio/issues/272


commodo:
Changelog v1 -> v2:
* updated README.md

README.md
```diff
@@ -25,7 +25,8 @@ As with many open source packages, we use [GitHub](https://github.com/analogdevi
...
+| OS X                    |  [![OSX Status](https://api.travis-ci.org/analogdevicesinc/libiio.svg?branch=master&label=osx&passingTex=foo)](https://travis-ci.org/analogdevicesinc/libiio) |  OS X High Sierra <br />(v 10.14) | [![OS-X package 10.14](https://raw.githubusercontent.com/wiki/analogdevicesinc/libiio/img/osx_box.png)](http://swdownloads.analog.com/cse/travis_builds/master_latest_libiio-darwin-10.14.4.pkg) | [![OS-X tarball 10.14](https://raw.githubusercontent.com/wiki/analogdevicesinc/libiio/img/osx_box.png)](http://swdownloads.analog.com/cse/travis_builds/master_latest_libiio-darwin-10.14.4.tar.gz) |
```
rgetz: I think that 10.14.4 is Mojave, not High Sierra - so the name needs an update. 

https://en.wikipedia.org/wiki/MacOS_Mojave

commodo:
Changelog v1 -> v2:
* fixed `High Sierra` -> `Mojave` name

README.md
commodo: updated;
thanks for catching this;
i think we would really need to see about a more dynamic update-thingi for this readme;
i'll try to think of something that scales;
maybe something that auto-generates from `.travis.yml`

rgetz: There isn't a way to codify the xcode version in travis-ci to the macos name; other than manually editing some file somewhere.

But I get your point. It's been on my list for awhile to have the ci system automatically generate in index.html file which points to everything in a flat list - but I don't think this would have the actual names (which is what most end users key off of)... I'm still noodling on that too...

PR #300, opened 2019-08-13
Udev rules are now also installed with respect to CMAKE_INSTALL_PREFIX

AdrK:
When using a CMAKE_INSTALL_PREFIX_PATH, the CMakeLists.txt didn't respected this var.

Example command to reproduce:
mkdir -p build/output && cmake -G "Unix Makefiles" -B ./build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX:PATH=${PWD}/build/output/ -S . && make -C build -j8 install

Not sure if this is on purpose or a bug, so please advice. It works for my case though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -468,7 +468,7 @@ if (WITH_USB_BACKEND AND CMAKE_SYSTEM_NAME MATCHES "^Linux")
 	option(INSTALL_UDEV_RULE "Install a udev rule for detection of USB devices" ON)
 
 	if (INSTALL_UDEV_RULE)
-		set(UDEV_RULES_INSTALL_DIR /lib/udev/rules.d CACHE PATH "default install path for udev rules")
+		set(UDEV_RULES_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/lib/udev/rules.d" CACHE PATH "default install path for udev rules")
 
 		configure_file(libiio.rules.cmakein ${CMAKE_CURRENT_BINARY_DIR}/90-libiio.rules @ONLY)
 		install(FILES ${CMAKE_CURRENT_BINARY_DIR}/90-libiio.rules DESTINATION ${UDEV_RULES_INSTALL_DIR})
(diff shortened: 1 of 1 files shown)


pcercuei:
It's not a bug, it's because udev looks for its config files inside /lib, and not /usr/lib...

rgetz:
@AdrK what distribution are you having issues on?

Like @pcercuei indicated - From: [doc](https://linuxconfig.org/tutorial-on-how-to-write-basic-udev-rules-in-linux):
```
There are two main locations for udev rules can be placed:
 /usr/lib/udev/rules.d it's the directory used for system-installed rules, 
 /etc/udev/rules.d/ is reserved for custom made rules. 
```

[debian](https://wiki.debian.org/udev) says:
```
 The rules files (which amount to more configuration for udevd) are taken from 
 /run/udev/rules.d, 
 /etc/udev/rules.d or 
 /lib/udev/rules.d 
(debian puts most of the rules in /lib/udev/rules.d).
```
None of those will be CMAKE_INSTALL_PREFIX

AdrK:
Hi,
Thanks for response.
Maybe I should explain a use case little bit more. Point is that I was trying to install liibiio locally, for development purposes. To do so, I have used '-DCMAKE_INSTALL_PREFIX' parameter for cmake. I noticed then that all files are being installed in directory that I have specified, except the udev rules. cmake was still trying to install them in /lib path. Which obviously fail with 'permission denied' on my local machine. Udev rules are unconditionally installed in /lib. For that specific reason I found this as a bug, as prefix I provided is not respected.
In second commit I have provided a fix that seems to be working fine for both cases: local installation, and system installation.
When checking for 'CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT' I can determine if the path of udev installation shall be prepended with prefix or not.

For me this is working, so it is no more issue for me.

rgetz:
I think we just suggest for people want to install in a non-standard place, and not to install the udev rules with something like:
```
cmake  -DCMAKE_INSTALL_PREFIX=/opt/install/libiio -DINSTALL_UDEV_RULE=OFF
```
Does that work for you?

rgetz:
OK - since there have't been any updates - I will close this - If you think this is still an issue - please let us know.

Issue #302, opened 2019-08-19
Wrong argument help in example.


dimasad:
In the examples/dummy-iiostream.c example, the help for the read method is 2 read, 3 read raw, while the order of the enum makes it 2 read raw, 3 read.

rgetz:
yeah, it looks like:

https://github.com/analogdevicesinc/libiio/blob/master/examples/dummy-iiostream.c#L159
```
	printf("  -r\tread method (default 0 pointer, 1 callback, 2 read, 3 read raw)\n");
```

and 
https://github.com/analogdevicesinc/libiio/blob/master/examples/dummy-iiostream.c#L94
```
enum {
	BUFFER_POINTER,
	SAMPLE_CALLBACK,
	CHANNEL_READ_RAW,
	CHANNEL_READ,
	MAX_READ_METHOD,
};
```

is in the incorrect order. I will update, and check the others.

Thanks for pointing out.

rgetz:
added a fix - should be resolved on the next release.

rgetz:
closing this - since the PR/merge is in progress.

Thanks again for pointing this out.

PR #304, opened 2019-08-19
Fixes #303 and #302

### LINKED ISSUE #303: printf types in dummy-iiostream example not platform independent
The examples/dummy-iiostream.c printf types are not platform independent, and do generate compilation error in the armhf architecture, as in the BeagleBone.

dimasad:
Fixes to the dummy-iiostream example.

### DIFF
--- a/examples/dummy-iiostream.c
+++ b/examples/dummy-iiostream.c
@@ -156,7 +156,7 @@ static void usage(int argc, char *argv[])
 	printf("  -d\tdevice name (default \"iio_dummy_part_no\")\n");
 	printf("  -t\ttrigger name (default \"instance1\")\n");
 	printf("  -b\tbuffer length (default 1)\n");
-	printf("  -r\tread method (default 0 pointer, 1 callback, 2 read, 3 read raw)\n");
+	printf("  -r\tread method (default 0 pointer, 1 callback, 2 read raw, 3 read)\n");
 	printf("  -c\tread count (default no limit)\n");
 }
 
--- a/examples/dummy-iiostream.c
+++ b/examples/dummy-iiostream.c
@@ -76,6 +76,7 @@
 #include <stdio.h>
 #include <errno.h>
 #include <getopt.h>
+#include <inttypes.h>
 
 #ifdef __APPLE__
 #include <iio/iio.h>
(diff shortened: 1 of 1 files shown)


rgetz:
Couple comments 

- we always use signed-off-by on commits (this will need to be added). to the first commit.

- int64_t is guaranteed by the C99 standard to be exactly 64 bits wide, long long normally is, but not always - and can cause problems in some places in this codebase. As long as your compiler supports C99, I don't know why this is necessary. To be truly portable - we should be using the Format macro constants, so it would look like: 
```
uint64_t t;
printf("%" PRIu64 "\n", t);
```

-Robin

dimasad:
I used long long because it is guaranteed to be *at least* 64 bits wide, which is enough to prevent overflow. I was using gcc on the BeagleBone Black, and the specifier "%ld" with int64_t did not work. In that architecture, a long is 32 bits and overflow occurs.

Anyways, format macros can lead to a more elegant solution. I'll change that and add the sign-off.

mhennerich:
Can you please add your signed-of-by tag to the commit message and then push again?
Without we can't merge this PR.

dimasad:
Sorry, I totally forgot about this. Will have time to do this Friday 2019-11-29.

dimasad:
> To be truly portable - we should be using the Format macro constants, so it would look like:
> 
> ```
> uint64_t t;
> printf("%" PRIu64 "\n", t);
> ```

@rgetz, I updated to your truly portable solution.

pcercuei:
Just a nitpick - you use ```PRI "i" 16``` and ```PRI "d" 64```, both variants are fine, but you could use the same one to be consistent :)

dimasad:
> Just a nitpick - you use `PRI "i" 16` and `PRI "d" 64`, both variants are fine, but you could use the same one to be consistent :)

Well, I also am very OCD about these things. Anyways, the previous coder of the example used both `%i` and `%d`. So what I did was replace all `%d` with `PRIdXX` and `%i` with `PRIiXX`. Do you still want me to change it or is my choice reasonable enough?

pcercuei:
Up to you.

dimasad:
I would rather just leave it like it is then.

PR #312, opened 2019-09-17
CMake: added option OSX_INSTALL_FRAMEWORKSDIR (default /Library/Frameworks)

matejk:
CMake: added option OSX_INSTALL_FRAMEWORKSDIR (default /Library/Frameworks) to specify custom install path. rpath of tools is corrected to reference the library relative to the binary's location.

This is useful when creating a build that is deployed on a computer other than where libiio was built. Corrected rpaths of tools also allow to execute tools directly from the framework regardless of where the framework resides.


rgetz:
Thanks - this makes sense to me, @tfcollins is more of the MAC person - I know there are some weird rpath things we do on deployment, that he worked on to make sure it works in various corner cases.

PR #314, opened 2019-09-17
Python: More robust loading of IIO library

matejk:
Python: More robust loading of IIO library that works on different platforms.

This improvement is based on work of Emmanuel Blot (eblot).

Tested-by: Matej Kenda <matejken@gmail.com>

### DIFF
--- a/bindings/python/iio.py
+++ b/bindings/python/iio.py
@@ -16,6 +16,7 @@
 from ctypes import Structure, c_char_p, c_uint, c_int, c_size_t, \
 		c_ssize_t, c_char, c_void_p, c_bool, create_string_buffer, \
 		POINTER as _POINTER, CDLL as _cdll, memmove as _memmove, byref as _byref
+from ctypes.util import find_library
 from os import strerror as _strerror
 from platform import system as _system
 import weakref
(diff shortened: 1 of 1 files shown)


rgetz:
Thanks for pushing upstream.

I think we need a signed-off-by @eblot before we can commit. (if you only tested it?)

eblot:
I can do that if you want me to, but I do not know how to sign at this stage. Any pointer to how to proceed would be warmly welcomed. Thanks.

mhennerich:
If you would have crated the PR - it would be as simple as amend commit and push.
The PR would update. 
Maybe you update your branch - and @matejk merges it?

matejk:
Would it work if I amend the commit and push it if @eblot agrees?

mhennerich:
I think so

eblot:
I do agree, it is really such a small patch... the simpler the better.

matejk:
Agreed changes of signed-off a few months ago. I hope it is OK to merge.

mhennerich:
looks good to me now - but adding some more reviewers.

commodo:
the code looks ok;
the commits look a bit weird;
i see 3 commits for the same change [1 merge commit + 2 other]

it would be nice to have a single one, but fine to leave it as-is, if the intent is to preserve hashes

mhennerich:
merged

PR #320, opened 2019-11-05
utilities.c: Consider OS X a platform that has locale support

dNechita:
__MACH__ is added to exclude iOS as it hasn't been tested whether iOS has locale support or not.

### DIFF
--- a/utilities.c
+++ b/utilities.c
@@ -28,7 +28,9 @@
 #include <stdlib.h>
 #include <string.h>
 
-#if defined(_WIN32) || (defined(__USE_XOPEN2K8) && \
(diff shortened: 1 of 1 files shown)


commodo:
Just as a note: maybe it's a good idea to reference here any [in the commit and/or PR] any issue that it is fixing. Maybe reference a Github issue, or something reported on EZ.

The change in itself looks good, but I am not sure whether this relates to a recent Github issue with French locale.

dNechita:
This addresses the issue reported for iio-oscilloscope: https://github.com/analogdevicesinc/iio-oscilloscope/issues/194

PR #328, opened 2019-12-11
gh-327: Use "include(FindPythinInterp)" when compiling with CMake older than 3.12

### LINKED ISSUE #327: Python not found in cmake anymore
Reported: https://ez.analog.com/linux-device-drivers/linux-software-drivers/f/q-a/119611/linux-libiio-python-bindings

I believe this is the problem: https://github.com/analogdevicesinc/libiio/blob/master/bindings/python/CMakeLists.txt#L4 and older cmake versions. The FindPython module doesn't seem to exist prior to cmake 3.12.

Bumping cmake versions would cause more issues since Ubuntu 18.04 ships with CMake 3.10.

matejk:
Fixes #327.

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -1,7 +1,16 @@
 cmake_minimum_required(VERSION 2.8.7)
 project(libiio-py NONE)
 
-find_package (Python COMPONENTS Interpreter)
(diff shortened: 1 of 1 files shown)


rgetz:
This looks good to me. Thanks for the fix.

tfcollins:
This fixed worked on Ubuntu 18.04 for me.

PR #332, opened 2020-01-07
Python: Fixed loading of libiio on non-Windows systems 

matejk:
Fixed loading of libiio on non-Windows systems where find_library shall find it in the path if the specified name is 'iio'.

Fixes error when importing module iio on Linux. 

### DIFF
--- a/bindings/python/iio.py
+++ b/bindings/python/iio.py
@@ -61,10 +61,9 @@ _BufferPtr = _POINTER(_Buffer)
 
 if 'Windows' in _system():
 	_iiolib = 'libiio.dll'
-elif 'Darwin' in _system():
-	_iiolib = 'iio'
 else:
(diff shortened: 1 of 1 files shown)


pcercuei:
I guess the new code works, but I wonder, what problem did you have?

matejk:
The code worked properly on MacOS and Windows. I did not verify it on Linux when including the fix from another cloned repo because I assumed that the if-else statements for different platforms were there like that for a purpose.

Unfortunately the library was not found by its full name on Linux (Ubuntu 16.04, for example).

Python's find_library shall have the same argument ('iio') on Posix platforms, because it adds prefix lib and proper suffixes when searching for the library.

So the correct solution is to use short library name ("iio") on all platforms except Windows where find_library would expect "iio.dll" for such library name and therefore "libiio.dll" must be used.

PR #333, opened 2020-01-07
Python3: fixed string type checking in iio.NetworkContext to be compa…

matejk:
Fixed string type checking in iio.NetworkContext to be compatible with Python 2 and Python 3.

### DIFF
--- a/bindings/python/iio.py
+++ b/bindings/python/iio.py
@@ -39,6 +39,15 @@ def _checkNegative(result, func, arguments):
 	else:
 		raise OSError(-result, _strerror(-result))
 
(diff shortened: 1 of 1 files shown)


pcercuei:
This makes Codacity unhappy.

Maybe try ```eval('basestring')``` instead of just ```basestring```?

matejk:
The solution is based on the quite long discussion in this SO thread:

https://stackoverflow.com/questions/11301138/how-to-check-if-variable-is-string-with-python-2-and-3-compatibility

matejk:
New implementation that passes Codacity check is provided.

PR #334, opened 2020-01-10
fix #302, wrong argument in usage in dummy-iiostream

### LINKED ISSUE #302: Wrong argument help in example.

rgetz:

### DIFF
--- a/examples/dummy-iiostream.c
+++ b/examples/dummy-iiostream.c
@@ -99,6 +99,13 @@ enum {
 	MAX_READ_METHOD,
 };
 static int buffer_read_method = BUFFER_POINTER;
+// these strings must be in the same order as the above enum
+static const char read_method_str[MAX_READ_METHOD][20] = {
+	"pointer",
+	"callback",
+	"read raw",
+	"read",
+};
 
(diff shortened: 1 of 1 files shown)


examples/dummy-iiostream.c
```diff
@@ -99,6 +99,13 @@ enum {
 	MAX_READ_METHOD,
 };
 static int buffer_read_method = BUFFER_POINTER;
+// these strings must be in the same order as the above enum
+static char read_method_str[MAX_READ_METHOD][20] = {
```
mhennerich: use:
static const char * const read_method_str[] = {

rgetz: changed

dNechita:
There's a small conflict that needs to be handled.

rgetz:
looks like I based this off an old unfixed/stale copy of master I had - master was fixed in August, and this patch isn't necessary (was fixed a different way). Sorry for the noise.

PR #343, opened 2020-01-16
fix #336, remove warnings on FreeBSD

### LINKED ISSUE #336: warnings on FreeBSD
based on debian testing, on (kfreebsd-amd64) 
https://buildd.debian.org/status/fetch.php?pkg=libiio&arch=kfreebsd-amd64&ver=0.18-1&stamp=1565884272&raw=0

```
/usr/bin/cc -DLIBIIO_EXPORTS=1 -DMATLAB_BINDINGS_API=1 -DNETWORK_BACKEND=1 -DSERIAL_BACKEND=1 -DXML_BACKEND=1 -D_POSIX_C_SOURCE=200809L -D__XSI_VISIBLE=500 -Diio_EXPORTS -I/<<PKGBUILDDIR>> -I/<<PKGBUILDDIR>>/build -I/usr/include/libxml2  -g -O2 -fdebug-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fvisibility=hidden -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wpedantic -fPIC   -std=c99 -o CMakeFiles/iio.dir/network.c.o   -c /<<PKGBUILDDIR>>/network.c
/<<PKGBUILDDIR>>/network.c: In function ‘create_cancel_fd’:
/<<PKGBUILDDIR>>/network.c:240:8: warning: implicit declaration of function ‘pipe2’; did you mean ‘pipe’? [-Wimplicit-function-declaration]
  ret = pipe2(io_ctx->cancel_fd, O_CLOEXEC | O_NONBLOCK);
        ^~~~~
        pipe
```

based on the freebsd pipe2 man page,https://www.freebsd.org/cgi/man.cgi?pipe(2) 
```
       #include <unistd.h>
       int pipe2(int pipefd[2], int flags);
```

which we are - but we are still getting the warning...

For Linux - pipe2 only shows up with _GNU_SOURCE, which we only do on Linux, not on freebsd (so I'm not familiar enough with freebsd to know if that is the problem or not).

-Robin

rgetz:
Previously, building libiio on FreeBSD would warn with:
network.c:240:8: warning: implicit declaration of function 'pipe2' is invalid in C99

This fixes that, since pipe2 (inside <unistd.h>) is protected by __BSD_VISIBLE. 
On Linux, it's protected by _GNU_SOURCE (which is already defined in the cmake).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,12 @@ IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 	add_definitions(-D_GNU_SOURCE=1)
 endif()
 
+# For BSD variants
+if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD|DragonFly|OpenBSD|NetBSD")
+	set(CMAKE_REQUIRED_DEFINITIONS "-D__BSD_VISIBLE")
+	add_definitions(-D__BSD_VISIBLE=1)
+endif()
+
 option(ENABLE_IPV6 "Define if you want to enable IPv6 support" ON)
 if (ENABLE_IPV6)
 	check_symbol_exists(in6addr_any "netinet/in.h" HAVE_IPV6)


rgetz:
This should likely be:

 if(CMAKE_SYSTEM_NAME MATCHES "FreeBSD|DragonFly|OpenBSD|NetBSD")

To catch all versions of BSD.

pcercuei:
Alternatively, define this macro unconditionally, it should have zero effect on non-BSDs.

rgetz:
Ok - updated. I left the conditional in there - just to stop pollution...

pcercuei:
What about MidnightBSD? :)

rgetz:
Even LLVM nacked that for them : https://reviews.llvm.org/D51577
I think they are using FreeBSD now - aren't they? 

The ones I added were the ones tracked at:
https://github.com/freebsd/freebsd/blob/master/share/misc/bsd-family-tree

PR #345, opened 2020-01-17
iio_attr: Anticipate the content of an iio attribute to be larger

### LINKED ISSUE #268: Device attributes returning multi-line strings can cause error -EIO
Device attributes returning multi-line strings can cause error -EIO (-5) when read via IIOD.

dev 'ad9361-phy', attr 'gain_table_config', value :ERROR: Input/output error (-5) 

This issue was reported here:
https://ez.analog.com/linux-device-drivers/linux-software-drivers/f/q-a/106570/gain-table-config

dNechita:
Partially fixes #268

### DIFF
--- a/tests/iio_attr.c
+++ b/tests/iio_attr.c
@@ -148,7 +148,7 @@ static void dump_device_attributes(const struct iio_device *dev,
 		const char *attr, const char *wbuf, bool quiet)
 {
 	ssize_t ret;
-	char buf[1024];
+	char buf[2048];
 
 	if (!wbuf || !quiet) {
 		if (!quiet)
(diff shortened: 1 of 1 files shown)


rgetz:
since this is the iio-utils, and not the core library - why not just malloc 16k, and free it when it is done? Never run into a problem again, or at least make it a #define, so it's easier to make bigger later...

@mhennerich : What is the largest kernel attribute size that anything should return?

rgetz:
I added #348 that implements the malloc change I was talking about - it might be more robust long term.

dNechita:
The other PR is more complete. Will close this one.

PR #347, opened 2020-01-17
Add Cmake flags to turn on/off bindings

rgetz:
Many distributions have patches to change the CMake to turn off bindings
This internalize that to the upstream package in a sustainable manner.

The flags default to on, so it will not change the default behaviour.
Flag names are borrowed from libm2k

### DIFF
--- a/CI/travis/make_darwin
+++ b/CI/travis/make_darwin
@@ -3,11 +3,11 @@
 if [ "x${COVERITY_SCAN_PROJECT_NAME}" != "x" ] ; then exit 0; fi
 
 cd $TRAVIS_BUILD_DIR/build
-cmake -DOSX_PACKAGE=ON ..
+cmake -DOSX_PACKAGE=ON -DPYTHON_BINDINGS=ON ..
 make
 ls
 
 cd $TRAVIS_BUILD_DIR/build_tar
-cmake -DOSX_PACKAGE=OFF -DENABLE_PACKAGING=ON ..
+cmake -DOSX_PACKAGE=OFF -DENABLE_PACKAGING=ON -DPYTHON_BINDINGS=ON ..
 make && make package
 ls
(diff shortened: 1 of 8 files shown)


pcercuei:
There are already the ```PYTHON_BINDINGS``` and ```CSHARP_BINDINGS``` options for that. What is wrong right now, is that the value of these options depend on the environment (presence of python / C# interpreter).

rgetz:
Thanks for the look - I moved the existing bindings up to the top level CMake. (and added one for the MATLAB bindings)

- Robin

pcercuei:
That's a good start, but you can go a bit further, e.g. in bindings/python/CMakeLists.txt you don't need to check for ```PYTHON_BINDINGS``` anymore; and if the Python interpreter is not found you can just issue an error. Same for the other bindings.

rgetz:
the way it is now, is it tries, and if it doesn't find the Python interpreter, it just doesn't build the bindings, but it doesn't effect the building of the library.

I think you are suggesting that we should default off, and if someone turns it on, if we don't find the interpreter - it should fail inside the CMake?

```
message(FATAL_ERROR "Can not build Python bindings, since could not find Python binary" ...)
```

that makes sense to me, but does change the default build behavior.

pcercuei:
Yes, I remember discussing this with Lars - the current behaviour of the cmake script is wrong, because features get enabled according to what's found in the build environment - and that's a nightmare for distribution packagers, that's why they have these patches on top.

rgetz:
OK - will do later tonight - thanks for the feedback.

rgetz:
Ok - if this looks correct - let me know, and I will add a separate commit to this pull request, which updates all the CI to turn on PYTHON_BINDINGS or CSHARP_BINDINGS based on the build environment...

rgetz:
OK - appveyor is failing, since I didn't turn on the C# bindings, and the build expects those (which is fine), just don't merge yet...

rgetz:
OK - I think this is all updated. CI will say for sure :)

rgetz:
I think this is good to approve/merge when folks have a chance.

PR #351, opened 2020-01-22
Improve URI Documentation for iio_create_context_from_uri

ekigwana:
Previously, one would have to check source code to determine URI
types and address part formatting.

Signed-off-by: Edward Kigwana <ekigwana@gmail.com>

### DIFF
--- a/iio.h
+++ b/iio.h
@@ -341,7 +341,25 @@ __api struct iio_context * iio_create_network_context(const char *host);
 /** @brief Create a context from a URI description
  * @param uri A URI describing the context location
  * @return On success, a pointer to a iio_context structure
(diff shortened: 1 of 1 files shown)


pcercuei:
Documentaion -> Documentation in the commit title :p 
Otherwise looks really good.

tfcollins:
Should there be a line about zeroconf?

rgetz:
👍 from me. Thanks for your contribution.

As @tfcollins indicates, "ip:" will work, as long as avahi or Bonjour is found. This will go out and look for any device on the network with a _iio._tcp service being advertised.

ekigwana:
@tfcollins, I have documented zeroconf as it seems really useful when dealing with one device on a network

PR #353, opened 2020-01-25
Promote with_doc

rgetz:
This shuffles around some of the top level cmake, so the doc isn't build if Doxygen is found - it needs to be turned on.

-Robin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -380,12 +380,12 @@ if(NOT SKIP_INSTALL_ALL)
 		PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 endif()
 
-find_package(Doxygen)
-if(DOXYGEN_FOUND)
-	option(WITH_DOC "Generate documentation with Doxygen" ON)
-
-	# It is not an error when 'dot' is not found, just switching off the Doxygen's HAVE_DOT option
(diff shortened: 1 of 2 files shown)


rgetz:
Things are working on the CI properly - so this is the version to commit (unless people have comments?)

CMakeLists.txt
```diff
@@ -380,21 +380,20 @@ if(NOT SKIP_INSTALL_ALL)
...
-
-	include(cmake/CheckCaseSensitiveFileSystem.cmake)
-	if (HAVE_CASE_SENSITIVE_FILESYSTEM)
-		set(CMAKE_CASE_SENSITIVE_FILESYSTEM "YES")
-	else()
-		set(CMAKE_CASE_SENSITIVE_FILESYSTEM "NO")
-	endif()
+option(WITH_DOC "Generate documentation with Doxygen" OFF)
+if(WITH_DOC)
+	find_package(Doxygen)
```
dNechita: Alternatively, you could add the REQUIRED option to find_package(). If doxygen isn't found, it will fail at this point, and print an error message. And it won't be necessary to interrogate DOXYGEN_FOUND and to print a FATAL_ERROR message.
This would simplify the code, functionality remains the same.

rgetz: Done

rgetz:
Updated based on feedback

Issue #352, opened 2020-01-23
Global attributes for differential channels


fpagliughi:
I'm playing with a couple types of little SPI devices that each have a single differential voltage input, such as the MPC3201 or the TI ADC0831. These have a single differential voltage channel and timestamp channel:
```
$ ls -1 /sys/bus/iio/devices/iio:device2/scan_elements
in_timestamp_en
in_timestamp_index
in_timestamp_type
in_voltage0-voltage1_en
in_voltage0-voltage1_index
in_voltage0-voltage1_type
```
But the scale attribute doesn't follow the channel naming convention, dropping the '0' and '1':
```
$ ls -1 /sys/bus/iio/devices/iio:device2
buffer
current_timestamp_clock
dev
in_voltage0-voltage1_raw        <-- Same channel, but
in_voltage-voltage_scale        <-- different naming convention
name
...
```
Notice that it's `in_voltage-voltage_scale` instead of `in_voltage0-voltage1_scale`.

This is causing libiio to interpret it as the attributes for a different channel (from iio_info):
```
	iio:device2: adc0831 (buffer capable)
		3 channels found:
			voltage0-voltage1:  (input, index: 0, format: le:U8/8>>0)
			1 channel-specific attributes found:
				attr  0: raw value: 24
			timestamp:  (input, index: 1, format: le:S64/64>>0)
			voltage-voltage:  (input)
			1 channel-specific attributes found:
				attr  0: scale value: 16.000000000
		1 device-specific attributes found:
				attr  0: current_timestamp_clock value: realtime
```
It interprets it as a different channel, `voltage-voltage:  (input)`,
claiming 3 channels found, instead of 2.

So I can't tell if this is a problem in the kernel or a weird corner case for libiio to handle.

fpagliughi:
Or is this a generic way to say, "all differential channels share this scaling," even though there's only one channel?

pcercuei:
Is that with an upstream driver? I never saw this "voltage0-voltage1" thing before.
Usually, when you have a "in_voltage_scale" attribute, that means it is shared across all "iio_voltageX" channels.

fpagliughi:
So when libiio sees an attribute like "in_voltage_scale", does it create a "scale" attribute for each voltage channel? Or does it create an additional channel called "voltage" and then the application needs to look for it and apply it manually?

[...]

PR #354, opened 2020-01-25
#352 Match global attributes for differential channels.

fpagliughi:
Signed-off-by: fpagliughi <fpagliughi@mindspring.com>

This probably isn't a complete solution for release, but does seem to fix the limited case of some differential channels. I also wasn't sure when some attributes might be private (if I should look for a special case to return 2).

### DIFF
--- a/local.c
+++ b/local.c
@@ -1483,15 +1483,15 @@ static unsigned int is_global_attr(struct iio_channel *chn, const char *attr)
 	// Check for matching global differential attr, like "voltage-voltage"
 	dashptr = strchr(attr, '-');
 	if (dashptr && dashptr > attr && dashptr < ptr) {
-		unsigned int len1 = dashptr - attr,
-					 len2 = ptr - dashptr - 1;
+		unsigned int len1 = dashptr - attr;
+		unsigned int len2 = ptr - dashptr - 1;
 		const char*  iddashptr = strchr(chn->id, '-');
-		if (iddashptr && strlen(iddashptr+1) > len2 &&
+		if (iddashptr && strlen(iddashptr + 1) > len2 &&
 			iddashptr - chn->id > len1 &&
 			chn->id[len1] >= '0' && chn->id[len1] <= '9' &&
 			!strncmp(chn->id, attr, len1) &&
-			iddashptr[len2+1] >= '0' && iddashptr[len2+1] <= '9' &&
-			!strncmp(iddashptr+1, dashptr+1, len2))
+			iddashptr[len2 + 1] >= '0' && iddashptr[len2 + 1] <= '9' &&
+			!strncmp(iddashptr + 1, dashptr + 1, len2))
 			return 1;
 	}
 
--- a/local.c
+++ b/local.c
@@ -1465,7 +1465,7 @@ static int add_channel(struct iio_device *dev, const char *name,
 static unsigned int is_global_attr(struct iio_channel *chn, const char *attr)
 {
 	unsigned int len;
-	char *ptr;
+	char *ptr, *dashptr;
 
 	if (!chn->is_output && !strncmp(attr, "in_", 3))
 		attr += 3;
@@ -1479,6 +1479,19 @@ static unsigned int is_global_attr(struct iio_channel *chn, const char *attr)
 		return 0;
 
 	len = ptr - attr;
+
+	// Check if matching global differential attr,
+	// like "voltage-voltage"
+	dashptr = strchr(attr, '-');
+	if (dashptr && dashptr > attr && dashptr < ptr) {
+		unsigned int len1 = dashptr - attr,
+					 len2 = ptr - dashptr - 1;
+		if (chn->id[len1] >= '0' && chn->id[len1] <= '9' &&
+			!strncmp(chn->id, attr, len1) &&
+			chn->id[len+1] >= '0' && chn->id[len+1] <= '9' &&
+			!strncmp(chn->id+len1+2, attr+len1+1, len2))
+			return 1;
+	}
 
 	if (strncmp(chn->id, attr, len))
 		return 0;
--- a/local.c
+++ b/local.c
@@ -1480,16 +1480,18 @@ static unsigned int is_global_attr(struct iio_channel *chn, const char *attr)
 
 	len = ptr - attr;
 
-	// Check if matching global differential attr,
-	// like "voltage-voltage"
+	// Check for matching global differential attr, like "voltage-voltage"
 	dashptr = strchr(attr, '-');
 	if (dashptr && dashptr > attr && dashptr < ptr) {
 		unsigned int len1 = dashptr - attr,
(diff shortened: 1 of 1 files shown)


rgetz:
Yeah, according to [the kernel doc](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio) we should be supporting names like:

```
What:		/sys/bus/iio/devices/iio:deviceX/in_voltageY-voltageZ_raw
KernelVersion:	2.6.35
Contact:	linux-iio@vger.kernel.org
Description:
		Raw (unscaled) differential voltage measurement equivalent to
		channel Y - channel Z where these channel numbers apply to the
		physically equivalent inputs when non differential readings are
		separately available. In differential only parts, then all that
		is required is a consistent labeling.  Units after application
		of scale and offset are millivolts.
```

@mhennerich : don't we have drivers that support this?

mhennerich:
Yes there are various drivers can can have differential channels

rgetz:
Sorry / do you have a specific one that someone can set up and test this out on?

mhennerich:
https://wiki.analog.com/resources/tools-software/linux-drivers/iio-adc/ad7266

fpagliughi:
@rgetz Thanks for pointing out the kernel docs. I was just looking through this the other day, but didn't think to refer back for this issue. It looks like only voltage channels use the differential nomenclature with the minus sign. So maybe there's an easier test than the one I proposed?

As for a driver to test, as I realized by the end of #352, this issue is present in the _iio_dummy_ driver. It simulates two differential voltage channels with a global scaling attribute for the two of them. 

Using it, we see the issue of the attribute being mistaken for a separate channel, This PR shows it being applied to the two channels, instead.

rgetz:
Thanks for the pointer to the dummy driver...

It should be limited to:

in_voltageY-voltageZ_raw
in_capacitanceY-in_capacitanceZ_raw

In the example AD7266 that was pointed to by Michael:

```
-r--r--r--    1 root     root          4096 Jan  1 04:03 in_voltage0-voltage1_raw
-r--r--r--    1 root     root          4096 Jan  1 04:03 in_voltage2-voltage3_raw
-r--r--r--    1 root     root          4096 Jan  1 04:03 in_voltage4-voltage5_raw
-r--r--r--    1 root     root          4096 Jan  1 04:03 in_voltage6-voltage7_raw
-r--r--r--    1 root     root          4096 Jan  1 04:03 in_voltage8-voltage9_raw
-r--r--r--    1 root     root          4096 Jan  1 04:03 in_voltage10-voltage11_raw
```
and it has a global attribute:

```
-rw-r--r--    1 root     root          4096 Jan  1 04:03 in_voltage-voltage_scale
```

it might be more robust in your check, to look at the string before the dash, and after the dash to see if they are equal.

something like if(dashptr && !strncmp(attr, dashptr +1, len) && .... ) that way you should end up with less false positives (if someone puts a dash in the attribute name for something else).

I'm not sure about the rest - @pcercuei would have to comment on that.

fpagliughi:
@rgetz Yes, it is separately comparing the strings before and after, with each of the two strncmp() calls.

What I would need to double-check is that it is properly handles the case of channel nubers with more than one digit:
```
-r--r--r--    1 root     root          4096 Jan  1 04:03 in_voltage10-voltage11_raw
```
I think so, but I will check. 

I also wondered about buffer overflow. It seemed like the existing code wasn't checking whether the channel ID was as long or longer than the attribute name. But I might have missed that.

rgetz:
I thought you were checking something else....

Looks like you check the right side of the channel, against the right side of the global attribute, and the left side of the channel against the left side of the global attribute. You don’t check to make sure the length of the right/left sides are the same or the left side is the same as the right side.

I think that is required for a differential attribute (according to the kernel doc). 

Maybe. I’m just looking at things on my phone at a airport.

local.c
```diff
@@ -1480,6 +1480,19 @@ static unsigned int is_global_attr(struct iio_channel *chn, const char *attr)
 
 	len = ptr - attr;
 
+	// Check if matching global differential attr,
+	// like "voltage-voltage"
+	dashptr = strchr(attr, '-');
+	if (dashptr && dashptr > attr && dashptr < ptr) {
+		unsigned int len1 = dashptr - attr,
+					 len2 = ptr - dashptr - 1;
+		if (chn->id[len1] >= '0' && chn->id[len1] <= '9' &&
```
pcercuei: You should check that chn->id has at least len1 characters, otherwise you're reading garbage here.

fpagliughi: Yes! I realized that I left a buffer overrun. I was about to push another commit, but I had mistakenly assumed a single digit for X, and @rgetz reminded me that the channel index can go multiple digits. So I think I need to do two things:

1. Check that chn->id also has a dash, (if not, it's not a match at all)
2. Check that it's total length is the dash offset plus len2.

I should get an update pushed in a few hours.

rgetz: I'm not sure about checking for multiple digits. if a channel has a global attribute, it applies to voltage1 the same as voltage11 the same as voltage111.

as long as there is a single char that is a number at the end of "voltage", it should be fine. It doesn't matter how many char/numbers there are.

I think this is how the code works today...

fpagliughi: Understood. I just made an assumption about the position of the dash in the chn->id, by assuming there was one digit. This is obviously wrong, so I will fix it to search for the dash in the id and use that as the relative position to start the search for the second part.

fpagliughi:
The second commit does a better job of ensuring that the comparison on either side of the minus sign doesn't compare garbage. And it removes any assumption about number of digits by basing the second comparison on the position of the minus in each buffer.

rgetz:
This version looks good to me. If its OK with @pcercuei & @dNechita - it should get committed.

local.c
```diff
@@ -1480,6 +1480,21 @@ static unsigned int is_global_attr(struct iio_channel *chn, const char *attr)
 
 	len = ptr - attr;
 
+	// Check for matching global differential attr, like "voltage-voltage"
+	dashptr = strchr(attr, '-');
+	if (dashptr && dashptr > attr && dashptr < ptr) {
+		unsigned int len1 = dashptr - attr,
+					 len2 = ptr - dashptr - 1;
```
dNechita: Things could look more aligned if you declare len2 along with the type:
`unsigned int len2 = ptr - dashptr - 1;`

local.c
```diff
@@ -1480,6 +1480,21 @@ static unsigned int is_global_attr(struct iio_channel *chn, const char *attr)
...
+		const char*  iddashptr = strchr(chn->id, '-');
+		if (iddashptr && strlen(iddashptr+1) > len2 &&
```
dNechita: Please add spaces before and after the '+' sign here and in the lines that follow.

fpagliughi:
Done

dNechita:
Thank you for contributing!

PR #355, opened 2020-01-26
doc: update code model picture

rgetz:
This is updating to the same diagram we have been using in
various IIO presentations, and shows which have attributes.


rgetz:
The new picture looks like:

image

old picture looks like:
http://analogdevicesinc.github.io/libiio/codemodel.svg


If you have any comments, let me know.

PR #358, opened 2020-01-28
Licences: fix #350, and clean up license.

### LINKED ISSUE #350: License issue
This was pointed out in : 
https://bugzilla.redhat.com/show_bug.cgi?id=1426465

In all of the libiio *.c files (backend.c  buffer.c  channel.c  context.c  device.c  iiod-client.c  local.c  lock.c  network.c  scan.c  serial.c  sort.c  usb.c  utilities.c  xml.c), it's indicated:
```
 * This library is free software; you can redistribute it and/or
 * modify it under the terms of the GNU Lesser General Public
 * License as published by the Free Software Foundation; either
 * version 2.1 of the License, or (at your option) any later version.
```
In mainpage.dox it indicates:
```
Libiio has been developed and is released under the terms of the GNU Lesser General Public
License, version 2. 
```
which is missing the "any later version". Same for the README.md file.

Also, the utilities, which are in the `./test` directory, aren't a library, and LGPL doesn't really make sense for those applications. Same for the `./examples` directory. Section 2a of the LGPL states:
https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
```
a) The modified work must itself be a software library.
```

I would propose that:
-  I fix the README and Dox to match the license in the code (which I will do).
- the files in the ./test directory be updated to "GPLv2 or any later version"
- the files in the ./examples directory be updated to "GPLv2 or any later version"

I just need everyone's OK for 2 & 3 (if anyone doesn't answer, I won't do it).

Authors of c files in the ./test directory: @johnnykv @julienmalik @larsclausen @lucasrangit @mtitinger @mfornero @mhei @mhennerich @pcercuei @rgetz : I need your permission to do this.

Authors of c files in the ./examples directory @dNechita @dimasad @geertu @larsclausen @lucasrangit @mhennerich @pcercuei @SrikanthPagadarai @tfcollins : I need your permission to do this. 

Thanks

rgetz:
This ensures that applications are released under a license appropriate
for applications (GPL).
And doc (readme & Dox) ensures that everything is clear.


README.md
```diff
@@ -4,9 +4,10 @@ Library for interfacing with Linux IIO devices
...
 
-License : [![License](https://img.shields.io/badge/license-LGPL2-blue.svg)](https://github.com/analogdevicesinc/libiio/blob/master/COPYING.txt)
+Library License : [![Library License](https://img.shields.io/badge/license-LGPL2-blue.svg)](https://github.com/analogdevicesinc/libiio/blob/master/COPYING.txt)
+Application License : [![Application License](https://img.shields.io/badge/license-GPL2-blue.svg)](https://github.com/analogdevicesinc/libiio/blob/master/COPYING_GPL.txt)
```
pcercuei: I'd just change 'Application' with 'Tests/Examples' to avoid confusion, I don't want people to believe that their own application must be GPL.

rgetz: Changed/Updated

dNechita:
There are some examples at:
/bindings/<csharp | matlab | python>/
Should these files be included as well?
e.g.
https://github.com/analogdevicesinc/libiio/blob/master/bindings/python/examples/iio_info.py
https://github.com/analogdevicesinc/libiio/blob/master/bindings/csharp/examples/ExampleProgram.cs

rgetz:
If something is part of the binding, it's part of the libiio library, and should be covered under the LGPL.

If something is an example (stand alone application), it should be covered under the GPL. (LGPL is for libraries only).

You are correct - I didn't look in the binding directory. I can do that now or later...

README.md
pcercuei: Another thing: for the SVG you should now use ```https://img.shields.io/badge/license-LGPL2+-blue.svg```

rgetz: Ok - that’s easy. Thanks for the review.

rgetz: fixed

rgetz:
The most recent fix adds cleanups to c# and python. The MATLAB pieces are already properly licensed, so I will merge later if no one else has any objections.

pcercuei:
LGPL really is meant for software libraries, wouldn't it force the apps built around these bindings to be LGPL as well?

rgetz:
No. Using it for applications doesn't really make sense according to the license text.

[LGPL](https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html). 

```
A "library" means a collection of software functions and/or data prepared so as to be 
conveniently linked with application programs (which use some of those functions and 
data) to form executables. 

The "Library", below, refers to any such software library or work which has been 
distributed under these terms. A "work based on the Library" means either the 
Library or any derivative work under copyright law: that is to say, a work containing 
the Library or a portion of it, ...

<snip>

 2. You may modify your copy or copies of the Library or any portion of it, thus 
forming a work based on the Library, and copy and distribute such modifications or 
work under the terms of Section 1 above, provided that you also meet all of these
conditions:

    a) The modified work must itself be a software library.

```

So, we have the library (libiio), and works based on the library (ie - applications, or in our case iio-utils, or test apps, or examples, which are distributed under their own license - this pull request would make things GPL). For convenience, they are in the same repo, and can be turn on in the same build system, but their source is unique, and separate.

Whether someone builds their own application (which can be open or closed) the LGPL library (libiio) doesn't force people to use any specific license, as long as end users can replace the LGPL library, and still have that application re-link. [See reference](https://www.gnu.org/licenses/gpl-faq.en.html#LGPLStaticVsDynamic)

Does that make sense?

pcercuei:
Sorry, I wasn't clear - I meant for the bindings.

If the Python or C# bindings are LGPL, would it force the Python / C# applications using them to be LGPL too?

rgetz:
Paul:

Sorry - got it now.

Yeah, that's a more murky question. In discussing with someone from the FSF at FOSDEM, their opinion was if the bindings are part of the library, and can be replaced or re-linked as part of the library with the application code, then it's better to have them be considered as a part of the library, and be under LGPL (if someone wants to extend the bindings, they can do so, but they should provide the source to people who get the binary).

If the bindings are an intrinsic part of the application, (there is no way to re-link, or replace the .dll/.so) then it's better to release them under a BSD or something similar to the GCC RUNTIME LIBRARY EXCEPTION license (granting permission to propagate a work of Target Code formed by combining the Runtime Library with Independent Modules). 

And that's where the conversation stopped. It's a little complicated in action. Dynamically linking typically means any mechanism that loads the library into a process, including .so shared objects, dlopen(), and Python's import facilities. Copyright (the branch of law that the GPL/LGPL falls into) is technology-agnostic.

if "import iio" is dynamic linking, the bindings are fine under the LGPL. It was my interpretation that the python bindings could/should be updated independently from a python application. 

C# is the same. My understanding is that the C# bindings, create a different .dll that the C# applications links against. It's not included directly in the C# application. (but I could be wrong on this).

As always - I am not an attorney. More importantly, I am not your attorney. The above is not legal advice. If you desire legal advice, consult a competent, licensed attorney in your area.

pcercuei:
Note that both the Python bindings and the C# bindings also dynamically load libiio, so the bindings could very well be under a more permissive license (if there's desire for it - LGPL is fine for me).

rgetz:
I think it's fine - the bindings (IMHO) be under LGPL, and that way when people make improvements to the bindings - we can see them as well...

So, I think this is OK to merge

PR #361, opened 2020-02-03
Create README_BUILD.md

rgetz:
pre requests from some distribution maintainers, add a small readme for build instructions.

### DIFF
--- a/README_BUILD.md
+++ b/README_BUILD.md
@@ -1,0 +1,66 @@
+# Build instructions for libiio
+
+## Install Prerequisites/Dependancies
+
+Basic system setup
+```shell
+analog@precision:~$ sudo apt-get update
+analog@precision:~$ sudo apt-get install build-essential
+```
+Install Prerequisites
+```shell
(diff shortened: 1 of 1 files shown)


README_BUILD.md
```diff
@@ -0,0 +1,66 @@
...
+analog@precision:~$ sudo apt-get update
+analog@precision:~$ sudo apt-get install build-essential
+```
+Install Prerequisites
+```shell
+analog@precision:~$ sudo apt-get install libxml2-dev bison flex libcdk5-dev cmake
+```
+Install Backends
+```shell
+analog@precision:~$ sudo apt-get install libaio-dev libusb-1.0-0-dev libserialport-dev libxml2-dev libavahi-client-dev
```
dNechita: libxml2-dev is already in the install prerequisites list. Was it added intentionally to show that it is also a backend dependency?

rgetz: that was a bug/oversight. Thanks for checking, should be OK now.

Issue #190, opened 2018-09-07
iio_buffer_push_partial() is broken


aholtzma:
As of commit 19c1f6decdf28058bdbd0211b08ed0c9052661ed, calls to iio_buffer_push_partial() will always return -EBIG. This commit fixed one problem, but introduced a new problem later on here: 
```
	/* Requested buffer size is too big! */
	if (pdata->last_dequeued < 0 && bytes_used != block.size)
		return -EFBIG;
```
In the case of iio_buffer_push_partial(), bytes_used != block.size and will get erroneously caught here. 

It's not clear to me what this check is actually doing. Can someone comment on the rationale here?

pcercuei:
Sorry for the delay. The ```bytes_used != block.size``` expression is used to verify that the kernel does not return a buffer that is smaller than the number of bytes that are requested. If you call ```iio_buffer_push_partial``` right after creating the buffer with a small number of bytes, the call will fail, which is indeed a bug. I think it should be changed to ```bytes_used > block.size```.

mndza:
I am also using an old version of libiio due to this bug.
Tried the `bytes_used > block.size` fix and it seems to work OK.

rgetz:
@dNechita has an open pull request for this issue, so closing it here.

PR #362, opened 2020-02-04
local: Fix the checking of the buffer size returned by kernel

dNechita:
This PR solves #190 

Verify that the kernel does not return a buffer smaller than the number of bytes that are requested.

Issue: a call to iio_buffer_push_partial() right after creating the buffer with a small number of bytes, fails with -EFBIG.
This fixes that.

### DIFF
--- a/local.c
+++ b/local.c
@@ -509,7 +509,7 @@ static ssize_t local_get_buffer(const struct iio_device *dev,
 	}
 
 	/* Requested buffer size is too big! */
-	if (pdata->last_dequeued < 0 && bytes_used != block.size)
+	if (pdata->last_dequeued < 0 && bytes_used > block.size)
 		return -EFBIG;
 
 	pdata->last_dequeued = block.id;


commodo:
the fix looks good;
but, let's see what @pcercuei has to add;
this seems to have been introduced via d5fd005db06c9fc940b4de24b4474f4c1bf50dd4 `Local backend: Issue an error if the buffer size requested is too big`, but the information there is insufficient to draw a conclusion on why it was done this way initially

commodo:
optionally, the comment could have a `Fixes` tag, as done in the kernel.
or a reference to the commit that added these lines initally; i.e.  d5fd005d  `Local backend: Issue an error if the buffer size requested is too big`

dNechita:
> the fix looks good;
> but, let's see what @pcercuei has to add;
> this seems to have been introduced via [d5fd005](https://github.com/analogdevicesinc/libiio/commit/d5fd005db06c9fc940b4de24b4474f4c1bf50dd4) `Local backend: Issue an error if the buffer size requested is too big`, but the information there is insufficient to draw a conclusion on why it was done this way initially

I think Paul added some explanation in the issue that this PR addresses (see PR description).

dNechita:
> optionally, the comment could have a `Fixes` tag, as done in the kernel.
> or a reference to the commit that added these lines initally; i.e. [d5fd005](https://github.com/analogdevicesinc/libiio/commit/d5fd005db06c9fc940b4de24b4474f4c1bf50dd4) `Local backend: Issue an error if the buffer size requested is too big`

Done

commodo:
> > the fix looks good;
> > but, let's see what @pcercuei has to add;
> > this seems to have been introduced via [d5fd005](https://github.com/analogdevicesinc/libiio/commit/d5fd005db06c9fc940b4de24b4474f4c1bf50dd4) `Local backend: Issue an error if the buffer size requested is too big`, but the information there is insufficient to draw a conclusion on why it was done this way initially
> 
> I think Paul added some explanation in the issue that this PR addresses (see PR description).

Oh right.
The maybe we could include that discussion in the commit log as well.
Sorry for not looking into that link closer.

rgetz:
The bug link needs to go back in the commit message... Which thing it fixes doesn't matter to me - that is what git history/blame is for. (for something this easy).

PR #367, opened 2020-02-07
iio_readdev: Fix corrupt data that was being captured on Windows

dNechita:
The captured data is sent to stdout but during this process (on Windows)
any byte that has the same value as line feed character (LF) gets
converted to CF-LF characters. This dissrupts the alignment of bytes
that represent the samples of the device.

### DIFF
--- a/tests/iio_readdev.c
+++ b/tests/iio_readdev.c
@@ -84,6 +84,8 @@ static void quit_all(int sig)
 #ifdef _WIN32
 
 #include <windows.h>
+#include <io.h>
+#include <fcntl.h>
 
 BOOL WINAPI sig_handler_fn(DWORD dwCtrlType)
 {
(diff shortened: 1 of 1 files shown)


tests/iio_readdev.c
```diff
@@ -378,6 +380,12 @@ int main(int argc, char **argv)
 		return EXIT_FAILURE;
 	}
 
+#ifdef _WIN32
+	// Deactivate the translation for the stdout. Otherwise, bytes that have
+	// the same value as line feed character (LF) will be translated to CR-LF.
+	_setmode(_fileno(stdout), _O_BINARY );
```
commodo: nitpicks only:
1. `_setmode(_fileno(stdout), _O_BINARY );` ->   `_setmode(_fileno(stdout), _O_BINARY);`  ; remove space before last round bracket
2. generally, C-style comments are recommended; something like
```
	/*
	 * Deactivate the translation for the stdout. Otherwise, bytes that have
	 * the same value as line feed character (LF) will be translated to CR-LF.
	 */
```

dNechita: At the moment I was under the impression that I'm looking at C++ code.
Thanks!

PR #369, opened 2020-02-10
iio_info: use malloc rather than stack buffers

rgetz:
This applies the same fix as cc9b3c7 to iio_info, which fixes some
issues on AD936x based devices, which were giving:
	attr  9: gain_table_config ERROR: Input/output error (-5)

Now gives the correct data:
	attr  9: gain_table_config value: <gaintable AD9361 <snip>

This was discussed in #268

### DIFF
--- a/tests/iio_info.c
+++ b/tests/iio_info.c
@@ -1,7 +1,7 @@
 /*
  * iio_info - Part of Industrial I/O (IIO) utilities
  *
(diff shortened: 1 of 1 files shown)


rgetz:
I think this is OK to merge - the travis-ci failures are some weird CI MacOS failure.

PR #375, opened 2020-02-11
iio_attr : Add a feature which auto-generates C code

rgetz:
End users have said that they need an on-ramp for libiio,
so make it easy to generated little examples that show how to
read/write attributes. This supports both python and C.

This is a follow-on from #289, where I think I incorporated all the input.


tests/gen_code.c
```diff
@@ -0,0 +1,302 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * */
+
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <iio.h>
+//#include "gen_code.h"
```
dNechita: Please remove if not needed.

tests/gen_code.c
```diff
@@ -0,0 +1,302 @@
...
+{
+	if (!gen_file)
+		return;
+
+	if (lang == UNSUPPORTED_LANG)
+		return;
+
+	fd = fopen(gen_file, "w");
+	if (!fd) {
+		printf("Error '%s' opening file: %s\n", strerror(errno),gen_file);
```
dNechita: The message could also go to the 'stderr' stream.

tests/gen_code.c
rgetz: fixed

tests/gen_code.c
rgetz: fixed

tests/gen_code.c
```diff
@@ -0,0 +1,301 @@
...
+}
+
+void gen_context_timeout(unsigned int timeout_ms)
+{
+	if (!fd)
+		return;
+
+	if (lang == C_LANG) {
+		fprintf(fd, "\t/* Set the context timeout in ms */\n");
+		fprintf(fd, "\tiio_context_set_timeout(ctx, %i);\n", timeout_ms);
```
dNechita: static analysis complains about %i. It should be replaced with %u as timeout_ms is unsigned.

rgetz: sorry - should have looked at that - fixed now.

PR #378, opened 2020-02-17
Libiio streaming Python example.

cristi-iacob:
Added an example for streaming data using the libiio Python bindings.
This example works on AD9361. We generate two sine waves (one for I, one for Q), push them and read them afterwards.

### DIFF
--- a/bindings/python/examples/ad9361_iiostream.py
+++ b/bindings/python/examples/ad9361_iiostream.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+#
+# Copyright (C) 2020 Analog Devices, Inc.
+# Author: Cristian Iacob <cristian.iacob@analog.com>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+# AD9361 IIO streaming example - Python bindings
+
+import iio
+import numpy as np
+from sys import argv
+
+DEVICE_RX = "cf-ad9361-lpc"
+DEVICE_TX = "cf-ad9361-dds-core-lpc"
+DEVICE_PHY = "ad9361-phy"
+
+
+def main():
+	if len(argv) == 3 and argv[1] == '--uri':
+		uri = argv[2]
+	else:
+		contexts = iio.scan_contexts()
+		if len(contexts) > 1:
+			print('Multiple contexts found. Please select one using --uri:')
+			for index, each in enumerate(contexts):
+				print('\t%d: %s [%s]' % (index, contexts[each], each))
+			return
+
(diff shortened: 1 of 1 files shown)


bindings/python/examples/iio_stream.py
```diff
@@ -0,0 +1,76 @@
+"""
```
dNechita: I believe it is a good practice for the script to be able to detect if it is being run as script or imported as a module. Maybe others can offer more suggestions on this.
https://docs.python.org/3/library/__main__.html

bindings/python/examples/iio_stream.py
```diff
@@ -0,0 +1,76 @@
...
+AD9361 IIO streaming example - Python bindings
+"""
+import iio
+import time
+import numpy as np
+import matplotlib.pyplot as plt
+
+USB_CONTEXT = "ip:192.168.2.1"
```
dNechita: The contexts suggests an Ethernet context not an USB one.

dNechita: Would be nice to choose the context from command line.

dNechita:
ad9361_iiostream.py might be a more suitable name as it works only with ad9361.

rgetz:
For devices which are supported by https://github.com/analogdevicesinc/pyadi-iio I would rather just point people there... ?? 

An AD936x example is there, with many less lines, and much more pythonic.
https://github.com/analogdevicesinc/pyadi-iio/blob/master/examples/pluto.py

Even a simple watefall is less than 100 lines of code:
https://github.com/analogdevicesinc/pyadi-iio/blob/master/examples/waterfall.py

adisuciu:
@rgetz , like you mentioned, not all devices are supported by pyadi-iio.

The purpose of this example is to show how to acquire/push data to a buffered device in Python using the libiio bindings, not to show how to get data from the Pluto. 

There are currently no python examples to showcase this as far as I'm aware.

bindings/python/examples/iio_stream.py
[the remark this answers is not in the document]
cristi-iacob: Solved.

cristi-iacob: Solved.

bindings/python/examples/iio_stream.py
cristi-iacob: Solved.

cristi-iacob: Solved.

bindings/python/examples/ad9361_iiostream.py
```diff
@@ -0,0 +1,110 @@
...
+
+	while True:
+		rx_buf.refill()
+		samples = rx_buf.read()
+
+		dt = np.dtype(np.int16)
+		dt = dt.newbyteorder('<')
+		values = np.frombuffer(samples, dtype=dt)
+
+		for value in values:
```
tfcollins: Can you split the data into channels (I/Q) so users correctly interpret the data

rgetz:
> @rgetz , like you mentioned, not all devices are supported by pyadi-iio.
> 
> The purpose of this example is to show how to acquire/push data to a buffered device in Python using the libiio bindings, not to show how to get data from the Pluto.
> 
> There are currently no python examples to showcase this as far as I'm aware.

@adisuciu : yeah - agree - why not make the equivalent of read_dev and write_dev in python to be used as the example - similar to what iio_info.py does.

If _this_ example didn't use something that was already supported in peyote, I wouldn't have an issue. That's the point. Pick something else, that isn't supported. Don't encourage raw libiio for ad936x, or the other devices that _are_ supporting in peyote.

dNechita:
Do we keep this example or we drop it?

rgetz:
I think @tfcollins was going to have a chat with @adisuciu, I just never heard the conclusion.

tfcollins:
Closing for now.

PR #380, opened 2020-02-19
Add pypi installer support

tfcollins:


libiio.iss.cmakein
```diff
@@ -71,9 +68,3 @@ Source: "C:\projects\libiio\build-win32\bindings\csharp\libiio-sharp.dll"; DestD
 Source: "C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\redist\x86\Microsoft.VC120.CRT\msvcr120.dll"; DestDir: "{sys}"; Flags: onlyifdoesntexist 32bit
 Source: "C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\redist\x64\Microsoft.VC120.CRT\msvcr120.dll"; DestDir: "{sys}"; Check: Is64BitInstallMode; Flags: onlyifdoesntexist
 
-[Tasks]
```
commodo: my only question here is: do we lose any test/build coverage by removing these?
and if yes, any way to compensate? we could also add a CI run somewhere

tfcollins: We could actually install the built wheel... This would require installing the dll (then removing it for each variant)

commodo: guess it should work;
let's see later

PR #295, opened 2019-07-24
README.md: remove OS X El Capitan (10.11) from readme

commodo:
Given that suport for Xcode 11 is being added, maybe it's a good idea to
drop support for Xcode 9.

The readme file also contained a left-over for Xcode8 (OS X 10.10) which
was removed a while back. This was removed with this change.

Signed-off-by: Alexandru Ardelean <alexandru.ardelean@analog.com>

PR #381, opened 2020-02-19
ci,build: drop xcode9 build

commodo:
Mathworks no longer supports macOS 10.12
https://www.mathworks.com/support/requirements/matlab-system-requirements.html
```
Operating Systems
    macOS Catalina (10.15)
    macOS Mojave (10.14)
    macOS High Sierra (10.13.6)
Note:
    macOS Sierra (10.12) is no longer supported.
    On macOS High Sierra, version 10.13.6 is required.
```

Travis-CI's Xcode + macOS matrix lists Xcode 9.4 (only) running on macOS
10.13 and all Xcode10 running on 10.13.
All other XCode 9 versions only support 10.12
  https://docs.travis-ci.com/user/reference/osx/#macos-version

We can suggest that people upgrade their Xcode to version 10. It's also
likely that most people running macOS 10.13, are running Xcode 10 already.

In any case, we can start to consider dropping the Xcode9 build to reduce
the overall build-time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,11 +68,6 @@ matrix:
         - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
     - compiler: "gcc"
       os: osx
-      osx_image: xcode9.2
-      env:
-        - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
-    - compiler: "gcc"
-      os: osx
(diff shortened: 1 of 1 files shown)


commodo:
I forgot to add.
This is in continuation of https://github.com/analogdevicesinc/libiio/pull/295#issuecomment-514690136

Issue #263, opened 2019-03-07
Use strerror() wherever possible


commodo:
Some error reports are "I get code X", which prompts questions.
We could mitigate this by putting `strerror()` wherever possible, to also print strerror() strings.
Especially in examples

rgetz:
I think iio_attr and iio_info already do this.

Which other examples did you have in mind?

pcercuei:
There's ```iio_strerror``` btw.

rgetz:
Ok - a little check...

```
rgetz@brain:~/github/libiio$ grep iio_strerr ./tests/* -c
./tests/CMakeLists.txt:0
./tests/iio_adi_xflow_check.c:1
./tests/iio_attr.c:10
./tests/iio_genxml.c:0
./tests/iio_info.c:6
./tests/iio_readdev.c:3
./tests/iio_reg.c:0
./tests/iio_writedev.c:3
```

which isn't too bad - 

but

```
rgetz@brain:~/github/libiio$ grep iio_strerr ./examples/* -c
./examples/ad9361-iiostream.c:0
./examples/ad9371-iiostream.c:0
./examples/adrv9009-iiostream.c:0
./examples/dummy-iiostream.c:0
./examples/iio-monitor.c:0
./examples/Makefile:0
```

those need some work/care...

commodo:
As I remember, I created this issue because of some repeated reports on EZ would be "running example/code X, I get -22"; 

I forget which examples they were providing, but likely some of them were `ad9361-iiostream.c` [for Pluto and other boards with AD9361] and  `ad9371-iiostream.c` .
Naturally, some of our replies were: "-22 is Invalid Argument" or "-110 is Timedout ", etc;  so, we tell people to check things, but usually after we look for the `strerror()` value of that error code.

Not sure how much reduction in EZ questions [of this sort] we'd get by doing this; I just added it here, so I won't forget about it.

commodo:
i guess i'll close this for now;
it's a bit too generic;
and probably won't solve enough EZ issues anyway;

PR #386, opened 2020-02-21
warn when ignore return vals

rgetz:
I was starting to look at #263 (which I read as report more errors, better, so users can diagnose things themselves).

I thought I would start with the "more errors" first.

not 100% sure this is a good idea or not. It will ensure that people who use it - always check their return values. (that's when/if it gets used), and hopefully experience less issues using the library. I have not tried compiling osc with this yet... (and am a little afraid to).

Would like to get @dNechita & @pcercuei & @commodo & @tfcollins opinion.

### DIFF
--- a/iio.h
+++ b/iio.h
@@ -59,10 +59,16 @@ typedef ptrdiff_t ssize_t;
 #define __pure __attribute__((pure))
 #endif
 #define __notused __attribute__((unused))
+#ifdef IIO_CHECK_RET
+#define __check_ret __attribute__((warn_unused_result))
+#else
+#define __check_ret
+#endif
 #else
 #define __cnst
 #define __pure
 #define __notused
+#define __check_ret
 #endif
 
 #ifdef _WIN32
@@ -193,7 +199,7 @@ enum iio_modifier {
  * @param flags Unused for now. Set to 0.
  * @return on success, a pointer to a iio_scan_context structure
  * @return On failure, NULL is returned and errno is set appropriately */
-__api struct iio_scan_context * iio_create_scan_context(
+__api __check_ret struct iio_scan_context * iio_create_scan_context(
 		const char *backend, unsigned int flags);
 
 
@@ -211,7 +217,7 @@ __api void iio_scan_context_destroy(struct iio_scan_context *ctx);
  * @returns On success, the number of contexts found.
  * @returns On failure, a negative error number.
  */
-__api ssize_t iio_scan_context_get_info_list(struct iio_scan_context *ctx,
+__api __check_ret ssize_t iio_scan_context_get_info_list(struct iio_scan_context *ctx,
 		struct iio_context_info ***info);
 
 
@@ -225,7 +231,7 @@ __api void iio_context_info_list_free(struct iio_context_info **info);
  * @param info A pointer to an iio_context_info structure
  * @return A pointer to a static NULL-terminated string
  */
-__api __pure const char * iio_context_info_get_description(
+__api __check_ret __pure const char * iio_context_info_get_description(
 		const struct iio_context_info *info);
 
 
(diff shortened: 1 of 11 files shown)


commodo:
I like the change.
Naturally there would be a slightly bumpy period where some users are bothered by the [potential] noise introduced by this change; and the fact that the need to update old code.
But, I feel that long-term this benefits us a bit more.

commodo:
Regarding the Codacy complaint; we can ignore it.
The scope-reduction thingi, we could probably remove from the checks.

iio.h
```diff
@@ -59,10 +59,16 @@ typedef ptrdiff_t ssize_t;
 #define __pure __attribute__((pure))
 #endif
 #define __notused __attribute__((unused))
+# ifdef IIO_CHECK_RET
+#  define __check_ret __attribute__((warn_unused_result))
+# else
+#  define __check_ret
```
commodo: after a bit of chatting with @dNechita we noticed the `# __2spaces__ define`
it's a good idea to avoid spaces there; since it can be a bit undefined behavior;
the precompilers typycally look for `#define`  keywords, and some precompilers are smart enough to ignore spaces, but some aren't

dNechita:
IIO Scope will complain about 200 times about this.

I think these changes look good. Of course the "IIO_CHECK_RET" needs to be documented somewhere so that is visible to users.

iio.h
rgetz: Ok - changed.

It's just easier to read this way, and some are doing the same in the same file.

rgetz:
does that mean it should be turned on by default, and the option should be to turn it off?

I would say we do that in a future version.
1. let people turn it on optionally now, then 
2. let people turn if off optional in a future release
3. eventually take the option away, so checking is warned all the time.

rgetz:
Also - this patch only effects gcc builds,

The [MSVC](https://docs.microsoft.com/en-us/cpp/code-quality/annotating-function-behavior?view=vs-2019) equivalent is  `_Must_inspect_result_` or `_Check_return_` however:
- SAL annotations like `_Check_return_` and `_Must_inspect_result_` are only checked during code analysis builds (either by starting a code analysis build in the IDE or by building with the /analyze flag on the command line), not regular builds like gcc does.
- unlike gcc et al, (a) MSVC requires annotations on both declaration and definition of a function, and (b) the annotation needs to be at the start of the declaration/definition (gcc allows either).

So, if we want to support MSVC with the same - the patch just got alot bigger...

commodo:
> So, if we want to support MSVC with the same - the patch just got alot bigger...

a lot of compilers do the same thing;
we can choose one compiler [for the _Check_return_ behavior] and hope it's good enough;
and we can extend stuff later;
maybe with some macros we could have things neatly in all compilers, but that sounds like just too much work; i'm fine with just having the GCC complaints

rgetz:
OK - let me turn this on for the tests by default , and add the doc to the man page.

rgetz:
OK - this is ready for a final review and merge if no one else has any issues.

commodo:
all good from my side

dNechita:
Can someone configure Codacy to ignore the scope reduction warning? Otherwise we will start ignoring it, allowing other issues to pile up.
Also, a just as good alternative would be to declare the 'ret' variable in the block of code where it is used.

man/make_man.sh
```diff
@@ -46,6 +46,22 @@ subsystem and devices
...
 .B -liio
+.SH OPTIONS
+The define
+.B
+IIO_CHECK_REG
+will warn if return values are not checked. Most
+.B
+libiio
+functions, if/when a failure occurs will return a negative error number.
+this warning will ensure these error numbers are looked it. There is
```
dNechita: shouldn't it be: "are looked at" ?

commodo:
removed scope check; 
closed + opened PR: hopefully this re-triggers the check

rgetz:
I just fixed it.

man/make_man.sh
rgetz: fat fingers and lazy eyes, fixed now. Thanks for checking.

Issue #307, opened 2019-09-02
libiio only supports Avahi, and doesn't natively support Bonjour on Apple and Windows.


matejk:
As far as I was able to understand, IIOD can broadcast its availability using Avahi.

Client library libiio can discover such devices on Linux using Avahi library.

Since Avahi and Bonjour are compatible, is there a way to use discovery feature in libiio on macOS somehow or shall this part be implemented separately using macOS native APIs?

tfcollins:
Auto-discovery works fine out of the box:

```
analogs-mini:~ analog$ uname -a
Darwin analogs-mini.lan 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64
analogs-mini:~ analog$ iio_attr -u ip:analog -d
IIO context has 6 devices:
	iio:device0: ams, found 1 device attributes
	iio:device1: ad7291, found 0 device attributes
	iio:device2: ad9523-1, found 10 device attributes
	iio:device3: axi-ad9680-hpc, found 0 device attributes
	iio:device4: axi-ad9144-hpc, found 0 device attributes
	iio_sysfs_trigger:, found 2 device attributes
analogs-mini:~ analog$ 
```

matejk:
@tfcollins: I see that I wrote misleading text. The devices that I was referring to were IIOD servers on the network, discoverable with Zeroconf.

![macos-discovery](https://user-images.githubusercontent.com/699334/64936143-d1c96500-d854-11e9-8e45-f01118ff8936.png)

libiio has client API to find such servers and I was wondering if it is possible to use that on macOS somehow or shall I use native functions.

tfcollins:
This is basically what I showed with my commands. IIOD with autodiscovery is set up on a development board with hostname analog on the network. iio_attr is using the avahi/bonjour to find it.

pcercuei:
@tfcollins: you used ```iio_attr -u ip:analog -d```. I believe 'analog' is the local name of the board?

Bonjour discovery works without any name. ```iio_attr -u ip: -d``` should auto-connect to an instance of IIOD running on the local network if Avahi is enabled.

rgetz:
```analog``` is the name of the local host of the remote context iiod is running on.

[...]

PR #388, opened 2020-02-25
DNS SD implementation for macOS (Bonjour)

matejk:
Related to #307.


rgetz:
This looks great to me - Just need to test things. Thanks for submitting.

rgetz:
when running the small shell script
```while [ 1 ] ; do  x=0 ; i=0 ; while [ "${x}" -eq "0" ] ; do ./tests/iio_attr -d -u ip: >/dev/null ; x=$? ; i=$((i+1)) ; if [ "${x}" -eq "1" ] ; then echo "$i : ret= $x" ; fi; done; done```

I get:
```
Unable to create IIO context: Connection refused
120 : ret= 1
Unable to create IIO context: Connection refused
87 : ret= 1
Unable to create IIO context: Connection refused
1 : ret= 1
Unable to create IIO context: Invalid argument
1 : ret= 1
```

but it looks like that isn't new - so this works as well as it did before :)

If I look at the device it's attaching to, it's ZedBoard on my network, where I'm running iiod.
```
root@analog:~# ip  addr show dev eth0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 00:0a:35:00:01:22 brd ff:ff:ff:ff:ff:ff
    inet 192.168.1.120/24 brd 192.168.1.255 scope global eth0
       valid_lft forever preferred_lft forever
```

If I print out the address/port that it finds, it normally files the ZedBoard.

Sometimes it finds an IPv6 number.and I think we are passing that to the IPv4 framework..

```
found : fe80::689c:66a7:fbaa:ce88:30431
Unable to create IIO context: Invalid argument
94 : ret= 1
```

The weird part is that is my host PC with a Pluto attached.
```
rgetz@brain:~/github/temp/libiio/build$ ip -6 addr
4: enx00e0229bee20: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UNKNOWN qlen 1000
    inet6 fe80::689c:66a7:fbaa:ce88/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
```

and

```
found : 192.168.2.10:30431
Unable to create IIO context: Connection refused
32 : ret= 1
```

`192.168.2.10` is the PC that I'm running on (again - the local IP of the PlutoSDR connected USB)
```
rgetz@brain:~/github/temp/libiio/build$ ip addr show dev enx00e0229bee20
4: enx00e0229bee20: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UNKNOWN group default qlen 1000
    link/ether 00:e0:22:9b:ee:20 brd ff:ff:ff:ff:ff:ff
    inet 192.168.2.10/24 brd 192.168.2.255 scope global dynamic noprefixroute enx00e0229bee20
       valid_lft 813811sec preferred_lft 813811sec
    inet6 fe80::689c:66a7:fbaa:ce88/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
```
so that is weird/wrong. It should respond on 192.168.2.1, but not 10.

Again - I think these both exist on master as well, I just don't know if we wanted to fix them as part of this or not...

matejk:
@rgetz : Did you run the script on Linux (Avahi) or macOS (Bonjour)?

Implementation for macOS prefers ipv4 address over ipv6 if the OS gives both.

rgetz:
This was on Linux.I don't know if @tfcollins can test on macOS with a similar setup (something on LAN that would respond as well as Pluto plugged in).

matejk:
Implementation in this PR implements same behaviour as it is there for Avahi already.

Current implementation of DNS SD (either Avahi or Bonjour) in libiio is intended only for network environments with only one IIOD. To support multiple IIOD servers it would have to be implemented differently. But that it is out of scope of this PR.

rgetz:
Agreed. I can replicate the behavior described on old releases as well.

Sometimes - it would respond with the IP address of the machine it was running on, so there is something weird going on. I will open a separate issue, and wait for @tfcollins to confirm operation on macOS.

That only leaves Win32 as something that isn't supported... Now that we have a Bonjour implementation (thanks again) - it should be straightforward to do something eventually. (not in this PR, as I think it would be much more complicated - you would want to dlopen() the library yourself, since it's not always on Windows, and I don't think you want to force people to install it - its part of iTunes, or Apple Print Services https://support.apple.com/kb/DL999?viewlocale=en_US&locale=en_US ).

matejk:
Perhaps it is better to use MS implementation for DNS SD as described here:

https://docs.microsoft.com/en-us/uwp/api/windows.networking.servicediscovery.dnssd

I am not sure for which Windows versions is this available, but it is on Windows 10 almost definitely.

rgetz:
Thanks for pointer - I will have a look at those.

PR #392, opened 2020-03-03
Update the channel type and modifier lists against Linux 5.4.22

ynezz:
Update the channel type and modifier lists against Linux 5.4.22.

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -56,6 +56,9 @@ static const char * const iio_chan_type_name_spec[] = {
 	[IIO_COUNT] = "count",
 	[IIO_INDEX] = "index",
 	[IIO_GRAVITY] = "gravity",
+	[IIO_POSITIONRELATIVE] = "positionrelative",
+	[IIO_PHASE] = "phase",
+	[IIO_MASSCONCENTRATION] = "massconcentration",
 };
 
 static const char * const modifier_names[] = {
@@ -79,6 +82,7 @@ static const char * const modifier_names[] = {
 	[IIO_MOD_LIGHT_GREEN] = "green",
 	[IIO_MOD_LIGHT_BLUE] = "blue",
 	[IIO_MOD_LIGHT_UV] = "uv",
+	[IIO_MOD_LIGHT_DUV] = "duv",
 	[IIO_MOD_QUATERNION] = "quaternion",
 	[IIO_MOD_TEMP_AMBIENT] = "ambient",
 	[IIO_MOD_TEMP_OBJECT] = "object",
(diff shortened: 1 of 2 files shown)


rgetz:
Thanks - going forward (new releases), we should always make sure we are up to date with this:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/iio/types.h
(which hasn't changed in a while)

Just to make checking easier - I wrote a shell script that compares the kernel against libiio's enums, and it does print out this today.
```
looking for iio_chan_type
Differences in iio_chan_type
--- /tmp/libiio_iio_chan_type	2020-03-04 10:15:47.294648694 -0500
+++ /tmp/kernel_iio_chan_type	2020-03-04 10:15:47.290648671 -0500
@@ -30,3 +30,6 @@
 	IIO_COUNT,
 	IIO_INDEX,
 	IIO_GRAVITY,
+	IIO_POSITIONRELATIVE,
+	IIO_PHASE,
+	IIO_MASSCONCENTRATION,
looking for iio_modifier
Differences in iio_modifier
--- /tmp/libiio_iio_modifier	2020-03-04 10:15:47.298648717 -0500
+++ /tmp/kernel_iio_modifier	2020-03-04 10:15:47.298648717 -0500
@@ -35,3 +35,10 @@
 	IIO_MOD_CO2,
 	IIO_MOD_VOC,
 	IIO_MOD_LIGHT_UV,
+	IIO_MOD_LIGHT_DUV,
+	IIO_MOD_PM1,
+	IIO_MOD_PM2P5,
+	IIO_MOD_PM4,
+	IIO_MOD_PM10,
+	IIO_MOD_ETHANOL,
+	IIO_MOD_H2,
```

Which is fixed by this patch. 

@dNechita Did you want me to add this to the CI? or just put it in somewhere to run manually before a release?

dNechita:
I would add it to CI but run it only when a release candidate tag gets created. I don't think these types will change frequently.

rgetz:
> when a release candidate tag gets created.

Isn't it too late then? once the tag is created - it's pushed to github releases.

To run the script - it's less that half a second.
```
real	0m0.495s
user	0m0.359s
sys	0m0.086s
```

(and it does point out - `IIO_NO_MOD` is missing from channel.c modifier_names)

dNechita:
> > when a release candidate tag gets created.
> 
> Isn't it too late then? once the tag is created - it's pushed to github releases.
Yes, it is. I guess it is ok to send the script to me and I will run it before each release.
> 
> To run the script - it's less that half a second.
> 
> ```
> real	0m0.495s
> user	0m0.359s
> sys	0m0.086s
> ```

ynezz:
> I wrote a shell script that compares the kernel against libiio's enums, and it does print out this today.

Nice idea!

> and I will run it before each release

IMHO you're just one step away from automation, that script is probably the hardest part.

> Isn't it too late then? once the tag is created - it's pushed to github releases.

This could be probably solved in a few different ways, those came to mind right now:

* ci/pre-release branch, where one could add any pre-release checks before tagging
* weekly/monthly [scheduled CI job](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule) which could for example (sorted by complexity):
   * simply just fail and inform the project maintainers via standard CI failure channels 
   * create a new issue informing about the header desync
   * prepare the PR with the patch directly

rgetz:
We already do a nightly cron on travis-ci; let me see about scheduling more than one.

Looks like you can - https://docs.travis-ci.com/user/cron-jobs/

https://github.com/travis-ci/travis-ci/issues/7199 - I don't think you can create an issue automatically (easily). so some other annoying things will need to work. https://docs.travis-ci.com/user/notifications/

rgetz:
I will merge this (thanks again), and start a new issue tracking the script.

PR #395, opened 2020-03-06
improve dns_sd to handle multiple hosts & enable scanning on network

rgetz:
Sorry - this should be marked as a draft - don't merge this.

First take on fixing #395 

Just trying to get some feedback from @commodo & @dNechita @tfcollins  @matejk 

### DIFF
--- a/dns_sd_avahi.c
+++ b/dns_sd_avahi.c
@@ -3,6 +3,7 @@
  *
  * Copyright (C) 2014-2020 Analog Devices, Inc.
  * Author: Paul Cercueil <paul.cercueil@analog.com>
+ *         Robin Getz <robin.getz@analog.com>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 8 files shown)


scan.c
```diff
@@ -146,6 +160,10 @@ struct iio_scan_context * iio_create_scan_context(
 	if (!backend || !strcmp(backend, "usb"))
 		ctx->usb_ctx = usb_context_scan_init();
 #endif
+#ifdef HAVE_DNS_SD
+	if (!backend || strcmp(backend, "ip"))
```
commodo: curios: was this supposed to be `!strcmp(backend, "ip"))` ?

network.c
```diff
@@ -1356,8 +1356,8 @@ struct iio_context * network_create_context(const char *host)
 			errno = -ret;
 			return NULL;
 		}
-
 		iio_snprintf(port_str, sizeof(port_str), "%hu", port);
+		fprintf(stderr, "Looking for %s:%s\n", addr_str,port_str);
```
commodo: this looks like a debug print();
maybe should be wrapped in a `DEBUG()` or similar macro?

dns_sd_avahi.c
```diff
@@ -14,42 +15,191 @@
...
+#include <avahi-common/malloc.h>
 #include <avahi-client/client.h>
 #include <avahi-client/lookup.h>
 
 #include "debug.h"
 
+/* in network.c */
+int create_socket(const struct addrinfo *addrinfo, unsigned int timeout);
+#define DEFAULT_TIMEOUT_MS 5000
+#define IIOD_PORT 30431
```
commodo: maybe it's worth adding a `network.h` and moving things there; especially if exporting `create_socket()`

dns_sd_avahi.c
```diff
@@ -14,42 +15,191 @@
...
+
+static struct avahi_discovery_data * new_discovery_data(void)
+{
+	struct avahi_discovery_data *data;
+
+	data = zalloc(sizeof(struct avahi_discovery_data));
+	if (!data)
+		return NULL;
+	data->address = zalloc(sizeof(struct AvahiAddress));
+	if (!data->address)
```
commodo: this return path should probably `free(data)`;

dns_sd_avahi.c
```diff
@@ -61,82 +211,279 @@ static void __avahi_browser_cb(AvahiServiceBrowser *browser,
...
+
+	ddata->poll = avahi_simple_poll_new();
+	if (!ddata->poll) {
+		*ret = -ENOMEM;
+		return NULL;
+	}
 
-	client = avahi_client_new(avahi_simple_poll_get(poll),
-			0, NULL, NULL, &ret);
+	client = avahi_client_new(avahi_simple_poll_get(ddata->poll),
```
commodo: this change with `poll`  -> `ddata->poll`  makes things a bit hard to follow; and looks un-needed;
maybe some of this stuff could be moved into `new_discovery_data()` before  adding new logic?

dns_sd_avahi.c
```diff
@@ -61,82 +211,279 @@ static void __avahi_browser_cb(AvahiServiceBrowser *browser,
...
+	}
+
+	iio_context_destroy(ctx);
+
+	info->uri = iio_strdup(uri);
+	if (!info->uri)
+		return -ENOMEM;
+	info->description = iio_strdup(description);
+	if (!info->description)
+		return -ENOMEM;
```
commodo: `free(info->uri)` here?

dns_sd_avahi.c
```diff
@@ -14,42 +15,191 @@
...
+	uint16_t port;
+	struct avahi_discovery_data *next;
 };
 
+/*
+ * Fundamentally, this builds up a linked list to manage
+ * potential clients on the network
+ */
+
+static struct avahi_discovery_data * new_discovery_data(void)
```
commodo: if this extends a linked-list, could we make this also add the new elements on the list?
and move some common stuff here?
maybe that could be among the first patches

network.c
rgetz: yeah, I left this once, so I would remember to completely remove it before submission.

dns_sd_avahi.c
rgetz: yeah, wasn't sure - makes the most sense - I assume `network-private.h`?

dns_sd_avahi.c
rgetz: Thanks - good catch

dns_sd_avahi.c
rgetz: The issue is - the poll needs to be kept track of, and passed to __avahi_browser_cb, so that it's in the linked list...

What was happening before - was things were allocated, and the poll and ddata->poll were pointing to the same object, this just removes that (what I considered) obfuscation.

Your point of breaking this up into multiple commits is spot on through...

dns_sd_avahi.c
rgetz: yes - same problem exists in usb.c

commodo: at some point we may have to look at running valgrind or some other tool to catch leaks;
or maybe writing some mcheck code
http://man7.org/linux/man-pages/man3/mcheck.3.html

dns_sd_avahi.c
rgetz: The current logic is different.

There is always a "blank" node ready for data to be populated in. 
See below, new_discovery_data is called/gets added to ->next when the current is populated. (as a blank one for next time). That is also how we find the end of the list in all the iterators.

dns_sd_avahi.c
rgetz: we do run coverity on this - https://scan.coverity.com/projects/analogdevicesinc-libiio

it should detect these sorts of things as well...

dns_sd_avahi.c
commodo: could we reverse the logic a bit and populate the node right after it gets created?

dns_sd_avahi.c
commodo: i think here the issue may be that we're calling `iio_strdup()`;
no idea how coverity works, but cppcheck [for example] needs to know that the return value of `iio_strdup()` allocates memory/resource;

that way, it can guess/infer which error path needs free-ing;

dns_sd_avahi.c
rgetz: Not really - it needs a blank node to store the pointer returned from `avahi_simple_poll_new();` in the `dnssd_find_hosts()` function - so it can be passed as part of the data structure (the linked list) to the browse and resolve function.

commodo: oh, ok
makes sense

rgetz:
This is closer, I will leave things util #398 is fixed.

dns_sd_avahi.c
```diff
@@ -125,18 +362,47 @@ int discover_host(char *addr_str, size_t addr_len, uint16_t *port)
...
+	return ret;
+}
+
+int discover_host(char *addr_str, size_t addr_len, uint16_t *port)
+{
+	struct dns_sd_discovery_data *ddata;
+	int ret = 0;
+
+	ret = dnssd_find_hosts(&ddata);
+
```
matejk: I propose that this function fails if discovery finds more than one IIOD server. Behaviour is not deterministic otherwise, because it is no guaranteed that the first element on the list is always the same.

rgetz: correct - the proposed solution is that the first that the DNS SD passes back is the one that will get passed to the user.

This is what happens today as well - it's just we don't decode things. In today's code - the last one which goes into the single buffer is the one that is passed back...

matejk: Yes, this change would break backward compatibility. Is it possible to somehow deprecate the functionality and remove it in one or two releases?

rgetz: It's sort of up to the application that uses libiio if they want to pass blank ip: context names to the library or not. As long as people understand what the library does, it should be OK.(maybe?)

What I will do in things like iio_attr is use the rest of the arguments that are passed to the application as "hints" to pass back the right context, and only fail if there are multiple contexts that match. Same as what the hints argument does for `getaddrinfo()`.

Things like:
```
rgetz@brain:~/github/libiio$ iio_attr -a -C hw_model
Using auto-detected IIO context at URI "usb:3.15.5"
hw_model: Analog Devices PlutoSDR Rev.B (Z7010-AD9364)
```

will only work if no other context has a context attribute of hw_model. 
Right now, it does:
```
rgetz@brain:~/github/libiio-dns/build$ ./tests/iio_attr -a -C hw_model
Multiple contexts found. Please select one using --uri:
	0: 0456:b673 (Analog Devices Inc. PlutoSDR (ADALM-PLUTO)), serial=104473222a87000618000600473ed57ae0 [usb:3.15.5]
	1: 192.168.1.120 (ad7291,ad9361-phy,xadc,cf-ad9361-dds-core-lpc,cf-ad9361-lpc) [ip:analog.local]
```

which is silly, since only one context has a context attribute with that name.

rgetz:
I'm going to close this one, and re-open a new one tomorrow - thanks for feedback.

PR #399, opened 2020-03-12
Cmake: restore installing the Avahi service file

rgetz:
During commit 804216ac90db998ad01b1774e93cb49497b90187 we accidently
dropped the CMAKE setting HAVE_AVAHI, however, this flag controls
the setting to install the avahi service file, so it's no longer
installed.

This fixes it by adding it back.


matejk:
I think that Avahi service file would be only necessary if `avahi-daemon` publishes iio service from static XML file.

Since iiod does publishing itself using functions from Avahi library, the Avahi service file is not needed. Both Avahi and iiod want to publish iio. Perhaps that is the reason why some "ghost" iio services appear on the network.

Publishing iio service is being fixed in #402.

PR #400, opened 2020-03-12
ci,build: split 'make && make package' commands

### LINKED ISSUE #398: all travis-ci builds on Linux give false positives
If you look at a recent build that fails, like:

https://travis-ci.org/github/analogdevicesinc/libiio/builds/661315711?utm_source=github_status&utm_medium=notification

The Linux builds are all passing - which is wrong. Looking at the logs:

/docker_build_dir/dns_sd_avahi.c:384: warning: implicit declaration of function 'avahi_simple_poll_free'
/docker_build_dir/dns_sd_avahi.c:384: error: 'struct dns_sd_discovery_data' has no member named 'poll'
make[2]: *** [CMakeFiles/iio.dir/dns_sd_avahi.c.o] Error 1
make[1]: *** [CMakeFiles/iio.dir/all] Error 2
make: *** [all] Error 2

But the script continues and provides a 0 error code on exit.

(I think) since the exit code is dropped from the make, since the script library runs and rate checks things (which functions).

commodo:
Fixes #398 

For some reason, doing `make && make package` doesn't seem to yield a
compiler/build error on Linux.
This may be due to how '/bin/sh' interprets command concatenation via '&&'.
Splitting the make commands seems to help.

See:
  https://travis-ci.org/github/analogdevicesinc/libiio/builds/661401114

This is essentially branch 'rgetz-review-zeroconfupdates' rebased with this
change.

### DIFF
--- a/CI/travis/before_install_linux
+++ b/CI/travis/before_install_linux
@@ -19,7 +19,9 @@ handle_ubuntu_docker() {
 
 handle_default() {
 	sudo apt-get -qq update
-	sudo apt-get install -y cmake graphviz libaio-dev libavahi-client-dev libavahi-common-dev libusb-1.0-0-dev libxml2-dev rpm tar bzip2 gzip flex bison git
+	sudo apt-get install -y cmake graphviz libaio-dev libavahi-client-dev \
+		libavahi-common-dev libusb-1.0-0-dev libxml2-dev rpm tar \
+		bzip2 gzip flex bison git python-pip
 	if [ -n "${GH_DOC_TOKEN}" ] ; then
 		sudo apt-get install -y doxygen
 	fi
--- a/CI/travis/make_darwin
+++ b/CI/travis/make_darwin
@@ -9,5 +9,6 @@ ls
 
 cd $TRAVIS_BUILD_DIR/build_tar
 cmake -DOSX_PACKAGE=OFF -DENABLE_PACKAGING=ON -DPYTHON_BINDINGS=ON ..
-make && make package
+make
+make package
 ls
--- a/CI/travis/make_linux
+++ b/CI/travis/make_linux
@@ -24,7 +24,8 @@ handle_default() {
 	cmake -DENABLE_PACKAGING=ON -DDEB_DETECT_DEPENDENCIES=ON \
 		-DPYTHON_BINDINGS=ON ${DOC_FOUND} ..
 
-	make && make package
+	make
+	make package
 	if [ -n "${GH_DOC_TOKEN}" ] && \
 			[ -f "./generateDocumentationAndDeploy.sh" ] ; then
 		sh generateDocumentationAndDeploy.sh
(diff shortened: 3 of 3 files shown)


commodo:
seems that pip not being installed on Ubuntu 18.04 was masked for a while by this issue

rgetz:
yeah, I don't know how long this has been happening
Thanks for the quick fix.

I will test this on the build that is failing...

rgetz:
should note - this does fix things on the other pull request, and does make things fail, and be noticed.

commodo:
> Looks good to me.
> 
> The changes to ./before_install.sh don't seem to have anything to do with this though (and should be in a separate commit - since I don't care about that (since I do the same), I will just approve things. :)

hmm; those changes should be in a separate commit (i.e 4da169fbb99f38abd1314a9c0944034747941ef0);
albeit, i will admit that i am re-arranging things on multiple lines while adding python-pip;

they should fix one of the ubuntu builds; particularly this  https://travis-ci.org/github/analogdevicesinc/libiio/jobs/661405250#L1969

while on the subject [or close to it], for Linux builds [right now] we typically don't merge any PR if the build isn't green;
that does mean that we need to ignore some builds [if they fail]; travis-ci has some facility to mark these flaky/not-very-important-builds [like checkpatch for Linux] 

any thoughts on making this mandatory for libiio [and maybe everywhere eventually]?
[ it's just one checkbox we need to tick here https://github.com/analogdevicesinc/libiio/settings/branch_protection_rules/470471 ]
one of the main ideas of CI is to not allow failing builds to be merged, which also means adding more checks to catch them;

rgetz:
I'm a little hesitant about being so process driven, as when the build system has issues - it makes it harder to fix. I think as the norm - we shouldn't accept things that aren't building. but this for example - when you split it in two (which is great), the build error for the missing pip still causes a build error (which we want to accept and fix)...

I'm OK either way - its easy to uncheck it, as it as easy to check it. :) 

Might be better, as it will force people to see what is really failing, and pay more attention to the CI...

commodo:
> I'm a little hesitant about being so process driven, as when the build system has issues - it makes it harder to fix. I think as the norm - we shouldn't accept things that aren't building. but this for example - when you split it in two (which is great), the build error for the missing pip still causes a build error (which we want to accept and fix)...
> 
> I'm OK either way - its easy to uncheck it, as it as easy to check it. :)
> 
> Might be better, as it will force people to see what is really failing, and pay more attention to the CI...

I totally get being hesitant about this.
The CI system has some faults, and sometimes infrastructure issues get in the way.
The fault-rate isn't too bad, but it's not near-perfect either.
I think for Linux [when I enabled this about 1-2 months ago] I only had to untick the checkbox once [temporarily] because the link between Travis-CI & Gihub had an issue: the build was green in Travis-CI, but was still showing-up as pending in Github.

As with being process-driven/or-not: that always depends on how quality-driven we want to be. I feel that CI-tools/checks are always best to enforce things [for quality]. That way, during reviews, there are fewer shades of grey of what is acceptable/unacceptable to merge as a team grows.
Evey-now-n-then [when teams grow], reviews can become difficult because more opinions can conflict; but I've always felt [and also noticed] that there are fewer conflicting opinions with CI.
Which is why I've developed an affinity towards developing CI.

Issue #403, opened 2020-03-14
CMake build with -DWITH_LOCAL_CONFIG=ON fails


matejk:
I wanted to include support for reading local ini file.

```
$ cmake ../libiio -DWITH_LOCAL_CONFIG=ON
CMake Error at CMakeLists.txt:170 (message):
  WITH_LOCAL_CONFIG option requires libini to be installed
```

Build instruction to not contain information about the libini library.

How is this supposed to be built?

rgetz:
with libini

I will add it to the build instructions.

rgetz:
I just checked the pluto and m2k firmware - you need to check out https://github.com/pcercuei/libini/ and build/install it.

rgetz:
I updated the README_BUILD.md and https://wiki.analog.com/resources/tools-software/linux-software/libiio

rgetz:
I looked at adding this to the Linux images on the SD Cards:

```
diff --git a/adi_update_tools.sh b/adi_update_tools.sh
index 49b36d4..4e4da9c 100755
--- a/adi_update_tools.sh
+++ b/adi_update_tools.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 if [ "$(id -u)" != "0" ] ; then
        echo "This script must be run as root"
@@ -261,6 +262,24 @@ do
        # Non-essential applications, which help out sometime
        apt-get -y install gpsd gpsd-clients u-boot-tools evtest
 
+       # install libini
+       pushd $(pwd)
+       cd /usr/local/src
+        if [ -d "libini" ] ; then
+          cd ./libini
+          git pull
+        else
+         git clone https://github.com/pcercuei/libini.git
+          cd ./libini
+        fi
+       mkdir -p build
+       cd build
+       rm * -rf
+       cmake ../
+       make
+       make install
+
        if [ ! -f /etc/fw_env.config ]
        then
          cp ./fw_env.config /etc/
```

but this fails, since Paul has a min cmake of 3.8, and in the 2019 release, it is still cmake 2.12

libini needs this change:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 0a7dee6..c3773cd 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,9 @@
-cmake_minimum_required(VERSION 3.1.0)
-project(libini LANGUAGES C VERSION 0.1)
[...]

PR #405, opened 2020-03-16
libini added as a submodule and compiled/linked directly into libiio.

matejk:

Done in a similar way as it is done for wingetopt.

Related to #403.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "deps/wingetopt"]
 	path = deps/wingetopt
 	url = https://github.com/alex85k/wingetopt.git
+[submodule "deps/libini"]
+	path = deps/libini
+	url = https://github.com/pcercuei/libini.git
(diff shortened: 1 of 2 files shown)


mhennerich:
Let's wait for @pcercuei comments here...

The options are:
1) Add copy of libini to libiio (my preference)
2) Fork libini fix the cmake issue
3) Fix libini upstream
3) This patch 

-Michael

matejk:
This PR was not intended to fix libini CMake build but to simplify build of libiio.

pcercuei:
What's the issue with upstream libini?

I have no problem if libini is added as a submodule, it would probably be a good idea since it's not packaged anywhere.

rgetz:
Issue with upstream libini is discussed at #403 

It's not a big deal - it's just a matter if you wanted to maintain this in your personal repo forever (as long as people are using libiio), or if it should be moved somewhere else?

-Robin

rgetz:
I opened #428 - I think it's the way to go.

rgetz:
I will close this, unmerged, as it looks like #428 will be merged instead.

PR #406, opened 2020-03-17
iiod: allow multiple "iio" services to be published on DNS-SD

### LINKED ISSUE #402: IIOD DNS-SD: Only one iio service can be registered on the network
With current implementation, where service name and type are hardcoded in iiod, only first service is successfully published with DNS-SD, subsequent fail because the type/name pair collides with the existing.

```
	if (group && !avahi_entry_group_add_service(group,
			AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC,
			0, "iio", "_iio._tcp", NULL, NULL, IIOD_PORT, NULL)) {
		avahi_entry_group_commit(group);
		INFO("Registered to ZeroConf server %s\n",
				avahi_client_get_version_string(client));
	}
```

"iio" service name shall be replaced with something that is unique on the network.

I propose using hostname.

rgetz:
This fixes #402 by handling more error codes in iiod for avahi.
and more closely matches libavahi examples.

This seems to work for me, I tried creating a few network
contexts, and didn't have any issues finding the various ones


iiod/iiod.c
```diff
@@ -106,57 +110,255 @@ static const char *options_descriptions[] = {
...
+	if (!avahi.group) {
+		avahi.group = avahi_entry_group_new(c, __avahi_group_cb, NULL);
+		if (!avahi.group) {
+			ERROR("avahi_entry_group_new() failed: %s\n",
+					avahi_strerror(avahi_client_errno(c)));
+			goto fail;
+		}
+	}
+
+	if (avahi_entry_group_is_empty(avahi.group)) {
```
commodo: this can be left as-is;
but this can save one level of indentation [for this if-block] by doing:
```
if (!avahi_entry_group_is_empty(avahi.group)) {
    DEBUG("Avahi group not empty\n");
     return;
}
```

iiod/iiod.c
```diff
@@ -106,57 +110,255 @@ static const char *options_descriptions[] = {
...
+			break;
+		case AVAHI_ENTRY_GROUP_COLLISION : {
+			char *n;
+			/* A service name collision, pick a new name */
+			n = avahi_alternative_service_name(avahi.name);
+			avahi_free(avahi.name);
+			avahi.name = n;
+			INFO("Avahi: Group Service name collision, renaming service to '%s'\n",
+					avahi.name);
+			create_services(avahi_entry_group_get_client(group));
```
commodo: curios: is there a chance that services need to be free'd here before create_services()?

iiod/iiod.c
```diff
@@ -106,57 +110,255 @@ static const char *options_descriptions[] = {
...
 static void __avahi_group_cb(AvahiEntryGroup *group,
 		AvahiEntryGroupState state, void *d)
 {
+	/* Called whenever the entry group state changes */
+	if (!group)  {
+		ERROR("__avahi_group_cb with no valid group\n");
+		return;
+	}
+
+	avahi.group = group;
```
commodo: can `avahi.group` be non-null?
if so, do we need to handle it?

iiod/iiod.c
```diff
@@ -106,57 +110,255 @@ static const char *options_descriptions[] = {
...
+		avahi.group = avahi_entry_group_new(c, __avahi_group_cb, NULL);
+		if (!avahi.group) {
+			ERROR("avahi_entry_group_new() failed: %s\n",
+					avahi_strerror(avahi_client_errno(c)));
+			goto fail;
+		}
+	}
+
+	if (avahi_entry_group_is_empty(avahi.group)) {
+		ret = avahi_entry_group_add_service(avahi.group,
```
commodo: are you sure this registers more than 1 service?
the way i'm reading the code is: `if (group_empty) (add_service()`; which may mean that only one service can get added;

iiod/iiod.c
```diff
@@ -106,57 +110,255 @@ static const char *options_descriptions[] = {
...
-			0, "iio", "_iio._tcp", NULL, NULL, IIOD_PORT, NULL)) {
-		avahi_entry_group_commit(group);
-		INFO("Registered to ZeroConf server %s\n",
-				avahi_client_get_version_string(client));
+	switch (state) {
+		case AVAHI_CLIENT_S_RUNNING:
+			/* Same as AVAHI_SERVER_RUNNING */
+			DEBUG("Avahi: create services\n");
+			/* The server has startup successfully, so create our services */
+			create_services(client);
```
commodo: is there a chance that `create_services()` requires some cleanup before getting called?
i.e. can the state-machine of avahi do something like `RUNNING, <other-state>, RUNNING`?

iiod/iiod.c
commodo: ah; nevermind; just saw the null check

iiod/iiod.c
commodo: hmm; this looks the same in the example;
hopefully it's fine

iiod/iiod.c
commodo: looks the same in the example; maybe it's fine;

rgetz: only one service does get added, it's the "iiod" service - which sometimes collides with other "iiod" services, and needs to be added again with a different name, that's all.

iiod/iiod.c
rgetz: thanks

iiod/iiod.c
rgetz: There isn't any cleanup, in create_services(), it checks if the group exists, if not it creates it, makes sure the group is empty (the first time being called), and adds the service to the newly created group.

avahi_entry_group_reset(avahi.group); will reset the group the way that the libavahi likes it; which is called in the necessary places (according to the example) before create_services();

iiod/iiod.c
rgetz: should be updated in latest pull request.

iiod/iiod.c
rgetz: no, when necessary - group_reset() is called. this does what avahi wants.

PR #410, opened 2020-03-27
iio_writedev: fix off by one bug when pushing buffers.

cristi-iacob:
When requesting num_sample samples and num_samples is a multiple of buffer_size, the last data bulk of size buffer_size (i.e. the last buffer) is not pushed to the device.
This is happening because the program is entering the if (!num_samples) statement (line 460), closing the buffer before reaching the line where the buffer is pushed.
To solve this problem, I've modified the if statement, adding the condition that the buffer should not be filled up when calling the quit_all() method.
Therefore, the program will now push the last buffer and quit on the next iteration, when the buffer will be empty.
In this situation, num_samples will become 0 and the program will behave like no num_samples was provided (and therefore pushing all the remaining samples from stdin).
To solve this case, I have added a num_samples_set flag, which is True if and only if num_samples war provided in the program call.

### DIFF
--- a/tests/iio_writedev.c
+++ b/tests/iio_writedev.c
@@ -421,6 +421,7 @@ int main(int argc, char **argv)
 	}
 
 	sample_size = iio_device_get_sample_size(dev);
+	bool num_samples_set = num_samples > 0;
 
 	buffer = iio_device_create_buffer(dev, buffer_size, cyclic_buffer);
 	if (!buffer) {
@@ -443,7 +444,7 @@ int main(int argc, char **argv)
 			size_t write_len, len = (intptr_t) iio_buffer_end(buffer)
 				- (intptr_t) start;
 
-			if (num_samples && len > num_samples * sample_size)
+			if (num_samples_set && len > num_samples * sample_size)
 				len = num_samples * sample_size;
 
 			for (write_len = len; len; ) {
@@ -455,9 +456,9 @@ int main(int argc, char **argv)
 				start = (void *)((intptr_t) start + nb);
 			}
 
-			if (num_samples) {
+			if (num_samples_set) {
 				num_samples -= write_len / sample_size;
-				if (!num_samples)
+				if (!num_samples && write_len < buffer_size * sample_size)
 					quit_all(EXIT_SUCCESS);
 			}
 		} else {


tests/iio_writedev.c
```diff
@@ -421,6 +421,7 @@ int main(int argc, char **argv)
 	}
 
 	sample_size = iio_device_get_sample_size(dev);
+	bool num_samples_set = num_samples > 0;
```
commodo: can `num_samples` be negative?
if so, maybe we should add a check in another patch and exit early;
maybe here:
```
		case 's':
			num_samples = atoi(optarg);
			break;
```
the reason i am asking, is because this is checking `num_samples > 0;`

in any case, adding an extra variable `num_samples_set` seems to be extra;
you can probably remove it & and just check  `if (num_samples)`  or `if (!num_samples)` where needed

tests/iio_writedev.c
```diff
@@ -455,9 +456,9 @@ int main(int argc, char **argv)
 				start = (void *)((intptr_t) start + nb);
 			}
 
-			if (num_samples) {
+			if (num_samples_set) {
 				num_samples -= write_len / sample_size;
-				if (!num_samples)
+				if (!num_samples && write_len < buffer_size * sample_size)
 					quit_all(EXIT_SUCCESS);
```
commodo: would it work to just do ?:
```
if (!num_samples)
    app_running = false;
```

this would also make sure that we go through the proper exit-path below:
```
	iio_buffer_destroy(buffer);
	iio_context_destroy(ctx);
	return exit_code;
```

tests/iio_writedev.c
cristi-iacob: num_samples can't be negative; it is of type size_t, which is an unsigned data type. 
I have added the num_samples_set variable because when num_samples becomes 0, these lines `if (num_samples && len > num_samples * sample_size)
				len = num_samples * sample_size;`
will be ignored, letting len = buffer_size. Therefore, from now on, the program will behave like no num_samples was provided and will push all the available samples from stdin (after pushing the requested number of samples).

tests/iio_writedev.c
cristi-iacob: No, it doesn't work. This way it will allow the program to send an not entirely filled up buffer (for example, when requesting 252 samples, having a buffer of size 256).
I think that the exit-path will be executed in all scenarios anyway.

tests/iio_writedev.c
rgetz: so, how about `bool num_samples_zero = (num_samples != 0)` ? might be more intuitive.

tests/iio_writedev.c
commodo: ah; you're right here;
`quit_all()` sets `app_running = false;`

one point i feel is worth resolving, is if we can get an underflow with this new condition check `if (!num_samples && write_len < buffer_size * sample_size)`
```
iterN: num_samples -= write_len / sample_size;
##  num_samples becomes zero in the above iterN
iterN + 1: num_samples -= write_len / sample_size; 
## num_samples underflows after this? 
## or does the condition `if (!num_samples && write_len < buffer_size * sample_size)` still hold here?
```

so, then another question: what parts of the loop do we need to do on exit?
would it be easier to do? [1]
```
if (!num_samples) {
      /* send last sample in case we need to send last sample; any idea how much code this would be? */
     quit_all(EXIT_SUCCESS);
}
```
or how much code do we need to add above to make this work?

for reviewing at least, it would be a bit easier o review the change above, since i don't need to properly understand the whole loop;
you've found a bug in this loop [and that's great];
your proposal is to go through the whole loop one last time with some condition check-change;
since the loop already has a bug [which this is fixing], I would avoid going through it [if possible], since it may have chances of introducing another bug in favor of fixing this one;
obviously, who wrote this loop did not understand it too well either [i did not check who wrote it];
and i would avoid trying to run it, as that would take some time;

as an epilogue to the comment above: any idea if we need to fix `iio_readdev.c` too [in a similar fashion]?
we can leave that for later when we run into it;

cristi-iacob: 1. Underflow

No underflow will occur. 
```
if (num_samples) {
	num_samples -= write_len / sample_size
...
```

On iter N + 1 num_samples is 0, therefore the program will not enter inside the if statement and won't decrease num_samples again, so num_samples won't underflow. Furthermore, these lines 

```
if (num_samples && len > num_samples * sample_size)
	len = num_samples * sample_size;

	for (write_len = len; len; ) {
...
```

assure that num_samples won't be decreased with a greater value than itself.

2. > what parts of the loop do we need to do on exit? would it be easier to do? or how much code do we need to add above to make this work?

We can add the push-samples logic in this block as well, making it look like this:
```
if (!num_samples) {
        int ret = iio_buffer_push(buffer);
	if (ret < 0) {
		if (app_running) {
			char buf[256];
			iio_strerror(-ret, buf, sizeof(buf));
			fprintf(stderr, "Unable to push buffer: %s\n", buf);
		}
		break;
	}
	quit_all(EXIT_SUCCESS);
}
```
The idea which made me implement it the way I did, was to keep the big while-loop logic. For example, if you request 400 samples with a buffer of size 256, it will send the first 256 samples and on the next iteration it will try to push the remaining 144 samples; now it will find out that there are not enough samples to fill up the buffer and therefore quit. In the same way, when requesting 256 samples with a buffer of size 256, it will push 256 and on the next iteration will try to push the remaining samples (0), will find out that there are not enough samples to fill up the buffer and therefore quit. 
The difference when pushing the last samples before calling quit_all is that the big while-loop will not be executed for the exit-conditioning situation (i.e. not having enough samples to fill up the buffer) and therefore changing a little bit the generalization of the big while-loop.
I think that if we add the push-logic before the quit_all call is like treating a corner case, losing the generality of the big while-loop; on the other hand, letting the program enter the while-loop one more time (for the exit condition) keeps the intended flow of the while-loop. 
I have tested this modification on nearly 20 use-cases and it seems to work fine. 

3. iio_readdev

It seems like there is not the case of this bug, because readdev refills the buffer first and afterwards prints the requested number of samples to stdout. The difference to writedev is that the last buffer refill is not conditioned from num_samples in the way it is in writedev, because the refill is done before updating num_samples.

tests/iio_writedev.c
cristi-iacob: Wouldn't it be more intuitive this way:
`bool num_samples_zero = (num_samples == 0)`, num_samples_zero being true if num_samples is 0? 
On the other hand, wouldn't it be a little confusing this way, in the situation in which num_samples becomes 0 (after pushing some samples)? I think on the situation in which someone is debugging the program and expecting num_samples_zero to become true when num_samples becomes 0.

tests/iio_writedev.c
commodo: > if (num_samples) {
> 	num_samples -= write_len / sample_size
> ...

your new code is
```
if (num_samples_set) {
     num_samples -= write_len / sample_size
	if (!num_samples && write_len < buffer_size * sample_size)
		quit_all(EXIT_SUCCESS);
} 
```

`num_samples_set` is set once; so you can get into this place even after num_samples == 0

> if (!num_samples) {
>         int ret = iio_buffer_push(buffer);
> 	if (ret < 0) {
> 		if (app_running) {
> 			char buf[256];
> 			iio_strerror(-ret, buf, sizeof(buf));
> 			fprintf(stderr, "Unable to push buffer: %s\n", buf);
> 		}
> 		break;
> 	}
> 	quit_all(EXIT_SUCCESS);
> }

isn't this equivalent to?
```
if (!num_samples)
    app_running = false
```

cristi-iacob: > num_samples_set is set once; so you can get into this place even after num_samples == 0

Yes, you are right; sorry for the misunderstanding. Although, in this situation write_len will be 0 (because of these lines)
```
if (num_samples_set && len > num_samples * sample_size)
	len = num_samples * sample_size;

for (write_len = len; len; ) {
```

and therefore num_samples will remain 0 and will not underflow. 

> isn't this equivalent to?
> if (!num_samples)
>  app_running = false

Yes, they are (almost) equivalent. To make this work, we should probably change it this way:
```
if (!num_samples) {
        if (buffer is filled up) {
            // push buffer
        }
        quit_all();
}
```

Basically the condition I have added `&& write_len < buffer_size * sample_size` is for exactly the same thing (checking if the buffer is filled up).

commodo: there's a subtle difference between [1]:
```
if (!num_samples) {
        if (write_len < buffer_size * sample_size) {
            // push buffer
        }
        quit_all();
}
```

and [2]
```
if (!num_samples && write_len < buffer_size * sample_size) {
        // push buffer
        quit_all();
}
```

whether it makes any different here or not, i don't know;

but if [1] works, I prefer that; regardless of whether  `write_len` is zero or not;

cristi-iacob: The problem with [1] is that a call with sum_samples = 256, buffer_size = 256, cyclic = True won't work, because quit_all() will set app_running on False before entering the cyclic-responsible while loop. Therefore it will push these 256 samples and the program will quit, without running in a cyclic fashion.

[2] doesn't run either.

mhennerich: ok - unless we have a better idea to solve this let's got with this patch.

dNechita:
I will take another look at this today.

mhennerich:
Why did we close this?

cristi-iacob:
I have opened another one on the same issue, because the discussions from here were not necessarily related to the new solution.
It can be found here https://github.com/analogdevicesinc/libiio/pull/431

mhennerich:
It has the same subject and trying to solve the same issue. Or do I miss something. Simply force push your new  version and mark some discussions no longer relevant as resolved.

Issue #397, opened 2020-03-09
code formatting


rgetz:
with so many people working on libiio, It's time to put some formatting rules in place...

I was looking at clang-format, since we can include that as a pull request check, similar to:
https://github.com/pierremoreau/SPIRV-LLVM-Translator/blob/master/utils/check_code_format.sh
(which checks only the diff).

I think I tried to minimize things - but it's alot of current inconsistencies...

Things like screen width:

75 columns :  45 files changed, 3822 insertions(+), 3051 deletions(-)
76 columns :  45 files changed, 3720 insertions(+), 3032 deletions(-)
77 columns :  44 files changed, 3599 insertions(+), 2993 deletions(-)
78 columns :  44 files changed, 3535 insertions(+), 2986 deletions(-)
79 columns :  44 files changed, 3476 insertions(+), 3005 deletions(-)
80 columns :  44 files changed, 3365 insertions(+), 2980 deletions(-)
81 columns :  44 files changed, 3346 insertions(+), 3016 deletions(-)
82 columns :  44 files changed, 3309 insertions(+), 3042 deletions(-)
83 columns :  44 files changed, 3288 insertions(+), 3075 deletions(-)
84 columns :  44 files changed, 3255 insertions(+), 3103 deletions(-)
85 columns :  44 files changed, 3220 insertions(+), 3128 deletions(-)

some are obvious, just based on the number of changes:
AlignAfterOpenBracket: DontAlign  44 files changed, 3365 insertions(+), 2980 deletions(-)
AlignAfterOpenBracket: Align  44 files changed, 3781 insertions(+), 3263 deletions(-)

-ContinuationIndentWidth: 16 44 files changed, 3365 insertions(+), 2980 deletions(-)
+ContinuationIndentWidth: 8 44 files changed, 3672 insertions(+), 3316 deletions(-)

-IndentCaseLabels: true  44 files changed, 3365 insertions(+), 2980 deletions(-)
+IndentCaseLabels: false  44 files changed, 3087 insertions(+), 2755 deletions(-)

-SpaceAfterCStyleCast: true 44 files changed, 3087 insertions(+), 2755 deletions(-)
+SpaceAfterCStyleCast: false 44 files changed, 3289 insertions(+), 2962 deletions(-)
```
-const char * iio_get_backend(unsigned int index)
+const char *iio_get_backend(unsigned int index)
```

1) Is this a worthwhile thing?

[...]

PR #411, opened 2020-03-27
Cleanup includes

larsclausen:
Use a consistent style for include directives throughout the project.
  * Sort includes in a group alphabetically
  * Include global headers before local headers

The main motivation behind this is to eliminate duplicated includes.
Secondary motivation is to have consistent style.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -15,10 +15,10 @@
  * Lesser General Public License for more details.
  */
 
+#include <string.h>
+
 #include "iio-config.h"
 #include "iio-private.h"
-
-#include <string.h>
 
 unsigned int iio_get_backends_count(void)
 {
--- a/buffer.c
+++ b/buffer.c
@@ -16,11 +16,11 @@
  *
  * */
 
-#include "iio-config.h"
-#include "iio-private.h"
-
 #include <errno.h>
 #include <string.h>
+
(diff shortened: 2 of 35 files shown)


rgetz:
I think this will happen with the appropriate .clang-format which was discussed in #397 

You are just a little ahead of me...

the linux `.clangformat` uses:
```
IncludeCategories:
  - Regex: '.*'
    Priority: 1
```

which I think just sorts everything alphabetically  - I think what you are looking for is something like:
```
IncludeCategories: 
  - Regex:           '^<.*\.h>'
    Priority:        1
  - Regex:           '^<.*'
    Priority:        2
  - Regex:           '.*'
    Priority:        3
```

haven't tested that yet - just by the [Doc](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) - scroll down to the _IncludeCategories_ section.

rgetz:
I did add #413, which I think covers all the changes in here (and more).

Let me know if you seen any differences - the only one I noticed different was in lock.c, where your change was split by a `#ifdef` which clang-format will not change. Let me know if you see others.

If you are Ok with it - I will close this (without merging) and just continue to work on the larger version.

larsclausen:
looks ok.

PR #412, opened 2020-03-28
dns-sd (avahi/bonjour) improvements

### LINKED ISSUE #391: filtering issues with avahi/bonjour
As discussed in #388 we aren't doing something right in the avahi/bonjour code - since many devices (non-IIO) respond back when looking on the network.

rgetz:
This makes dns-sd work on Linux and MacOS for scan contexts. This still needs a Windows implementation. 

On MAC:
```
analogs-mini:build analog$ ./tests/iio_info -s
Available contexts:
	0: 192.168.86.35 (ad7291-ccbox,ad9361-phy,xadc,cf-ad9361-dds-core-lpc,adis16460,cf-ad9361-lpc) [ip:analog-2.local.]
```

on Linux (different network):
```
getz@brain:~/github/libiio$ ./build/tests/iio_info -s
Library version: 0.19 (git tag: d5b4247)
Available contexts:
	0: 0456:b673 (Analog Devices Inc. PlutoSDR (ADALM-PLUTO)), serial=104473222a87000618000600473ed57ae0 [usb:3.28.5]
	1: 192.168.1.120 (ad7291,ad9361-phy,xadc,cf-ad9361-dds-core-lpc,cf-ad9361-lpc) [ip:analog.local]
```

This should also resolve #391

### DIFF
--- a/dns_sd.c
+++ b/dns_sd.c
@@ -59,6 +59,93 @@ static void dnssd_remove_node(struct dns_sd_discovery_data **ddata, int n)
 	*ddata = d;
 }
 
+/* The only way to support scan context from the network is when
+ * DNS Service Discovery is turned on
+ */
+
+struct iio_scan_backend_context {
+	struct addrinfo *res;
+};
+
+static int dnssd_fill_context_info(struct iio_context_info *info,
+		char *hostname, char *addr_str, int port)
+{
+	struct iio_context *ctx;
+	char uri[MAXHOSTNAMELEN + 3];
+	char description[255], *p;
+	const char *hw_model, *serial;
+	int i;
+
+	ctx = network_create_context(addr_str);
+	if (!ctx) {
+		ERROR("No context at %s\n", addr_str);
+		return -ENOMEM;
+	}
+
+	if (port == IIOD_PORT)
+		sprintf(uri, "ip:%s", hostname);
+	else
+		sprintf(uri, "ip:%s:%d", hostname, port);
+
+	hw_model = iio_context_get_attr_value(ctx, "hw_model");
+	serial = iio_context_get_attr_value(ctx, "hw_serial");
+
+	if (hw_model && serial) {
+		snprintf(description, sizeof(description), "%s (%s), serial=%s",
+				addr_str, hw_model, serial);
+	} else if (hw_model) {
+		snprintf(description, sizeof(description), "%s %s", addr_str, hw_model);
+	} else if (serial) {
+		snprintf(description, sizeof(description), "%s %s", addr_str, serial);
+	} else if (ctx->nb_devices == 0) {
+		snprintf(description, sizeof(description), "%s", ctx->description);
+	} else {
+		snprintf(description, sizeof(description), "%s (", addr_str);
+		p = description + strlen(description);
+		for (i = 0; i < ctx->nb_devices - 1; i++) {
+			if (ctx->devices[i]->name) {
+				snprintf(p, sizeof(description) - strlen(description) -1,
+						"%s,",  ctx->devices[i]->name);
+				p += strlen(p);
+			}
+		}
+		p--;
+		*p = ')';
+	}
+
+	iio_context_destroy(ctx);
+
+	info->uri = iio_strdup(uri);
+	if (!info->uri)
+		return -ENOMEM;
+
+	info->description = iio_strdup(description);
+	if (!info->description) {
+		free(info->uri);
+		return -ENOMEM;
+	}
+
+	return 0;
+}
+
+struct iio_scan_backend_context * dnssd_context_scan_init(void)
+{
+	struct iio_scan_backend_context *ctx;
+
+	ctx = malloc(sizeof(*ctx));
+	if (!ctx) {
+		errno = ENOMEM;
+		return NULL;
+	}
+
+	return ctx;
+}
+
+void dnssd_context_scan_free(struct iio_scan_backend_context *ctx)
+{
+	free(ctx);
+}
+
 /*
  * remove the ones in the list that you can't connect to
  * This is sort of silly, but we have seen non-iio devices advertised
@@ -145,6 +232,42 @@ void remove_dup_discovery_data(struct dns_sd_discovery_data **ddata)
 	iio_mutex_unlock(d->lock);
 
 	*ddata = d;
+}
+
+int dnssd_context_scan(struct iio_scan_backend_context *ctx,
+		struct iio_scan_result *scan_result)
+{
+	struct iio_context_info **info;
+	struct dns_sd_discovery_data *ddata, *ndata;
+	int ret = 0;
+
+	ret = dnssd_find_hosts(&ddata);
+
+	/* if we return an error when no devices are found, then other scans will fail */
+	if (ret == -ENXIO)
+		return 0;
+
+	if (ret < 0)
+		return ret;
+
+	for (ndata = ddata; ndata->next != NULL; ndata = ndata->next) {
+		info = iio_scan_result_add(scan_result, 1);
+		if (!info) {
+			ERROR("Out of memory when adding new scan result\n");
+			ret = -ENOMEM;
+			break;
+		}
+
+		ret = dnssd_fill_context_info(*info,
+				ndata->hostname, ndata->addr_str,ndata->port);
+		if (ret < 0) {
+			DEBUG("Failed to add %s (%s) err: %d\n", ndata->hostname, ndata->addr_str, ret);
+			break;
+		}
+	}
+
+	dnssd_free_all_discovery_data(ddata);
+	return ret;
 }
 
 int dnssd_discover_host(char *addr_str, size_t addr_len, uint16_t *port)
--- a/dns_sd.c
+++ b/dns_sd.c
@@ -1,0 +1,175 @@
+/*
+ * libiio - Library for interfacing industrial I/O (IIO) devices
+ *
+ * Copyright (C) 2014-2020 Analog Devices, Inc.
+ * Author: Robin Getz <robin.getz@analog.com>
+ *         Matej Kenda <matejken@gmail.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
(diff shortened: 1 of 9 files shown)


rgetz:
fixed a codacy issue https://app.codacy.com/manual/commodo/libiio/pullRequests (which is now OK)
and a Windows Build issues (which is also now OK).

dns_sd.c
```diff
@@ -0,0 +1,172 @@
...
+			DEBUG("Unable to find host ('%s'): %s\n",
+					ndata->hostname,
+					gai_strerror(ret));
+			dnssd_remove_node(&d, i);
+		} else {
+			for (rp = res; rp != NULL; rp = rp->ai_next) {
+				fd = create_socket(rp, DEFAULT_TIMEOUT_MS);
+				if (fd < 0) {
+					DEBUG("Unable to open %s%s socket ('%s:%d' %s)\n",
+							res->ai_family == AF_INET ? "ipv4" : "",
```
commodo: should this be  `rp` vs `res` ?
i.e. `res->ai_family == AF_INET` ?

not a big one, since it's mostly debugging print;

dns_sd.c
```diff
@@ -0,0 +1,172 @@
...
+}
+
+int dnssd_discover_host(char *addr_str, size_t addr_len, uint16_t *port)
+{
+	struct dns_sd_discovery_data *ddata;
+	int ret = 0;
+
+	ret = dnssd_find_hosts(&ddata);
+
+	if (ddata) {
```
commodo: we should either check that `ret < 0` or initialize `ddata` to NULL;
on x86 I've seen that values on stack get initialized to zero [most times], but on ARM it can be anything, and then you get weird behavior/crashes;

dns_sd_bonjour.c
```diff
@@ -44,124 +52,197 @@ static void __cfnet_browser_cb (
...
+	CFStreamError	anError;
+
 	if ((flags & kCFNetServiceFlagIsDomain) != 0) {
-		ERROR("DNS SD: FATAL. Callback called for domain, not service.\n");
+		ERROR("DNS SD: FATAL! Callback called for domain, not service.\n");
 		goto stop_browsing;
 	}
 	
-	struct cfnet_discovery_data *dd = (struct cfnet_discovery_data *)info;
+	struct dns_sd_discovery_data *dd = (struct dns_sd_discovery_data *)info;
```
commodo: this isn't new to the file;
but these declarations in the middle of the code-block are C99 specific;
do we want to maintain/have C90 compatibility?

dns_sd_avahi.c
```diff
@@ -14,42 +15,115 @@
...
-	AvahiAddress *address;
-	uint16_t *port;
-	bool found, resolved;
-};
+/*
+ * Fundamentally, this builds up a linked list to manage
+ * potential clients on the network
+ */
+
+static int new_discovery_data(struct dns_sd_discovery_data **data)
```
commodo: this function seems to be defined twice in the same patch;
it makes it a bit hard to follow;

but then also a question: is it meant to be the same thing [in both files] or have different behavior in each?

dns_sd_bonjour.c
```diff
@@ -44,124 +52,197 @@ static void __cfnet_browser_cb (
...
+		WARNING("DNS SD: Can't resolve valid address for service %s.\n", name);
+		goto exit;
+	}
+
+	/* Set properties on the last element on the list. */
+	while (dd->next)
+		dd = dd->next;
+
+	dd->port = port;
+	dd->hostname = strdup(hostname);
```
commodo: i can't seem to find any equivalent `free(dd->hostname)`;
does avahi handle that?

dns_sd_bonjour.c
```diff
@@ -44,124 +52,197 @@ static void __cfnet_browser_cb (
...
+	if (have_v4) {
+		strncpy(dd->addr_str, address_v4, sizeof(dd->addr_str));
+	} else if(have_v6) {
+		strncpy(dd->addr_str, address_v6, sizeof(dd->addr_str));
+	} else {
```
matejk: else is redundant because of the check in line 134.

dns_sd_bonjour.c
```diff
@@ -44,124 +52,197 @@ static void __cfnet_browser_cb (
...
+		CFRunLoopRunResult runRes = CFRunLoopRunInMode(kCFRunLoopDefaultMode, 2, TRUE);
+
+		if (runRes != kCFRunLoopRunHandledSource) {
+			if (runRes == kCFRunLoopRunFinished)
+				ERROR("DSN SD: CFRunLoopRunInMode completed kCFRunLoopRunFinished (%d)\n", runRes);
+			else if (runRes == kCFRunLoopRunStopped)
+				ERROR("DSN SD: CFRunLoopRunInMode completed kCFRunLoopRunStopped (%d)\n", runRes);
+			else if (runRes == kCFRunLoopRunTimedOut)
+				ERROR("DSN SD: CFRunLoopRunInMode completed kCFRunLoopRunTimedOut (%d)\n", runRes);
```
matejk: IMHO timeout is not an error but indication that no iio services were found.
```
$ iio.framework/Tools/iio_info -s
Library version: 0.19 (git tag: 478b1ac)
Compiled with backends: xml ip
ERROR: DSN SD: CFRunLoopRunInMode completed kCFRunLoopRunTimedOut (3)
No contexts found.
```

dns_sd.c
```diff
@@ -0,0 +1,295 @@
...
+void dnssd_context_scan_free(struct iio_scan_backend_context *ctx)
+{
+	free(ctx);
+}
+
+/*
+ * remove the ones in the list that you can't connect to
+ * This is sort of silly, but we have seen non-iio devices advertised
+ * and discovered on the network. Oh well....
+ */
```
matejk: Were these "ghost" devices a side effect of publishing iio service with Avahi using the iio.service file and with code in iiod at the same time?

dns_sd_bonjour.c
```diff
@@ -44,124 +52,197 @@ static void __cfnet_browser_cb (
...
 	
-	dd->port = CFNetServiceGetPortNumber(netService);
+	CFStringRef targetHost = CFNetServiceGetTargetHost(netService);
+	if (targetHost == NULL) {
+		DEBUG("DNS SD: No valid target host for service.\n");
+		goto exit;
+	}
+
+	char hostname[MAXHOSTNAMELEN];
+	if (!CFStringGetCString(targetHost, hostname, sizeof(hostname), kCFStringEncodingASCII)) {
```
matejk: Why is `kCFStringEncodingASCII` hardcoded here?

Would't it be better to user current system encoding?
```
CFStringEncoding encodingMethod = CFStringGetSystemEncoding();
```

dns_sd_bonjour.c
```diff
@@ -44,124 +52,197 @@ static void __cfnet_browser_cb (
...
 		goto exit;
 	}
-	result = CFNetServiceBrowserSearchForServices(
-		gServiceBrowserRef, domain, type, &error);
+
+	CFRunLoopRef runLoop = CFRunLoopGetCurrent();
+	CFNetServiceBrowserScheduleWithRunLoop(serviceBrowser, runLoop, kCFRunLoopDefaultMode);
+
+	CFStringRef type = CFSTR("_iio._tcp.");
+	CFStringRef domain = CFSTR("local.");
```
matejk: I think that domain shall be empty string to be consistent with Avahi implementation. It is stated on [Apple's page](https://developer.apple.com/library/archive/documentation/Networking/Conceptual/NSNetServiceProgGuide/Articles/CFNetServices.html) that:

> Domain. The domain for the service, typically the empty string (CFSTR(""))for default domain(s), or local. for the local domain only

dns_sd_bonjour.c
rgetz: I think we have lots of C99 isms in the codebase already - don't we?

dns_sd_avahi.c
rgetz: yes, there are some implementation specific (avahi on Linux vs Bonjour on macos) things in the different dns_sd_avahi and dns_sd_bonjour files - some are specific to the library being used, some are specific to the OS. 

The common file dns_sd.c calls them with the same name.

dns_sd.c
rgetz: No - this happens when the service file is installed with no backing iiod.

dns_sd_bonjour.c
rgetz: Not what I saw others doing.

The hostname, you need to pass to other networking functions, not print out to people, and the other system level networking functions want ASCII.

dns_sd_bonjour.c
rgetz: and since we want the local domain only - isn't this what we _do_ want?

dns_sd.c
rgetz: good eye. Thanks - fixed.

dns_sd.c
rgetz: `dnssd_find_hosts()` allocates memory or returns NULL on most failures. (just looking with a better eye now), so yes - you are right.

fixed.

dns_sd_bonjour.c
rgetz: fixed.

dns_sd_bonjour.c
rgetz: thanks - fixed.

dns_sd_bonjour.c
rgetz: fixed

dns_sd_avahi.c
commodo: ack

dns_sd_bonjour.c
commodo: yep; don't mind c99 or c90 or c11;
let's hope nobody asks us for anything old

dns_sd_bonjour.c
matejk: I compared with the Avahi implementation where NULL is passed for domain and[ Avahi documentation](https://www.avahi.org/doxygen/html/lookup_8h.html#a52d55a5156a7943012d03e6700880d2b) says:

> avahi_service_browser_new()
> 
> domain | A domain to browse in. In most cases you want to pass NULL here for the default domain (usually ".local")

That sounds to me the same like the passing `CFSTR("")` on macOS.

rgetz: Ok - I updated to a empty string.

commodo:
i'll leave other stuff to @matejk 
he seems to have a better handle on this Avahi stuff;
i don't see any other issues here;

rgetz:
Thanks for your review.

matejk:
> i'll leave other stuff to @matejk
> he seems to have a better handle on this Avahi stuff;
> i don't see any other issues here;

I don't have any comments on this also. We did a good job here, I think.

dns_sd.c
matejk: OK.

dns_sd_bonjour.c
matejk: OK.

PR #413, opened 2020-03-29
updating and codify coding style

rgetz:
This is step 1 for #397 

this is a draft for comment. It's expected that people will comment on this, and maybe a few tweaks to the `.clang-format` files will be made to to things before this is merged to master. (it will also need to be re-based based on what else is merged to master as well.

We also may want to include some python linting and potentially [Cmake Format](https://github.com/cheshirekow/cmake_format)


local.c
```diff
@@ -273,21 +266,20 @@ static int get_rel_timeout_ms(struct timespec *start, unsigned int timeout_rel)
...
-		}
-	};
+			      struct timespec *start)
+{
+	struct pollfd pollfd[2] = { {
+						    .fd = dev->pdata->fd,
+						    .events = events,
+				    },
+				    {
+						    .fd = dev->pdata->cancel_fd,
```
mhennerich: Hmm - Placing Braces - I don't think this is proper Linux coding style. In fact you can find this or the other upstream.
But my preference is to put the opening brace last on the line '}, {'. This is more in sync with 
https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces

examples/dummy-iiostream.c
```diff
@@ -315,27 +322,43 @@ int main (int argc, char **argv)
...
-				for (p_dat = iio_buffer_first(rxbuf, channels[i]); p_dat < p_end; p_dat += p_inc) {
+				for (p_dat = iio_buffer_first(rxbuf,
+							      channels[i]);
+				     p_dat < p_end; p_dat += p_inc) {
 					for (int j = 0; j < repeat; ++j) {
-						if (fmt->length/8 == sizeof(int16_t))
-							printf("%" PRIi16 " ", ((int16_t *)p_dat)[j]);
-						else if (fmt->length/8 == sizeof(int64_t))
-							printf("%" PRId64 " ", ((int64_t *)p_dat)[j]);
+						if (fmt->length / 8 ==
```
commodo: this part looks pretty bad when trying to avoid the 80 col width limit;
we can leave it like this for now, but we may want to move this to it's own function with early-exits

examples/dummy-iiostream.c
```diff
@@ -358,24 +383,38 @@ int main (int argc, char **argv)
...
 				printf("%s ", iio_channel_get_id(channels[i]));
-				for (int sample = 0; sample < bytes / sample_size; ++sample) {
+				for (int sample = 0;
+				     sample < bytes / sample_size; ++sample) {
 					for (int j = 0; j < repeat; ++j) {
-						if (fmt->length / 8 == sizeof(int16_t))
-							printf("%" PRIi16 " ", ((int16_t *)buf)[sample+j]);
-						else if (fmt->length / 8 == sizeof(int64_t))
-							printf("%" PRId64 " ", ((int64_t *)buf)[sample+j]);
+						if (fmt->length / 8 ==
```
commodo: same for this part when trying to avoid the 80 col width limit

examples/iio-monitor.c
```diff
@@ -122,11 +123,13 @@ static struct {
...
-	{ "temp",	"°C" },
-	{ "voltage",	"V" },
-	{ 0, },
+	{ "current", "A" },
+	{ "power", "W" },
+	{ "temp", "°C" },
+	{ "voltage", "V" },
+	{
+			0,
+	},
```
commodo: the null-terminator looks funky;

examples/iio-monitor.c
```diff
@@ -122,11 +123,13 @@ static struct {
 	const char *id;
 	const char *unit;
 } map[] = {
-	{ "current",	"A" },
-	{ "power",	"W" },
-	{ "temp",	"°C" },
-	{ "voltage",	"V" },
-	{ 0, },
+	{ "current", "A" },
```
commodo: curios: is there an option to somehow align table columns?
we can leave it for later;

iio.h
```diff
@@ -376,23 +360,22 @@ __api __check_ret struct iio_context * iio_create_network_context(const char *ho
  *   ('<b>\0</b>' none, 'x' Xon Xoff, 'r' RTSCTS, 'd' DTRDSR) parts separated
  *   with a comma. For example <i>"serial:/dev/ttyUSB0,115200"</i>,
  *   <i>"serial:/dev/ttyUSB0,115200,n,8,1"</i>*/
-__api __check_ret struct iio_context * iio_create_context_from_uri(const char *uri);
-
+__api __check_ret struct iio_context *
+iio_create_context_from_uri(const char *uri);
```
commodo: i wonder if for these, it works to do:
```
__api __check_ret
struct iio_context * iio_create_context_from_uri(const char *uri);
```

if we need to break the line, it may worth putting the specifiers on a separate line

iiod/iiod.c
```diff
@@ -159,110 +158,109 @@ static void shutdown_avahi()
...
+		create_services(avahi_entry_group_get_client(group));
+		break;
+	}
+	case AVAHI_ENTRY_GROUP_FAILURE:
+		ERROR("Entry group failure: %s\n",
+		      avahi_strerror(avahi_client_errno(
+				      avahi_entry_group_get_client(group))));
+		break;
+	case AVAHI_ENTRY_GROUP_UNCOMMITED:
+		/* This is normal,
```
commodo: this comment looks funky indented with the rest of the comment block

iiod/iiod.c
```diff
@@ -658,7 +655,7 @@ int main(int argc, char **argv)
 	else
 		ret = main_server(ctx, debug);
 
-	/*
+		/*
```
commodo: this comment block looks funky indented as well

iiod/ops.h
```diff
@@ -19,31 +19,31 @@
...
+#define __bswap_constant_16(x)                                                 \
+	((unsigned short int)((((x) >> 8) & 0xff) | (((x)&0xff) << 8)))
 #endif
 
 #ifndef __bswap_constant_32
-#define __bswap_constant_32(x) \
-	((((x) & 0xff000000) >> 24) | (((x) & 0x00ff0000) >>  8) | \
-	 (((x) & 0x0000ff00) <<  8) | (((x) & 0x000000ff) << 24))
+#define __bswap_constant_32(x)                                                 \
+	((((x)&0xff000000) >> 24) | (((x)&0x00ff0000) >> 8) |                  \
```
commodo: this is weird; it removes spaces around bitwise op

iio.h
mhennerich: I'm wondering - shouldn't this become 
*iio_create_context_from_uri(const char *uri)

"When declaring pointer data or a function that returns a pointer type, the preferred use of * is adjacent to the data name or function name and not adjacent to the type name. Examples:"

iiod/queue.h
```diff
@@ -81,494 +81,540 @@
...
-
-#define	LIST_HEAD_INITIALIZER(head)					\
-	{ NULL }
-
-#define	LIST_ENTRY(type)						\
-struct {								\
-	struct type *le_next;	/* next element */			\
-	struct type **le_prev;	/* address of previous next element */	\
-}
+#define LIST_HEAD(name, type)                                                  \
```
commodo: macros seems to get spaces vs tags before the backslash;
Linux style usually preferred tabs;
i guess either works if we are consistent

local.c
commodo: yeah, this looks funky indented;

iiod/ops.h
mhennerich: Yeah this weird - maybe a bug in clang-format?

tests/iio_attr.c
```diff
@@ -646,23 +663,27 @@ int main(int argc, char **argv)
...
 		for (i = 0; i < nb_ctx_attrs; i++) {
 			const char *key, *value;
 			ssize_t ret;
 
 			ret = iio_context_get_attr(ctx, i, &key, &value);
 			if (!ret) {
-				if (!attr_index || str_match(key, argv[attr_index], ignore_case)) {
+				if (!attr_index ||
+				    str_match(key, argv[attr_index],
+					      ignore_case)) {
```
commodo: this is another place where the 80-col-width limit seems to cause funky-ness;
maybe we should split it into another function

tests/iio_stresstest.c
```diff
@@ -152,12 +152,7 @@ static struct iio_device * get_device(const struct iio_context *ctx,
 	return NULL;
 }
 
-enum backend {
```
commodo: i like enums that are each on it's own line

tests/iio_stresstest.c
```diff
@@ -335,14 +340,16 @@ static void *client_thread(void *data)
 	if (info->verbose == VERYVERBOSE)
 		printf("%2d: Stopped normal\n", id);
 	info->tid[id] = 0;
-	info->start[id][stamp].tv_sec = 0; info->start[id][stamp].tv_usec = 0;
```
commodo: this fix i like;
the way it was before was a bit weird;

commodo:
i did a quick-pass through the changes; i think doing a thorough pass would have taken me a few hours, and a few more to rest :)

overall, i like the idea of adding a code-style-checker, but i would change the approach;
maybe let's do this patchset in semantic groups; such a patch can't be seriously reviewed, and when we get something on EZ about a libiio bug, then this is the first place we'll look into and ¯\_(ツ)_/¯

I'd say, let's do it like this:
1. add .clang-format with rules that update **only** switch statements; then apply that rule-set
2. update .clang-format with rules that update **only** spaces around instruction blocks; i.e.  `if()` vs `if__space__()`; the apply ruleset
3. update .clang-format with rules that update **only** spaces around pointers; apply that ruleset
.......
no idea about what else should be next here, but these should probably handle a reasonable percentage of this patchset;

some notes:
1. the 80-col-width-limit is a tricky one here;
i would leave it for the last patchset, because it would require to actually rework/cleanup the code in a different manner; i.e. group code into functions;
the kernel coding style guideline's first chapter is about indentation;
https://www.kernel.org/doc/html/v4.10/process/coding-style.html#indentation
when we review on the kernel [particularly the code that will go upstream] we try to discourage stuff like:
```
if () {
    if ()  {
          for() {
          }
    }
}
```
we sometimes get people saying/thinking (no, this can't be reworked with less indentation), and they're usually proven wrong;
2. can we also add a build-rule to run `clang-format`? for a while we can probably ignore the failure, even tell Travis-CI to ignore it; until we finish re-styling everything
3. i proposed astyle for the other projects; reason i proposed it, was because it's quite small and i used in a past project [no idea if clang-format was around at that time]; `clang-format` is slightly bigger; no idea which tool is better on the long-run between the 2;
with astyle, one thing that was a problem [and i'm hoping it won't be with clang-format] is that each version produces a slightly different output than the other;
which means that as you upgrade, some bits of code need to be re-formatted periodically [if upgrading];
for astyle, we locked the version that we build from their website;
this is something to consider as well for clang-format;
4. the C-Sharp changes: I don't know; they look a bit weird and big; do we need to update those?
5. CMake files could also be a separate patchset

rgetz:
The goal in making everything one giant patch, was it will be easier to ignore it later when we are looking at history or git blame. Ignoring one SHA is now built into modern git.

commodo:
> The goal in making everything one giant patch, was it will be easier to ignore it later when we are looking at history or git blame. Ignoring one SHA is now built into modern git.

i'm just worried about stuff slipping in and ending-up with a bad result;
while git can ignore hashes, it can't yet detect which line introduced a regression;
the more dangerous elements in this set are macro-changes; C-macros can be tricky, as sometimes, even spaces can make a difference;

local.c
rgetz: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

have a look at the BraceWrapping options, I think you are talking about:
AfterStruct : false
SplitEmptyRecord : false

examples/dummy-iiostream.c
rgetz: I'm Ok moving things to more than 80 col if that was the suggestion?

examples/iio-monitor.c
rgetz: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

there are alot of align options - I can try to figure it out.

examples/dummy-iiostream.c
commodo: either is fine; we can leave a higher-than-80-col-width;
or we can leave it like this;
though, the way it is now [with the tool applied], is kind of bad;

we can also clean it up later;

iio.h
rgetz: PointerAlignment can be set to Left, Right or Middle.

iiod/ops.h
rgetz: SpacesInParentheses is set to false.

tests/iio_stresstest.c
rgetz: BreakBeforeBraces should be set to Linux

rgetz:
> i'm just worried about stuff slipping in and ending-up with a bad result;

yeah, totally get that - a giant patch wouldn't help bisecting things, but it would help reverting things :)

tests/iio_stresstest.c
rgetz: Agree - reading more doc says:

clang-format contracts enums to a single line if all elements fit on one line. This conserves spaces and usually does not decrease readability. There is no style option, but you can override this by either adding a line comment somewhere or by adding a trailing comma after the last enumerator,

so I will do that manually.

examples/iio-monitor.c
rgetz: I poked around, and It doesn't look like it...

we can add /* clang-format off */ & /* clang-format on */ and that is about it....

iiod/iiod.c
rgetz: yeah, that is weird - I tried to make it go away, but no luck so far...

iio.h
```diff
@@ -134,7 +134,7 @@ enum iio_chan_type {
 	IIO_POSITIONRELATIVE,
 	IIO_PHASE,
 	IIO_MASSCONCENTRATION,
-	IIO_CHAN_TYPE_UNKNOWN = INT_MAX
+	IIO_CHAN_TYPE_UNKNOWN = INT_MAX,
```
commodo, line 138: this is fine;

tests/iio_attr.c
```diff
@@ -41,7 +41,7 @@
 enum backend {
 	LOCAL,
 	XML,
-	AUTO
+	AUTO,
```
commodo, line 45: will `AUTO` always be the last in the enum?
if so, this is fine;
typically the last element in an enum/series gets a comma, mostly for future-proofing the patches;
i.e. the patch becomes just an addition

CMakeLists.txt
```diff
@@ -5,517 +5,604 @@ set(LIBIIO_VERSION_MAJOR 0)
 set(LIBIIO_VERSION_MINOR 19)
 set(VERSION "${LIBIIO_VERSION_MAJOR}.${LIBIIO_VERSION_MINOR}")
 if (WIN32)
-	string(TIMESTAMP BUILD_YEAR "%Y")
-endif()
+    string(TIMESTAMP BUILD_YEAR "%Y")
+endif ()
```
commodo, line 9: the more popular version for the CMake constructs [at least in the CMake files of libiio] seems to be without the space;
i.e. `endif()`
can we make that the rule?
or, is there a convention listed somewhere?

rgetz: I think the CMake formatter will do that (no space). I will update.

tests/iio_attr.c
rgetz: The trailing comma was to fix some formatting issues that clang-format wanted to see.

rgetz:
closing - @pcercuei should look at doing this on the dev branch, shortly before it becomes the main development branch later this year.

PR #414, opened 2020-03-29
examples: Add -Wextra & -pedantic-errors to the example Makefile

rgetz:
and clean up the resulting errors. This resolves a few codacy issues
that have been lingering for awhile. There shouldn't be any functional
changes from these. I tested most, but not all, since they were all
pretty similar.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -16,7 +16,7 @@
 
 TARGETS := ad9361-iiostream ad9371-iiostream adrv9009-iiostream dummy-iiostream iio-monitor
 
-CFLAGS = -Wall -DIIO_CHECK_RET
+CFLAGS = -Wall -Wextra -pedantic -DIIO_CHECK_RET
 
 UNAME_S := $(shell uname -s)
 
--- a/examples/ad9371-iiostream.c
+++ b/examples/ad9371-iiostream.c
@@ -85,7 +85,7 @@ static void shutdown()
 
 static void handle_sig(int sig)
 {
-	printf("Waiting for process to finish...\n");
+	printf("Waiting for process to finish... Got signal %d\n", sig);
 	stop = true;
 }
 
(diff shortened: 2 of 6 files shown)


examples/dummy-iiostream.c
```diff
@@ -227,13 +227,14 @@ int main (int argc, char **argv)
...
+	unsigned int i, j, major, minor;
 	char git_tag[8];
 	iio_library_get_version(&major, &minor, git_tag);
 	printf("Library version: %u.%u (git tag: %s)\n", major, minor, git_tag);
 
-	/* check for struct iio_data_format.repeat support */
-	has_repeat = major >= 0 && minor >= 8 ? true : false;
+	/* check for struct iio_data_format.repeat support
+	 * 0.8 has repeat support, don't need to check major, it's unsigned */
+	has_repeat = minor >= 8 ? true : false;
```
commodo: this looks problematic when we bump version to 1.0;
we may want to make a note to do a `LIBIIO_VERSION(major, minor)    ((major << 8) | minor)` macro or something like that;
or maybe a helper function called  `bool iio_library_version_is_at_least(major, minor)`

obviously, this can be a separate patch;

examples/ad9371-iiostream.c
```diff
@@ -85,7 +85,7 @@ static void shutdown()
 
 static void handle_sig(int sig)
 {
-	printf("Waiting for process to finish...\n");
+	printf("Waiting for process to finish... Got signal %d\n",sig);
```
commodo: nitpick:  printf("Waiting for process to finish... Got signal %d\n",`__space__`sig);

commodo:
1 minor typo in commit comment; but we can ignore it;

the patch looks fine on it's own, but it did expose that some version checks need attention [if possible sometime soon]; otherwise we introduce some regression;

examples/dummy-iiostream.c
commodo: maybe a 2nd thought: is it worth doing this check? we don't recommend people use 0.8 anymore;
so, is it a good time to remove it now? [for such old versions]

examples/Makefile
```diff
@@ -16,7 +16,7 @@
 
 TARGETS := ad9361-iiostream ad9371-iiostream adrv9009-iiostream dummy-iiostream iio-monitor
 
-CFLAGS = -Wall -DIIO_CHECK_RET
+CFLAGS = -Wall -Wextra -pedantic-errors -DIIO_CHECK_RET
```
matejk: Shall `-pedantic-errors` (or just `-pedantic` be set when compiling libiio and iiod also?

rgetz: yes - I switched to `-pedantic`. Thanks

examples/ad9371-iiostream.c
rgetz: fixed

examples/dummy-iiostream.c
rgetz: I fixed this. 

0.8 was released on Nov 4, 2016; so that was awhile ago, since it's all over in the code (has_repeat) - untangling things is a pain.

Issue #396, opened 2020-03-06
sync with kernel


rgetz:
a discussed in #392, we need to set up a travis-ci cron the periodically check kernel headers with ours.
 and notify someone on failure.

rgetz:
This should be resolved as part of #415

PR #415, opened 2020-03-30
check kernel's headers against libiio headers in CI

rgetz:
As discussed in #392 and #396, this adds a small shell script which
downloads the kernel's iio header file, and checks channel types and
channel modifiers to make sure we aren't falling behind.

This will be checked once per Travis-CI build, and takes about 0.5 seconds.

### DIFF
--- a/CI/travis/check_kernel.sh
+++ b/CI/travis/check_kernel.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+set -e
+
+KERNEL_TYPES="/tmp/mainline_types.h"
+IIOH="./iio.h"
+CHANNELC="./channel.c"
+
+if [ ! -f ${IIOH} ] ; then
+	echo can not find ${IIOH}
+	exit 1
+fi
+
+if [ ! -f ${CHANNELC} ] ; then
+	echo can not find ${CHANNELC}
+	exit 1
+fi
+
+rm -f ${KERNEL_TYPES}
+wget -O ${KERNEL_TYPES} https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/include/uapi/linux/iio/types.h
+
+ret=0
+
+for enum in iio_chan_type iio_modifier
+do
+	echo looking for ${enum}
+	rm -f /tmp/kernel_${enum} /tmp/libiio_${enum}
+	sed "0,/${enum}/d" ${KERNEL_TYPES}  | sed -n '/}/q;p' > /tmp/kernel_${enum}
+	sed "0,/^enum.*${enum}/d" ${IIOH} | sed -n '/}/q;p' | grep -v IIO_CHAN_TYPE_UNKNOWN > /tmp/libiio_${enum}
+	echo Differences in ${enum}
+	# diff exit status of 1 means a difference, not an error
+	set +e
+	diff -u  /tmp/libiio_${enum} /tmp/kernel_${enum}
+	count=$(diff -u  /tmp/libiio_${enum} /tmp/kernel_${enum} | wc -l)
+	set -e
+	if [ "$count" -ne "0" ] ; then
+		ret=1
+		echo difference between upstream kernel types.h and iio.h in ${enum}
+	else
+		echo none
+	fi
+done
+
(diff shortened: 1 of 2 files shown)


CI/travis/before_install_linux
```diff
@@ -21,7 +21,7 @@ handle_default() {
 	sudo apt-get -qq update
 	sudo apt-get install -y cmake graphviz libaio-dev libavahi-client-dev \
 		libavahi-common-dev libusb-1.0-0-dev libxml2-dev rpm tar \
-		bzip2 gzip flex bison git python-pip
+		bzip2 gzip flex bison git python-pip wget
```
commodo: `wget` should already be installed via https://github.com/analogdevicesinc/libiio/blob/master/CI/travis/lib.sh#L317

CI/travis/check_kernel.sh
```diff
@@ -0,0 +1,70 @@
...
+	echo can not find ${CHANNELC}
+	exit 1
+fi
+
+rm -f ${KERNEL_TYPES}
+wget -O ${KERNEL_TYPES} https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/include/uapi/linux/iio/types.h
+
+ret=0
+
+for enum in iio_chan_type iio_modifier
```
commodo: what if we change it a bit, to take the contents of `include/uapi/linux/iio/types.h` between `#define _UAPI_IIO_TYPES_H_` and  `#endif /* _UAPI_IIO_TYPES_H_ */` and dump it in `iio.h` between 2  comment markers

something like:
```
......
struct iio_context_info;
struct iio_scan_context;

/* --- added automatically from 'include/uapi/linux/iio/types.h' */

contents of include/uapi/linux/iio/types.h

/* --- end 'include/uapi/linux/iio/types.h' */
```

we could then use  git diff to check if we need to update something;
we could also use the same logic to just update the header [for lazy people like me];

i know there are a few minor downsides [in the beginning]:
* some things do need to be reworked a bit though; like for example, we would need to change `IIO_CHAN_TYPE_UNKNOWN` to be a `#define IIO_CHAN_TYPE_UNKNOWN  UINT_MAX` or something; this would sit outside of that header-dump for a while, but we could also propose that field upstream into the enum;
* we would also get some event definitions; but i also don't see this as a big problem, as it just simplifies a few things when we add event support to libiio

atm, there's only 2 headers for IIO in `include/uapi/linux/iio/`: `types.h` & `events.h` ; assuming things will grow [in there], this sync-mechanism will be a bit more handy & scalable [1]

bonus-mention [related to point [1] ] : the AXI ADC driver has been partially upstreamed [rrecently], and we're currently working to get more high-speed buffer support mainlined; and then the AXI DAC;
with high-speed buffer support [ioctl definitions needing to go in there] this sync mechanism could also be handy

rgetz: Right now - I didn't want to change any of the libiio source - since things are changed so infrequently - I thought checking and error would be fine. When we support events - we can look at re-working things then as part of that.

CI/travis/make_linux
```diff
@@ -5,6 +5,11 @@ if [ "x${COVERITY_SCAN_PROJECT_NAME}" != "x" ] ; then exit 0; fi
 . CI/travis/lib.sh
 
 handle_default() {
+	# make sure we are up to date
+	if [ "$LDIST" = "DO_NOT_DEPLOY" ] ; then
+		./CI/travis/check_kernel.sh
+	fi
```
commodo: oh right; 1 more thing; i kept forgetting to mention this here;
would it be worth to move this to the end of the `handle_default()` function?

it would be an idea to let the compiler have a run first;
those errors would have a slightly higher priority [for this build];

CI/travis/check_kernel.sh
commodo: ok;
fine by me

rgetz:
Ok - updated with your feedback

CI/travis/make_linux
rgetz: fixed

PR #416, opened 2020-03-30
travis-ci: Do a make install so libiio is on system path

rgetz:
This is a small tweak to try to reduce the travis-ci noise
today, we get errors like:

running install
  File "/home/travis/build/analogdevicesinc/libiio/build/bindings/python/setup.py", line 89, in <module>
    setup(**config)
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/travis/build/analogdevicesinc/libiio/build/bindings/python/setup.py", line 43, in run
    self._check_libiio_installed()
  File "/home/travis/build/analogdevicesinc/libiio/build/bindings/python/setup.py", line 66, in _check_libiio_installed
    raise Exception(msg)
Exception: The libiio library could not be found.
            libiio needs to be installed first before the python bindings.
            The latest release can be found on GitHub:
            https://github.com/analogdevicesinc/libiio/releases

Since the libiio isn't installed, and on path during the make package
step. This commit does that.

### DIFF
--- a/CI/travis/make_linux
+++ b/CI/travis/make_linux
@@ -5,6 +5,7 @@ if [ "x${COVERITY_SCAN_PROJECT_NAME}" != "x" ] ; then exit 0; fi
 . CI/travis/lib.sh
 
 handle_default() {
+	echo "### making build dir"
 	mkdir -p build
 	cd build
 
@@ -21,10 +22,22 @@ handle_default() {
 	# set it back, to error immediately.
 	set -e
 
+	echo "### cmake -DENABLE_PACKAGING=ON -DDEB_DETECT_DEPENDENCIES=ON"
+	echo "          -DPYTHON_BINDINGS=ON ${DOC_FOUND}"
 	cmake -DENABLE_PACKAGING=ON -DDEB_DETECT_DEPENDENCIES=ON \
 		-DPYTHON_BINDINGS=ON ${DOC_FOUND} ..
 
+	echo "### make"
 	make
+	if [ "$TRAVIS" = "true" ] ; then
+		# make sure the *.so can be found during make package
+		echo "### Installing package as root"
+		sudo make install
+		# get rid of anything that will cause a problem
+		echo "### Deleting install_manifest.txt (created as root)"
+		sudo rm ./install_manifest.txt
+	fi
+	echo "### make package"
 	make package
 	if [ -n "${GH_DOC_TOKEN}" ] && \
 			[ -f "./generateDocumentationAndDeploy.sh" ] ; then
@@ -36,6 +49,7 @@ handle_default() {
 	if [ "$LDIST" = "DO_NOT_DEPLOY" ] ; then
 		./CI/travis/check_kernel.sh
 	fi
+	echo "### All done building"
 }
 
 handle_centos() {


rgetz:
Unfortunately we can't do a
```
sudo make install
make package
```
since `make install` creates some files in the build directory (as root) and then the `make package` tries to update those files, and ends up with a write error...

commodo:
those errors in Travis-CI should have failed the build;
do you have any idea why they didn't?

commodo:
@tfcollins 
can you take a closer look at?
https://github.com/analogdevicesinc/libiio/blob/master/bindings/python/setup.py.cmakein
and
https://github.com/analogdevicesinc/libiio/blob/master/bindings/python/CMakeLists.txt#L24

this should fail the build and stop it; but it doesn't;
i did not dig too deep into that;

i would fix this now if possible, otherwise the CI loses some usefulness;
we might get a failure, and won't notice it, and come back with fixes after people complain to us

CI/travis/make_linux
```diff
@@ -8,6 +8,16 @@ handle_default() {
...
 
+	# Do a install, to test things, and make sure that Cmake's
+	# package manager can find things on system path later
+	# then clean, so we can re-do things with the appropriate cmake
+	# flags, and sudo creates files as root, that make package can not
+	# write over...
+	cmake ../
+	make
+	sudo make install
+	rm * -rf
```
commodo: coming back to this;
i don't know if this is the best fix here;
it may be a good idea to try to pass an LD_LIBRARY_PATH or some other lib-path inside `bindings/python/setup.py.cmakein`
again, i'd ask @tfcollins to take a look and maybe consider a solution for this;

if we fix it here [in our build], others trying to build this manually might run into this and open issues;

if [however] this is the only possible fix, i would do something like:
```
if [ "$TRAVIS" = "true" ] ; then
	cmake ../
	make
	sudo make install
	rm * -rf
fi
``` 
this is also a bit of a fail-safe switch;

thing is: i run this script locally before doing patches [to these build-scripts]; saves some time during development; so the `sudo make install` installs things on the host, which i would avoid doing;
i also use some dockers, but sometimes it's quicker without them, since [currently] the build doesn't modify much on the host [other than the APT packages that it installs]

rgetz:
Ok - it looks like the env var "$TRAVIS" isn't set while running in the docker images...

I used `if [ "$TRAVIS" = "true" -o -n "$TRAVIS_BUILD_DIR" ] ;` that seems to work both places (inside and outside docker, but still in travis-ci

CI/travis/make_linux
rgetz: I did talk about this on the phone with @tfcollins ; and this is what we came to the conclusion was easier/best. The python parts need to get deployed automically to pypi on a release, and (in theory) they should now.

It does seem to build fine, it's just alot of annoying output. It also tests that "make install" works properly, and should let us build the examples in CI as well. (as a build test for more things).

tfcollins: I really have no idea why CPack doesn't error out. However, I would not package the bindings into the deb/rpm/tar.

Can cpack ignore files during packaging?

tfcollins:
> Ok - it looks like the env var "$TRAVIS" isn't set while running in the docker images...
> 
> I used `if [ "$TRAVIS" = "true" -o -n "$TRAVIS_BUILD_DIR" ] ;` that seems to work both places (inside and outside docker, but still in travis-ci

Variables can be passed at the run command: https://github.com/analogdevicesinc/libiio/blob/6f7ae64f74747b00dcc280a39950cc3c8b1982be/CI/travis/lib.sh#L274 -e "TRAVIS=$TRAVIS"

CI/travis/make_linux
rgetz: I think the issue is the way that the python bindings get added to the CODE

and we should be doing something like this in the install (splitting things)
```
install(TARGET app DESTINATION ... COMPONENT applications)
install(TARGET library DESTINATION ... COMPONENT libraries)
install(FILES <headers> DESTINATION ... COMPONENT headers)
```

and then in the `cmake/LinuxPackaging.cmake`
```
set(CPACK_ARCHIVE_COMPONENT_INSTALL ON)
set(CPACK_COMPONENTS_ALL applications)
include(CPack)
```

see [StackOverflow](https://stackoverflow.com/questions/27619234/how-to-exclude-files-dirs-from-cpack-binary-using-cmake) and [wiki](https://gitlab.kitware.com/cmake/community/-/wikis/doc/cpack/Component-Install-With-CPack)

rgetz:
Ok - did it that way.

It's still a good idea to make sure that things can be installed - even if it is not needed for the python bindings.

commodo:
> > Ok - it looks like the env var "$TRAVIS" isn't set while running in the docker images...
> > I used `if [ "$TRAVIS" = "true" -o -n "$TRAVIS_BUILD_DIR" ] ;` that seems to work both places (inside and outside docker, but still in travis-ci
> 
> Variables can be passed at the run command:
> 
> https://github.com/analogdevicesinc/libiio/blob/6f7ae64f74747b00dcc280a39950cc3c8b1982be/CI/travis/lib.sh#L274
> -e "TRAVIS=$TRAVIS"

hmm; i think maybe i should standardize somehow the passing of some env-vars to docker images;
for Linux builds I had to do some scripting to make it work for some env-vars;
i tried passing all, but it didn't work [some env-vars have too much whitespace to parse easily];
then it occurred to me that i probably shouldn't anyway;

but maybe it may make sense to pass all `TRAVIS*` env-vars in a standardized fashion;

CI/travis/make_linux
```diff
@@ -21,15 +22,32 @@ handle_default() {
...
+	if [ "$TRAVIS" = "true" ] ; then
+		# make sure the *.so can be found during make package
+		echo "### Installing package as root"
+		sudo make install
+		# get rid of anything that will cause a problem
+		echo "### Deleting install_manifest.txt (created as root)"
+		sudo rm ./install_manifest.txt
+	else
+		echo "### fail"
+		printenv
```
commodo, line 41: i would be careful with `printenv`
hopefully no secrets get exposed with this;

CI/travis/lib.sh
```diff
@@ -278,7 +278,7 @@ run_docker_script() {
 	local DOCKER_IMAGE="$2"
 	local OS_TYPE="$3"
 	local MOUNTPOINT="${4:-docker_build_dir}"
-	sudo docker run --rm=true \
+	sudo docker run --rm=true -e "TRAVIS=${TRAVIS}" \
```
commodo, line 281: specifically for this env-var passing, i did PR: https://github.com/analogdevicesinc/libiio/pull/420

this is a bit more flexible;
once you need an env-var in a docker, it's opening the gateway to more env-vars;
especially if you want to move more stuff into dockers [when wanting newer or weirder] images;

CI/travis/make_linux
rgetz: yeah, I really only put that there for debugging - you are right - it should be removed.

rgetz:
> but maybe it may make sense to pass all `TRAVIS*` env-vars in a standardized fashion;

I will rebased this on master when #420 gets merged.

commodo:
> > but maybe it may make sense to pass all `TRAVIS*` env-vars in a standardized fashion;
> 
> I will rebased this on master when #420 gets merged.

done
this can be rebased now

rgetz:
Ok - should be updated - will check to make sure all the errors/warnings are gone after it builds.

rgetz:
Looks like all the builds are OK.

PR #417, opened 2020-03-31
ci,build: split trigger-next-build into it's own stage

commodo:
This change creates a new stage for triggering the next build in the
pipeline and removes the old python script that checks if this is the last
job in the build.

At the current time, I am not sure if the triggering will happen if the
previous stage succeeded or not. But even so, it should be ok, since this
shouldn't occur too often, and would only happen on master and release
branches (which should be stable; and if they aren't this should show).

Also, one main advantage of this approach is that we reduce access to
Github's API which has a rate-limit.


commodo:
build looks like this:
https://travis-ci.org/github/analogdevicesinc/libiio/builds/669204339

rgetz:
This says what you are doing - but not why.

Why?

commodo:
The aim is to reduce the usage of the github API.

The rate limit for anonymous access is pretty low: 60 accesses per hour.

This was used previously to check if we need to trigger the next build in the pipeline (libad9361-iio).
And if the rate limit was hit, it wouldn’t trigger.

Thing is, we need to trigger the next build after all jobs are done.

Otherwise we trigger the next job multiple times.

Yeah I know: I’m terrible at explaining things after certain hour of the day.

rgetz:
Thanks - that helps.

So, why not just add a encrypted github token? We will need to double this for libm2k as well - won't we?

commodo:
so, with this change we don't need to care about GIthub API [token, or rate-limit];
it removes it, in favor of just using Travis-CI API to trigger the next build;
now it just needs some Travis-CI API access; i added a token inside Travis-CI a few months ago [I forget how I did it];

but i [probably] should add a trigger to libm2k from libiio

PR #420, opened 2020-04-01
ci,build: add generic mechanism for passing some Travis-CI envs

commodo:
This mechanism has been roughly adapted from the way it's done for the
Linux build.
A file is used on the mount-point that is passed to the docker, and the
'inside_docker.sh' script will load it if found.

Also added 3 common/variables. Two of them are used in the Linux build.

### DIFF
--- a/CI/travis/inside_docker.sh
+++ b/CI/travis/inside_docker.sh
@@ -17,6 +17,10 @@ else
 	exit 1
 fi
 
+if [ -f "/$LIBNAME/inside-travis-ci-docker-env" ] ; then
+	. /$LIBNAME/inside-travis-ci-docker-env
+fi
+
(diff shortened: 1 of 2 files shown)


rgetz:
Looks good to me - are those the only env vars we need? There are alot of others that [Travis-CI sets](https://docs.travis-ci.com/user/environment-variables/#convenience-variables).

commodo:
> Looks good to me - are those the only env vars we need? There are alot of others that [Travis-CI sets](https://docs.travis-ci.com/user/environment-variables/#convenience-variables).

so far they are the only ones that i found are needed;
we can add more later; as we decide that more stuff is required;
in the linux build only 2 are needed
https://github.com/analogdevicesinc/linux/blob/master/ci/travis/run-build-docker.sh#L7
[the lib.sh script is re-used there too]

PR #421, opened 2020-04-02
ci,build: add CentOS 8 docker build 

commodo:
There's CentOS 6 & 7 already.
Add CentOS 8 to the list.

We only need to enable to PowerTools repo to get doxygen & avahi-devel,
since they've been moved to that repo in 8.


rgetz:
Looks like the builds exhibit this upstream [fedora bug](https://bugzilla.redhat.com/show_bug.cgi?id=1773148) Which is minor.

Also looks like github didn't get the notification from travis-ci, as the builds are complete, but still say in progress. Kicking them off again, to see if that fixes things...

commodo:
CentOS 7 also had some annoying issues with CMake that went away;
I didn't see the Fedora bug you mentioned; maybe it went away in the meantime; no idea;

PR #393, opened 2020-03-04
Add sphinx doc for python bindings

tfcollins:
Users have been asking more and more about these bindings and the "magic" calls we sometimes do in pyadio-iio. This doc should help.

Open question is really around hosting. Ideally this would go on github pages along side doxygen (somehow), but we can go to RTD in the short term.

Just don't want to forget about this feature.

Signed-off-by: Travis F. Collins <travis.collins@analog.com>

PR #422, opened 2020-04-03
python: added iio_readdev.py and iio_writedev.py

cristi-iacob:
Implemented the iio_readdev and iio_writedev modules with Python bindings. These modules can be used as examples for sending and receiving buffers.

### DIFF
--- a/bindings/python/examples/iio_readdev.py
+++ b/bindings/python/examples/iio_readdev.py
@@ -49,6 +49,9 @@ channels = None
 
 
 def read_arguments():
+    """
+    Method for reading the command line parameters and setting the corresponding variables.
+    """
     global arg_ip, arg_uri, scan_for_context, buffer_size, num_samples, timeout, device_name, channels
 
     args = parser.parse_args()
@@ -76,6 +79,20 @@ def read_arguments():
 
 
 def create_context(scan_for_context, arg_uri, arg_ip):
+    """
+    Method for creating the corresponding context.
+
+    parameters:
+        scan_for_context: type=bool
+            Scan for available contexts and if only one is available use it.
+        arg_uri: type=string
+            The URI on which the program should look for a Context.
+        arg_ip: type=string
+            The IP on which the program should look for a Network Context.
+
+    returns: type:iio.Context
+        The resulted context.
+    """
     ctx = None
 
     try:
@@ -113,6 +130,18 @@ signal.signal(signal.SIGINT, keyboard_interrupt_handler)
 
 
 def read_data(buffer, num_samples):
+    """
+    Method for reading data from the buffer.
+
+    parameters:
+        buffer: type=iio.Buffer
+            Current buffer.
+        num_samples: type=int
+            Number of samples to capture, 0 = infinite. Default is 0.
+
+    returns: type=None
+        Reads data from buffer.
+    """
     if buffer is None:
         sys.stderr.write('Unable to create buffer!\n')
         exit(1)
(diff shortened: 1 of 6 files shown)


rgetz:
Since this will be an example - can we add some sphinx comments, so when we merge #393 things show up?

bindings/python/examples/iio_writedev.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+signal.signal(signal.SIGINT, keyboard_interrupt_handler)
+
+
+def main():
+    raw_arguments = sys.argv[1:]
+
+    try:
+        arguments, values = getopt.getopt(raw_arguments, unixOptions, gnuOptions)
```
tfcollins, line 99: Why was argparse not used? It much much simpler and in the core library.

bindings/python/examples/iio_writedev.py
```diff
@@ -0,0 +1,221 @@
...
+    num_samples_set = num_samples > 0
+
+    while app_running:
+        bytes_to_read = buffer._length if not num_samples_set else min(buffer._length, num_samples * dev.sample_size)
+        write_len = bytes_to_read
+        data = []
+
+        while bytes_to_read > 0:
+            data = sys.stdin.buffer.read(bytes_to_read)
+            bytes_to_read -= len(data)
```
tfcollins, line 197: Won't this loop simply overwrite data when bytes_to_read is not 0 after the first loop? Therefore the first array is never passed to the write buffer.

bindings/python/examples/iio_writedev.py
```diff
@@ -0,0 +1,221 @@
...
+
+            if num_samples == 0 and write_len < buffer_size * dev.sample_size:
+                exit(0)
+
+        ret = buffer.write(bytearray(data))
+        buffer.push()
+
+        if ret == 0:
+            if app_running:
+                print('Unable to push buffer!')
```
tfcollins, line 213: Should this be an exception?

bindings/python/examples/iio_readdev.py
```diff
@@ -0,0 +1,197 @@
...
+        for channel in dev.channels:
+            channel.enabled = True
+    else:
+        for channel_idx in channels:
+            dev.channels[int(channel_idx)].enabled = True
+
+    buffer = iio.Buffer(dev, buffer_size)
+
+    if buffer is None:
+        print('Unable to create buffer!')
```
tfcollins, line 179: Should this be an exception?

rgetz:
Also - until we have it fixed, check out: https://app.codacy.com/manual/commodo/libiio/pullRequest?prid=5299649#newIssuesView

bindings/python/examples/iio_writedev.py
cristi-iacob: Yes, you are right.  I have modified; I think this modification should solve the problem.

bindings/python/examples/iio_writedev.py
cristi-iacob: I didn't know argparse was in the core library; getopt just seemed simpler at the first glance for me. I have changed them to use argparse.

rgetz:
There still doesn't seem to be any markdown in the code that will help sphinx? Should there be?

cristi-iacob:
I have added some documentation for the functions from iio_readdev.py and iio_writedev.py.
Also I have added the .rst files for them, in order to be compatible with sphinx. If these files shouldn't have been added, I will delete the last commit.

tfcollins:
Can you add references to the index file to the examples? Otherwise I don't think you can get to them.

cristi-iacob:
I have added a new component on the main page, Tools. It contains references to iio_readdev and iio_writedev documentation pages.

rgetz:
Thanks - looks great.

PR #423, opened 2020-04-03
coverity fixes

rgetz:
Fixes a few new coverity issues 

https://scan.coverity.com/projects/analogdevicesinc-libiio

More to come.

### DIFF
--- a/context.c
+++ b/context.c
@@ -20,6 +20,7 @@
 #include "iio-config.h"
 #include "iio-private.h"
 #include "sort.h"
+#include "network.h"
 
 #include <errno.h>
 #include <string.h>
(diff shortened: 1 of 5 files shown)


commodo:
ummm; so it looks like coverity did not fire for these during review/build
is there a build issue with it that we should fix first?

rgetz:
the free version of coverity only allows one or two reviews per day, so we can't do them on pull requests. We only do them on master.

context.c
```diff
@@ -311,19 +312,18 @@ struct iio_context * iio_create_default_context(void)
...
 	if (hostname) {
-		struct iio_context *ctx;
+		if (strlen(hostname) > 2 && strlen(hostname) < MAXHOSTNAMELEN + 4) {
+			struct iio_context *ctx;
 
-		ctx = iio_create_context_from_uri(hostname);
-		if (ctx)
-			return ctx;
-
-#ifdef WITH_NETWORK_BACKEND
```
commodo, line 320: i'm having trouble following this;
is it ok to remove `#ifdef WITH_NETWORK_BACKEND` ?
is `#ifdef HAVE_DNS_SD` dependent on `#ifdef WITH_NETWORK_BACKEND`?

commodo:
> the free version of coverity only allows one or two reviews per day, so we can't do them on pull requests. We only do them on master.

ack

commodo:
only 1 comment;
rest looks good

context.c
rgetz: yes - `HAVE_DNS_SD` can only be set if `WITH_NETWORK_BACKEND` is set, and it is more appropriate for this code path.

rgetz: `HAVE_DNS_SD` is what allows empty strings to return as IP numbers - without that - calls to the networking will just fail with empty strings.

rgetz: [CMakeLists.txt](https://github.com/analogdevicesinc/libiio/blob/master/CMakeLists.txt#L235-L286)

`HAVE_DNS_SD` is set inside the `if(WITH_NETWORK_BACKEND)`

rgetz:
Just as FYI, this did clean up 7 issues.

image

PR #426, opened 2020-04-07
coverity fixes2

rgetz:
fix the remaining outstanding/recently added coverity issues, and one that has been there for awhile.

### DIFF
--- a/iiod/iiod.c
+++ b/iiod/iiod.c
@@ -407,9 +407,24 @@ static int main_interactive(struct iio_context *ctx, bool verbose, bool use_aio)
 
 	if (!use_aio) {
 		flags = fcntl(STDIN_FILENO, F_GETFL);
-		fcntl(STDIN_FILENO, F_SETFL, flags | O_NONBLOCK);
+		if (flags >= 0)
+			flags = fcntl(STDIN_FILENO, F_SETFL, flags | O_NONBLOCK);
+		if (flags < 0) {
+			char err_str[1024];
+			iio_strerror(errno, err_str, sizeof(err_str));
+			IIO_ERROR("Could not get/set O_NONBLOCK on STDIN_FILENO"
+					" %s (%d)\n", err_str, -errno);
+		}
+
 		flags = fcntl(STDOUT_FILENO, F_GETFL);
-		fcntl(STDOUT_FILENO, F_SETFL, flags | O_NONBLOCK);
+		if (flags >= 0)
+			flags = fcntl(STDOUT_FILENO, F_SETFL, flags | O_NONBLOCK);
+		if (flags < 0) {
+			char err_str[1024];
+			iio_strerror(errno, err_str, sizeof(err_str));
+			IIO_ERROR("Could not get/set O_NONBLOCK on STDOUT_FILENO"
+					" %s (%d)\n", err_str, -errno);
+		}
 	}
 
 	interpreter(ctx, STDIN_FILENO, STDOUT_FILENO, verbose,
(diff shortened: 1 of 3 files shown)


tests/iio_stresstest.c
```diff
@@ -385,15 +385,18 @@ int main(int argc, char **argv)
 			break;
 		case 'b':
 			info.arg_index += 2;
-			info.buffer_size = atoi(info.argv[info.arg_index]);
+			info.buffer_size = strtol(info.argv[info.arg_index], NULL, 10);
+			/* Max 4M */
+			if (info.buffer_size > (1024 * 1024 * 4))
+				info.buffer_size = 1024 * 1024 * 4;
```
commodo, line 391: this max looks like slightly new behavior;
but it's fine to leave it here

iiod/iiod.c
```diff
@@ -407,9 +407,24 @@ static int main_interactive(struct iio_context *ctx, bool verbose, bool use_aio)
 
 	if (!use_aio) {
 		flags = fcntl(STDIN_FILENO, F_GETFL);
-		fcntl(STDIN_FILENO, F_SETFL, flags | O_NONBLOCK);
+		if (flags > 0)
```
commodo, line 410: is `flags >= 0`  also valid? i.e. `flags` being also zero;

iiod/iiod.c
```diff
@@ -407,9 +407,24 @@ static int main_interactive(struct iio_context *ctx, bool verbose, bool use_aio)
...
+			char err_str[1024];
+			iio_strerror(errno, err_str, sizeof(err_str));
+			ERROR("Could not get/set O_NONBLOCK on STDIN_FILENO"
+					" %s (%d)\n", err_str, -errno);
+		}
+
 		flags = fcntl(STDOUT_FILENO, F_GETFL);
-		fcntl(STDOUT_FILENO, F_SETFL, flags | O_NONBLOCK);
+		if (flags > 0)
+			flags = fcntl(STDOUT_FILENO, F_SETFL, flags | O_NONBLOCK);
```
commodo, line 421: same question here about `flags >= 0`
thing is: depending on the case, `flags >= 0` can change behavior or introduce a new bug

iiod/iiod.c
```diff
@@ -407,9 +407,24 @@ static int main_interactive(struct iio_context *ctx, bool verbose, bool use_aio)
 
 	if (!use_aio) {
 		flags = fcntl(STDIN_FILENO, F_GETFL);
-		fcntl(STDIN_FILENO, F_SETFL, flags | O_NONBLOCK);
+		if (flags > 0)
+			flags = fcntl(STDIN_FILENO, F_SETFL, flags | O_NONBLOCK);
+		if (flags < 0) {
+			char err_str[1024];
+			iio_strerror(errno, err_str, sizeof(err_str));
+			ERROR("Could not get/set O_NONBLOCK on STDIN_FILENO"
```
dNechita, line 415: ERROR needs to be updated to: IIO_ERROR. Maybe rebase this onto master.

tests/iio_stresstest.c
rgetz: yes, it was complaining that the value was tainted (not checked), so check it....

iiod/iiod.c
rgetz: I'm not sure if that is practical for STDIN (it should be open as write), but in theory, yes - it's not an error, so I updated. Good catch. Thanks

iiod/iiod.c
rgetz: fixed / Thanks

iiod/iiod.c
rgetz: yeah, rebased on master.

rgetz:
Fixed 2.

image

PR #428, opened 2020-04-07
add libini

rgetz:
This will compete with #405, and only one should be committed to master.

I prefer this one. It's easy, simple and will be maintained.


local.c
```diff
@@ -19,6 +19,10 @@
 #include "debug.h"
 #include "iio-private.h"
 #include "sort.h"
+#ifdef WITH_LOCAL_CONFIG
+#include "libini/ini.h"
```
commodo, line 23: does clang-format have a preference for `#include "headers"` to be added after `#include <headers>`?
maybe it's a slightly better idea to move all this block after this;

but we can leave it as is for now;

pcercuei:
If you don't want a submodule, have a look at [git-subrepo](https://github.com/ingydotnet/git-subrepo).

dNechita:
A local copy seems the simplest way to me.

rgetz:
Ok I will merge this by end of today unless anyone has objections.

PR #429, opened 2020-04-08
coverity: fix Untrusted value as argument in iio_stresstest

rgetz:
make sure to apply limits to all input from user arguments.

I think this is the last coverity issue from recent commits.

### DIFF
--- a/tests/iio_stresstest.c
+++ b/tests/iio_stresstest.c
@@ -50,6 +50,48 @@ static int getNumCores(void) {
 #endif
 }
 
+
+/* Code snippet insired by Nick Strupat
+ * https://stackoverflow.com/questions/794632/programmatically-get-the-cache-line-size
+ * released under the "Feel free to do whatever you want with it." license.
+ */
+static size_t cache_line_size(void)
+{
+	size_t cacheline = 0;
+
+#ifdef _WIN32
+	DWORD buffer_size = 0;
+	DWORD i = 0;
+	SYSTEM_LOGICAL_PROCESSOR_INFORMATION * buffer = 0;
+
+	GetLogicalProcessorInformation(0, &buffer_size);
+	buffer = (SYSTEM_LOGICAL_PROCESSOR_INFORMATION *)malloc(buffer_size);
+	GetLogicalProcessorInformation(&buffer[0], &buffer_size);
+
+	for (i = 0; i != buffer_size / sizeof(SYSTEM_LOGICAL_PROCESSOR_INFORMATION); ++i) {
+		if (buffer[i].Relationship == RelationCache && buffer[i].Cache.Level == 1) {
+			cacheline = buffer[i].Cache.LineSize;
+			break;
+		}
+	}
+	free(buffer);
+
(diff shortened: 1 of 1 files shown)


tests/iio_stresstest.c
```diff
@@ -387,16 +387,29 @@ int main(int argc, char **argv)
...
+				fprintf(stderr, "Limiting timeout to once per day\n");
+				info.timeout = 60 * 60 * 24;
+			}
+			info.timeout = info.timeout * 1000;
 			break;
 		case 'T':
 			info.arg_index +=2;
 			info.num_threads = strtol(info.argv[info.arg_index], NULL, 10);
+			/* Max number threads 1024 */
+			if (info.num_threads > 1024) {
```
larsclausen, line 409: num_threads also needs to be >= 1

rgetz: Thanks - fixed.

rgetz:
> num_threads is signed. I think you have to check for >= 0.

Yes - you are right. I changed num_threads to unsigned. and made the strtol go into a signed tmp that is checked for negative now.

tests/iio_stresstest.c
```diff
@@ -50,6 +50,49 @@ static int getNumCores(void) {
...
+	FILE * p = 0;
+
+	p = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
+	if (p) {
+		fscanf(p, "%zu", &cacheline);
+		fclose(p);
+	}
+#endif
+	return cacheline;
+
```
commodo, line 93: nitpick: extra line

tests/iio_stresstest.c
```diff
@@ -385,18 +434,45 @@ int main(int argc, char **argv)
...
+			if (tmp < min_samples) {
+				fprintf(stderr, "Setting buffersize to %zu bytes\n", min_samples);
+				tmp = min_samples;
+			}
+			info.buffer_size = (unsigned int)tmp;
 			break;
 		case 't':
 			info.arg_index +=2;
-			info.timeout = 1000 * strtol(info.argv[info.arg_index], NULL, 10);
+			tmp = strtol(info.argv[info.arg_index], NULL, 10);
```
commodo, line 451: should we use `strtoul()` for all these?
doesn't look like we have values that should be negative yet;

then the `tmp < 0` check goes away;

also, this code block is repeated a few times;
it could be an idea to implement it as a function/routine called  `clamp()`
maybe something like:
```
int clamp(const char *name, val, min, max) {
       if (val > max) {
               val = max;
               fprtinf("Clamped %s to max %u", name, max)
       } 
       if (val < min) {
              val = min;
              fprtinf("Clamped %s to min %u", name, min)
        }
       return val;
}
```

tests/iio_stresstest.c
rgetz: fixed.

tests/iio_stresstest.c
rgetz: fixed

PR #431, opened 2020-04-09
iio_writedev: fix off by one when pushing buffers bug

cristi-iacob:
The bug was described in this pull request:
https://github.com/analogdevicesinc/libiio/pull/410

As @commodo suggested, the old solution was a little confusing for future development because it forced the while loop to do an extra iteration in order to find out that all requested samples were sent.
The modifications from this pull request avoid this thing to happen. Instead of calling quit_all(), which also closes the buffer, after reading from stdin the requested number of samples, the program now only sets app_running and exit_code. Therefore the program will now push even the last buffer, which was not pushed in the previous version, and afterwards will exit the while loop.
If the cyclic mode is enabled, app_running should not be set on false in order to enter the while loop which ensures the cyclic behaviour.
Also, the `if (app_running)` statement from the push-error-handling block was hiding some errors (because app_running was set on false by quit_all()).

### DIFF
--- a/tests/iio_writedev.c
+++ b/tests/iio_writedev.c
@@ -472,11 +472,9 @@ int main(int argc, char **argv)
 
 		int ret = iio_buffer_push(buffer);
 		if (ret < 0) {
-			if (app_running) {
-				char buf[256];
-				iio_strerror(-ret, buf, sizeof(buf));
-				fprintf(stderr, "Unable to push buffer: %s\n", buf);
-			}
+			char buf[256];
+			iio_strerror(-ret, buf, sizeof(buf));
+			fprintf(stderr, "Unable to push buffer: %s\n", buf);
 			break;
 		}
 
--- a/tests/iio_writedev.c
+++ b/tests/iio_writedev.c
@@ -91,7 +91,7 @@ static void quit_all(int sig)
 {
 	exit_code = sig;
 	app_running = false;
-	if (buffer)
+	if (buffer && exit_code != EXIT_SUCCESS)
 		iio_buffer_cancel(buffer);
 }
 
@@ -457,8 +457,9 @@ int main(int argc, char **argv)
 
 			if (num_samples) {
 				num_samples -= write_len / sample_size;
-				if (!num_samples)
+				if (!num_samples && !cyclic_buffer) {
 					quit_all(EXIT_SUCCESS);
+				}
 			}
 		} else {
 			ret = iio_buffer_foreach_sample(buffer, read_sample, NULL);


commodo:
let's wait a bit and see if there are other comments on this

tests/iio_writedev.c
```diff
@@ -457,8 +457,11 @@ int main(int argc, char **argv)
 
 			if (num_samples) {
 				num_samples -= write_len / sample_size;
-				if (!num_samples)
-					quit_all(EXIT_SUCCESS);
+				if (!num_samples) {
+					exit_code = EXIT_SUCCESS;
```
rgetz, line 461: Isn't `exit_code = EXIT_SUCCESS;` set by default on L88? Isn't it always this?

tests/iio_writedev.c
```diff
@@ -457,8 +457,11 @@ int main(int argc, char **argv)
...
+					if (!cyclic_buffer)
```
rgetz, line 462: If it is, then you should be able to do `if(!num_samples && !cyclic_buffer)` correct?

rgetz:
and thanks - this looks alot less complicated that the previous version.

tests/iio_writedev.c
commodo: there's various smarter ways to do this;
this was one that is a simple-looking one;

another one would be `app_running = cyclic_buffer`
either works;
i don't care much which we use as long as it's simple looking enough, so that nobody asks anything about it;
the combination of 2 bool vars can look slightly confusing in following code-path

commodo:
> Isn't the part of problem that we are doing a
> 
> ```
> 	if (buffer)
> 		iio_buffer_cancel(buffer);
> ```
> 
> in `quit_all`, and on a normal exit path (when `sig == EXIT_SUCCESS`) we shouldn't be doing that?

yep; that's the problem;
and that's also a potential fix, and i had thought about that, but i don't know of reproducing cases for other places where `quit_all(EXIT_SUCCESS)`

there's also this function:
```
static ssize_t read_sample(const struct iio_channel *chn,
                void *buf, size_t len, void *d)
{
        size_t nb = fread(buf, 1, len, stdin);
        if (num_samples != 0) {
                num_samples--;
                if (num_samples == 0) {
                        quit_all(EXIT_SUCCESS);
                        return -1;
                }
        }
        return (ssize_t) nb;
}
```

`return -1` either indicates an error? or is just some return we should not consider really important? why ` quit_all(EXIT_SUCCESS);`?  [1]
so, i'd like a reproducing case, or something that tries to run through that code to understand it better;

similarly, there's
```
static void * sig_handler_thd(void *data)
{
        sigset_t *mask = data;
        int ret, sig;

        /* Blocks until one of the termination signals is received */
        do {
                ret = sigwait(mask, &sig);
        } while (ret == EINTR);

        quit_all(ret);

        return NULL;
}
```

what are the potential values for `ret = sigwait(mask, &sig);`?
and then again, similar questions as [1]

naturally, if possible, i'd like some input from @pcercuei ?
or anyone else?
or at least until we get some test-coverage over the other code-paths, don't know

rgetz:
It was @mhennerich who wrote this - not paul.

commodo:
> It was @mhennerich who wrote this - not paul.

oh ok;
i admit i was lazy and did not go any git archeology

commodo:
will merge this tomorrow the latest if no objections;

PR #436, opened 2020-04-13
coverity: iio_stresstest : fix Untrusted value as argument (TAINTED_SCALAR)

rgetz:
According to a few places I found, the passing an argc to strtoul
without checking it's length first, will cause this error to be flagged,
so check the length.

Hopefully this finally resolves things.

### DIFF
--- a/tests/iio_stresstest.c
+++ b/tests/iio_stresstest.c
@@ -35,6 +35,13 @@
 #define SAMPLES_PER_READ 256
 #define NUM_TIMESTAMPS (16*1024)
 
+#ifdef _MSC_BUILD
+#define inline __inline
+#define iio_snprintf sprintf_s
+#else
+#define iio_snprintf snprintf
+#endif
+
 static int getNumCores(void) {
 #ifdef _WIN32
 	SYSTEM_INFO sysinfo;
@@ -82,11 +89,14 @@ static size_t cache_line_size(void)
 
 #elif __linux__
 	FILE * p = 0;
+	int ret;
 
 	p = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
 	if (p) {
-		fscanf(p, "%zu", &cacheline);
+		ret = fscanf(p, "%zu", &cacheline);
 		fclose(p);
+		if (ret != 1)
+			cacheline = 0;
 	}
 #endif
 	return cacheline;
@@ -388,16 +398,28 @@ thread_fail:
 	return (void *)EXIT_FAILURE;
 }
 
-static unsigned long int clamp(const char *name, unsigned long int val,
+static unsigned long int sanitize_clamp(const char *name, const char *argv,
 		unsigned long int min, unsigned long int max)
 {
+
+	unsigned long int val;
+	char buf[20];
+
+	if (!argv) {
+		val = 0;
+	} else {
+		/* sanitized buffer by taking first 20 (or less) char */
+		iio_snprintf(buf, sizeof(buf), "%s", argv);
+		val = strtoul(buf, NULL, 10);
+	}
+
 	if (val > max) {
 		val = max;
-		fprintf(stderr, "Clamped %s to max %lu", name, max);
+		fprintf(stderr, "Clamped %s to max %lu\n", name, max);
 	}
 	if (val < min) {
 		val = min;
-		fprintf(stderr, "Clamped %s to min %lu", name, min);
+		fprintf(stderr, "Clamped %s to min %lu\n", name, min);
 	}
 	return val;
 }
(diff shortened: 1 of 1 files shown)


tests/iio_stresstest.c
```diff
@@ -82,11 +89,14 @@ static size_t cache_line_size(void)
...
 #elif __linux__
 	FILE * p = 0;
+	int ret;
 
 	p = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
 	if (p) {
-		fscanf(p, "%zu", &cacheline);
+		ret = fscanf(p, "%zu", &cacheline);
 		fclose(p);
+		if (!ret || ret == EOF)
```
commodo, line 98: maybe coverity would complain that negative values are checked?
no idea;

in any case, maybe let's go for the simplest check for `ret = fscanf()`; i.e. `if (ret != 1) cacheline = 0`;
we know that the function should return exactly 1 scanned element;
enforcing anything-but 1 scanned elem returned should cover all other returns;

tests/iio_stresstest.c
```diff
@@ -388,18 +398,37 @@ static void *client_thread(void *data)
 	return (void *)EXIT_FAILURE;
 }
 
-static unsigned long int clamp(const char *name, unsigned long int val,
+static unsigned long int sanitize_clamp(const char *name, const char *val,
```
commodo, line 401: if we rename `const char *val` -> `const char *argv`, we can rename `tmp` -> `val` as a local variable;
that would reduce some bits of patch in the clamp() part;

tests/iio_stresstest.c
```diff
@@ -388,18 +398,37 @@ static void *client_thread(void *data)
...
 		unsigned long int min, unsigned long int max)
 {
-	if (val > max) {
-		val = max;
-		fprintf(stderr, "Clamped %s to max %lu", name, max);
+
+	unsigned long int tmp;
+	char buf[20];
+
+	if(!val) {
```
commodo, line 408: nitpick: `if__space__() {`

tests/iio_stresstest.c
```diff
@@ -388,18 +398,37 @@ static void *client_thread(void *data)
...
 }
 
-static unsigned long int clamp(const char *name, unsigned long int val,
+static unsigned long int sanitize_clamp(const char *name, const char *val,
 		unsigned long int min, unsigned long int max)
 {
-	if (val > max) {
-		val = max;
-		fprintf(stderr, "Clamped %s to max %lu", name, max);
+
```
commodo, line 404: nitpick: extra line

tests/iio_stresstest.c
```diff
@@ -388,18 +398,37 @@ static void *client_thread(void *data)
...
+	if(!val) {
+		tmp = 0;
+	} else {
+
+		/* find the length of chars of max, and make sure the string
+		 * is shorter than that
+		 */
+		iio_snprintf(buf, sizeof(buf), "%lu", max);
+		if (strnlen(val, strlen(buf) + 1) <= strlen(buf)) /* Flawfinder: ignore */
```
commodo, line 416: umm, i feel this is where pleasing coverity starts to reduce the value of the code [in terms of readability, maintainability,  understanding the intent of the code, etc];
[ to-be-fair: i also [sometimes] fall easily into the trap of pleasing static-checkers ; ]

using a buffer for argv[idx]  + strlen()/strlnlen() doesn't seem to be more efficient than simply using strtol()
if there is no null terminator, things will go sour anyway;
i'd expect strltoul() to use strlen() or something equivalent anyway;

now, there could be some better way to please coverity and still make this readable;
but if there is a flag/comment for coverity to ignore this, it's probably not a bad idea [either];

[fwiw: the idea of parsing argv[] in this function is nice though];

tests/iio_stresstest.c
rgetz: fixed

tests/iio_stresstest.c
rgetz: fixed

tests/iio_stresstest.c
rgetz: done

tests/iio_stresstest.c
rgetz: updated

tests/iio_stresstest.c
rgetz: Updated with a simpler solution.

tests/iio_stresstest.c
```diff
@@ -388,16 +398,28 @@ static void *client_thread(void *data)
...
 {
+
+	unsigned long int val;
+	char buf[20];
+
+	if (!argv) {
+		val = 0;
+	} else {
+		/* sanitized buffer by taking first 20 (or less) char */
+		iio_snprintf(buf, sizeof(buf), "%s",argv);
```
commodo, line 412: nitpick: `iio_snprintf(buf, sizeof(buf), "%s",__space__argv);`

rgetz: fixed - sorry - need to get the formatter installed...

rgetz:
Finally!

image

PR #433, opened 2020-04-09
add man to html doc

rgetz:
Let me know if anyone has any questions.

PR #437, opened 2020-04-15
cmake: man: Avoid installing empty directories

marv:
I noticed that empty `CMakeFiles` directories are installed into the man page directories:
```
$ DESTDIR=/tmp/libiio/ ninja -j1 install
[snip]
-- Installing: /tmp/libiio/usr/share/man/man3
-- Installing: /tmp/libiio/usr/share/man/man3/libiio.3
**-- Installing: /tmp/libiio/usr/share/man/man3/CMakeFiles**
-- Installing: /tmp/libiio/usr/share/man/man1
-- Installing: /tmp/libiio/usr/share/man/man1/iio_attr.1
-- Installing: /tmp/libiio/usr/share/man/man1/iio_writedev.1
**-- Installing: /tmp/libiio/usr/share/man/man1/CMakeFiles**
-- Installing: /tmp/libiio/usr/share/man/man1/iio_readdev.1
-- Installing: /tmp/libiio/usr/share/man/man1/iio_info.1
-- Installing: /tmp/libiio/usr/share/man/man1/iio_reg.1
[snip]
```
First patch is a non-functional change to simplify the cmake file by using `foreach`, the second fixes the problem by using a non-cmake directory in which man-pages are created

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -414,7 +414,7 @@ if(WITH_DOC)
 		COMMENT "Generating C# documentation with Doxygen" VERBATIM)
 
 	if(NOT SKIP_INSTALL_ALL)
-		install(DIRECTORY ${HTML_DEST_DIR}
+		install(DIRECTORY ${CMAKE_HTML_DEST_DIR}
 			DESTINATION ${CMAKE_INSTALL_DOCDIR})
 	endif()
 endif()
--- a/man/CMakeLists.txt
+++ b/man/CMakeLists.txt
@@ -9,7 +9,7 @@ if (WITH_MAN)
 		)
 	configure_file(
 		${CMAKE_BINARY_DIR}/libiio.3.in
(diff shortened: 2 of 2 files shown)


rgetz:
Thanks - I think this might clash with PR# #433 ; so I will add that first, and then rebase this one.

rgetz:
@marv : I think you can drop the first commit, and rebase the 2nd on master?

marv:
@rgetz Yep, the first patch was obsoleted by the recent changes - I rebased the other one. I noticed that the HTML docs weren't installed for me with `-DWITH_DOC=ON` because `HTML_DEST_DIR` was used instead of `CMAKE_HTML_DEST_DIR` (I'd recommend prefixing own variables with something like `LIBIIO_` by the way - `CMAKE_` is taken by cmake ;-))

rgetz:
Thanks for update and fix.

marv:
No problem, I'm glad to help :-)

PR #438, opened 2020-04-15
Codacy (shellcheck): Double quote to prevent globbing and word splitt…

rgetz:
…ing.

Many places in the CI infrastructure, we do things like:

echo $1

rather than:

echo "$1"

bash will actually:
  - Split the first argument by IFS (spaces, tabs and line feeds).
  - Expand each of them as if it was a glob.
  - Join all the resulting strings and filenames with spaces.
  - then finally print the result.

Quoting variables prevents word splitting and glob expansion, and prevents
scripts from breaking when input contains spaces.

So do that.

### DIFF
--- a/CI/travis/check_kernel.sh
+++ b/CI/travis/check_kernel.sh
@@ -50,7 +50,7 @@ while IFS="" read -r p ; do
 	key=$(echo ${p//[[:space:],]/})
 	count=$(grep "\[$key\]" /tmp/libiio_iio_chan_type_name_spec | wc -l)
 	if [ "$count" -eq "0" ] ; then
-		echo $key missing from channel.c iio_chan_type_name_spec
+		echo "$key missing from channel.c iio_chan_type_name_spec"
 		ret=1
 	fi
 done < /tmp/libiio_iio_chan_type
@@ -59,10 +59,10 @@ echo
 sed -i '/IIO_NO_MOD/d' /tmp/libiio_iio_modifier
 
 while IFS="" read -r p ; do
-	key=$(echo ${p//[[:space:],]/})
+	key=$(echo "${p//[[:space:],]/}")
 	count=$(grep "\[$key\]" /tmp/libiio_modifier_names | wc -l)
 	if [ "$count" -eq "0" ] ; then
-		echo $key missing from channel.c modifier_names
+		echo "$key missing from channel.c modifier_names"
 		ret=1
 	fi
 done < /tmp/libiio_iio_modifier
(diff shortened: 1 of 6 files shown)


rgetz:
https://app.codacy.com/gh/analogdevicesinc/libiio/pullRequest?prid=5365203 reports:

image

Which does't means it's good - I tried all the functions I could. the CI will test most of the rest...

rgetz:
fixed some minor issues - as far as I can tell - things are working.

CI/travis/check_kernel.sh
```diff
@@ -50,7 +50,7 @@ while IFS="" read -r p ; do
 	key=$(echo ${p//[[:space:],]/})
 	count=$(grep "\[$key\]" /tmp/libiio_iio_chan_type_name_spec | wc -l)
 	if [ "$count" -eq "0" ] ; then
-		echo $key missing from channel.c iio_chan_type_name_spec
+		echo "$key" missing from channel.c iio_chan_type_name_spec
```
commodo, line 53: i going this route, maybe quote then entire string?

CI/travis/check_kernel.sh
```diff
@@ -59,10 +59,10 @@ echo
 sed -i '/IIO_NO_MOD/d' /tmp/libiio_iio_modifier
 
 while IFS="" read -r p ; do
-	key=$(echo ${p//[[:space:],]/})
+	key=$(echo "${p//[[:space:],]/}")
```
commodo, line 62: my shell-fu is a bit vague;
especially with bash;
question is: does this still evaluate the same here after adding quotes?

fwiw/generally: i also tried a while back to do this shellcheck fixing of double-quotes, but quite often i would need some vars to evaluate without the quotes there;

the semantic with/without double-quotes is a bit too subtle sometimes, and may not always be worth the risk of fixing
that is: unless we re-test everything

CI/travis/lib.sh
```diff
@@ -155,9 +155,9 @@ get_ldist() {
 }
 
 __brew_install_or_upgrade() {
-	brew install $1 || \
-		brew upgrade $1 || \
-		brew ls --versions $1
+	brew install "$1" || \
```
commodo, line 158: IIRC: I did this on purpose [not adding quotes];
but it may be fine to add it;
not sure

CI/travis/check_kernel.sh
rgetz: I tried this by hand - and yes - it evaluates to the same.

CI/travis/lib.sh
rgetz: It doesn't change anything - and does protect things...

CI/travis/check_kernel.sh
rgetz: fixed

rgetz:
I did this to start - since these were the major issues that were causing:
image

rgetz:
This was part of #441

rgetz:
This got these up to an A or B rating.

image

PR #440, opened 2020-04-17
tests: put common/replicated functions into a single file

rgetz:
The scan function was copied/pasted in each utility, this moves
it to a a seperate c file, which is built as a library, which then
every utility statically links to, saving maintainance, and build time.


rgetz:
9 files changed, 199 insertions(+), 344 deletions(-)

At least a step in the right direction...

rgetz:
This does help with #441, as a lot of the issues in ./tests/* are duplicated code. (first box is "issues", 2nd is "duplicated code").

image

This pull request results:
image

but I don't know what -26 really means...

rgetz:
Ack on the clang-format, was trying to resolve some of the issues first. Working on trying to resolve some of the issues in clang-format that were pointed out.

PR #442, opened 2020-04-20
add examples to cmake & check on CI

rgetz:
A a small compile only Cmake for the examples directory, and then turn it on in CI (Linux only for now).


rgetz:
If everyone would put their include files in the same places - things like this would be much easier...

rgetz:
trying to track down a centos 8 issue.

CI/travis/before_install_linux
```diff
@@ -19,21 +19,21 @@ handle_centos() {
 	yum info epel-release
 	yum -y install epel-release
 
+	if is_centos_at_least_ver "8" ; then
```
commodo, line 22: this doesn't work before `redhat-lsb-core` is installed;
it requires `lsb_release` installed;

though; maybe it would be a better idea to rework those distro functions (i.e. `get_codename()`, `get_dist_id()` & `get_version()`) to just manually parse `/etc/os-release`;
no idea;
i liked using `lsb_release` since the output format is simple/unique, and doesn't need extra parsing, but given the hassle of installing it on bare-bone docker images, it may not be worth it;

commodo:
> trying to track down a centos 8 issue.

looks like cdk-devel is not part of any official CentOS/Fedora repo;
i searched a few repos;
maybe for a first iteration we could drop it for CentOS 8, and leave a comment about it;
maybe cdk-devel will get added later;
i remember with CentOS 7 [in the beginning] cmake was pretty bad and i had to waste time/effort to hack-patch cmake in the docker image until they fixed it;

maybe cdk-devel is something that may get re-added later?
no idea; but i feel adding now is a bigger/hackier effort on our part than is worth doing;
and we can end up with a worse result temporarily;
it's cheaper to just open a ticket here https://bugs.centos.org/main_page.php and see

CI/travis/before_install_linux
rgetz: Yeah, it was a last try before I went to bed last night... I was hoping it would make some of the logic easier - I will flip things back.

rgetz:
Thanks for having a look.

I will just turn it off for now, It was building on centos 6/7 and on all the debian releases, which should be all we need for CI testing - as long as it is getting compiled in a few places.

rgetz:
OK - this is ready to be reviewed. Thanks

rgetz:
Ok - formatted new Cmake file with cmake-format.

rgetz:
Ok - now we are testing, we can see where the issues are (before, the examples dir was 0 lines) ....

image

Issue #441, opened 2020-04-18
need to fix/resolve flawfinder issues


rgetz:
Codacy reports:

![image](https://user-images.githubusercontent.com/1017869/79643550-7d32f300-8171-11ea-98cf-9aeaf1ec5f71.png)

We should work on this, and get it to a reasonable number before the next release. 

Most of these are potential issues, not real issues (things like don't use strlen, because its bad), but there are lots of times it's OK in this code base, since we know all the strings are null terminated. However - if someone wanted to inject something into the library - it possible, and then the library would crash, so it still worth looking at...

rgetz:
Today was the first day we got to a rating of "A":

![image](https://user-images.githubusercontent.com/1017869/80146609-1c942380-8580-11ea-9dbe-fb0f13031bc0.png)

I'm mainly focusing on the C issues (C# issues should get resolved as part of #364) tagged as error or warning.
![image](https://user-images.githubusercontent.com/1017869/80146776-5e24ce80-8580-11ea-9d9d-ef921f310256.png) ![image](https://user-images.githubusercontent.com/1017869/80146868-8b717c80-8580-11ea-8891-aadce153cbec.png)

rgetz:
All the major issues were resolved.

![image](https://user-images.githubusercontent.com/1017869/83821671-fee8cc80-a69c-11ea-99eb-d8c211181aad.png)

We still have some C warnings, but no errors, (one which needs to be dismissed as a false positive).
![image](https://user-images.githubusercontent.com/1017869/83821732-2a6bb700-a69d-11ea-8bc4-cc8f49762072.png)

We still have alot of work to do on the C# bindings, but I think I will close this for now.

PR #448, opened 2020-04-22
flawfinder : sprintf doesn't check for buffer overflows.

rgetz:
https://cwe.mitre.org/data/definitions/120.html points out that
sprintf will overflow buffers, so don't use it.

We move to the internal iio_snprintf, and move a few remaining snprintf
to the same to handle the necessary windows/Linux differences.

### DIFF
--- a/dns_sd.c
+++ b/dns_sd.c
@@ -71,7 +71,7 @@ static int dnssd_fill_context_info(struct iio_context_info *info,
 		char *hostname, char *addr_str, int port)
 {
 	struct iio_context *ctx;
-	char uri[MAXHOSTNAMELEN + 3];
+	char uri[sizeof("ip:") + MAXHOSTNAMELEN + sizeof (":65535") + 1];
 	char description[255], *p;
 	const char *hw_model, *serial;
 	int i;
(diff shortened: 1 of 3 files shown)


dns_sd.c
```diff
@@ -71,7 +71,7 @@ static int dnssd_fill_context_info(struct iio_context_info *info,
 		char *hostname, char *addr_str, int port)
 {
 	struct iio_context *ctx;
-	char uri[MAXHOSTNAMELEN + 3];
+	char uri[sizeof("ip:") + MAXHOSTNAMELEN + sizeof (":65536") + 1];
```
dNechita, line 74: doesn't make a difference, but technically it's 65535

iiod/ops.c
```diff
@@ -428,14 +428,22 @@ static ssize_t send_data(struct DevEntry *dev, struct ThdEntry *thd, size_t len)
...
+			iio_snprintf(ptr, len, "%08x", mask[i - 1]);
+			len -= 8;
+		}
 
 		*ptr = '\n';
+		len--;
+
+		if (len < 0)
+			IIO_ERROR("send_data: string length error\n");
+
```
dNechita, line 446: I would return in case len < 0 but not sure which error code is appropriate.

dns_sd.c
rgetz: fixed

iiod/ops.c
rgetz: I added a `return -ENOSPC` (No space left on device)

rgetz:
This is part of #441; and fixes 6 issues:

image

PR #453, opened 2020-04-23
Cmake/Coverity: Force coverity to check _everything_

rgetz:
The current Cmake logic is set so that if things (like avahi) are not
installed, it doesn't build or link against it. On normal systems this
make sense.

I recently noticed that coverity builds (which happen on the coverity
servers, not on the Travis-CI servers), don't have external libraries
installed, and then a few options don't get configured/added, and
therefore, don't get checked.

Add an option to the Cmake which blindly adds things (which is fine,
since coverity doesn't actually compile/link things); so all backends
and options get checked.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,6 +352,34 @@ if (WIN32)
 	configure_file(properties.rc.cmakein ${LIBIIO_RC} @ONLY)
 endif()
 
+option(COVERITY_BUILD "Turn on things for Coverity Builds" OFF)
+if (COVERITY_BUILD)
+	# Make sure everything gets turned on
+	list(APPEND LIBIIO_CFILES local.c
+		./libini/libini.c
+		usb.c
+		serial.c
+		network.c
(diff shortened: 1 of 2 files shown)


rgetz:
So, github was having issues this am, and I think the appveyor <-> github messaging missed each other - it built fine.

CMakeLists.txt
```diff
@@ -352,6 +352,34 @@ if (WIN32)
 	configure_file(properties.rc.cmakein ${LIBIIO_RC} @ONLY)
 endif()
 
+option(COVERITY_BUILD "Turn on things for Coverity Builds" OFF)
```
dNechita, line 355: Shouldn't we document this on https://github.com/analogdevicesinc/libiio/blob/master/README_BUILD.md?

rgetz: No. Normal users should not use this flag. It will generate a source tree which may not build.

dNechita: Fair point.

PR #454, opened 2020-04-23
fix examples

rgetz:
Now that we added some things to the examples, fix the errors that were pointed out.

### DIFF
--- a/examples/adrv9009-iiostream.c
+++ b/examples/adrv9009-iiostream.c
@@ -36,7 +36,7 @@
 #define MHZ(x) ((long long)(x*1000000.0 + .5))
 #define GHZ(x) ((long long)(x*1000000000.0 + .5))
 
-#define ASSERT(expr) { \
+#define IIO_ASSERT(expr) { \
 	if (!(expr)) { \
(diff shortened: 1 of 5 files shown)


examples/dummy-iiostream.c
```diff
@@ -368,6 +368,10 @@ int main (int argc, char **argv)
 				size_t sample_size = fmt->length / 8 * repeat;
 
 				buf = malloc(sample_size * buffer_length);
+				if (!buf) {
+					perror("out of memory, malloc failed\n");
```
dNechita, line 372: perror() will append the string corresponding to the errno which more or less says the same thing. Maybe mention the function where this fails or "CHANNEL_READ".

examples/adrv9009-iiostream.c
```diff
@@ -36,7 +36,7 @@
 #define MHZ(x) ((long long)(x*1000000.0 + .5))
 #define GHZ(x) ((long long)(x*1000000000.0 + .5))
 
-#define ASSERT(expr) { \
+#define iio_ASSERT(expr) { \
```
dNechita, line 39: It would be more consistent to use uppercase for all macros.

rgetz: done

PR #459, opened 2020-04-24
xml buildups: fix bug introduced by recent changes

rgetz:
be more pedanic about the differences between strlen (doesn't include
null char) and sizeof (which does include null char).

Now, we calculate the exact length the the string (number of char)
add one for malloc, fill everything out, and then check it make sure
the length is one (the null char) left.

tested on host and local (Zedboard + FMCOMMS2)


rgetz:
Thanks - and sorry. When I broke the one large commit up, I missed some things, and I never tested on a remote context. I will be more careful going forward.

PR #461, opened 2020-04-24
fix remaining snprintf issues

rgetz:
this resolves/removes all the sprintf and strcpy, except in the examples, and tests.


rgetz:
This is part of #441

PR #462, opened 2020-04-24
cmake: support old Cmake for -std=c99

rgetz:
on old verisons of Cmake, they don't support the way we are setting the
"C_STANDARD 99" (in set_target_properties), so check for old verison and
then set the CMAKE_C_FLAGS by hand.

This eliminates all the warnings that happen on the Zynq images we use
(2019R1) and some of the older CI images we support, and make it alot
easier to see where the real errors/warnings are.


rgetz:
With this patchset, we can include -Werror if we wanted (turn all warnings into errors) - which would help with automated checking...

I think I would only do this for the Travis-CI, (during development, and when people check things out, a warning would still be a warning...)

Thoughts?

mhennerich:
> With this patchset, we can include -Werror if we wanted (turn all warnings into errors) - which would help with automated checking...
> 
> I think I would only do this for the Travis-CI, (only when TRAVIS or CI is set, so during development, and when people check things out, a warning would still be a warning...)
> 
> Thoughts?

Werror should only be used during development or CI/CT. If we leave things and someone uses a new compiler things won't build...

dNechita:
I think it is ok to treat warnings as errors only during development (including CI/CT).

examples/CMakeLists.txt
```diff
@@ -56,13 +56,24 @@ if (PTHREAD_LIBRARIES
...
-    )
-    set(IIO_TESTS_TARGETS ${IIO_TESTS_TARGETS} iio-monitor)
+
+    set(TEMP ${CMAKE_REQUIRED_LIBRARIES})
+    set(TEMP1 ${CMAKE_REQUIRED_INCLUDES})
+    list(APPEND CMAKE_REQUIRED_LIBRARIES ${CURSES_LIBRARY} ${CDK_LIBRARY})
+    list(APPEND CMAKE_REQUIRED_INCLUDES ${LIBCKD_INCLUDE_DIR})
+    check_symbol_exists(CDK_CSTRING2 "cdk.h" HAS_CDK_CSTRING2)
+    set(CMAKE_REQUIRED_LIBRARIES ${TEMP})
+    set(CMAKE_REQUIRED_INCLUDES ${TEMP1})
```
dNechita, line 66: So, there's a possibility that cdk.h exists but at the same time not include symbol CDK_CSTRING2?
If yes, then cmake should print a message with the reason why iio-monitor will not be included in the build.

rgetz: On old versions of CDK - it doesn't do it. I can add a message there.

rgetz:
> Werror should only be used during development or CI/CT.

agreed - it only gets turned on if BOTH the following environmental vars are set:
 - `CI=true`
 - `TRAVIS=true`

The [CI](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) already sets these, and developers can set those pretty easy, but normal people will not.

PR #463, opened 2020-04-27
Coverity: 'IIO_ASSERT' falsely flags just like 'ASSERT' does. Change it.

rgetz:
Since we are ensuring (to secure or guarantee:) something happens -
change the ASSERT macro to that name, so all the tools (Coverity?) that look for
ASSERTs will not flag this by accident.

Changes IIO_ASSERT to IIO_ENSURE


rgetz:
This fixed some issues:

image

PR #464, opened 2020-04-27
Coverity: Remove remaining asserts that were not caught

rgetz:
dummy-iiostream was still using the standard assert, so remove that and
replace it with the default macro that is being used in the other files.


rgetz:
This did build properly on appveyor, just github <-> appveyor logic isn't working...

PR #465, opened 2020-04-27
fix iio monitor

rgetz:
minor fixes to iio_monitor that coverity and codacy pointed out.

- replace sprintf
- fix double free


rgetz:
This is part of #441

PR #466, opened 2020-04-27
Coverity: Fix potential divide by zero in iio_readdev

rgetz:
in iio_readdev:

206    size_t sample_size;
...
354    sample_size = iio_device_get_sample_size(dev);
...
404    num_samples -= read_len / sample_size;

There are two issues:

iio_device_get_sample_size can return negative error codes,
and we would miss these. Fix that by making it a signed number.

this could potentially cause an error if sample_size is zero.
Catch that fault and error out. While we are here, check
for negative error codes, and also error out.

Reported by Coverity.


rgetz:
fixed

image

PR #467, opened 2020-04-27
csharp: Control Structures should use curly braces {}

rgetz:
According to Microsoft, and Sonar C#, all Control structures (do, for,
foreach, while, if) should use curly braces. While not technically incorrect,
the omission of curly braces can be misleading, and may lead to the
introduction of errors during maintenance.

See CERT, EXP19-C Use braces for the body of an if, for, or while statement
https://www.securecoding.cert.org/confluence/x/1QGMAg

See CERT, EXP52-J Use braces for the body of an if, for, or while statement
https://www.securecoding.cert.org/confluence/x/3wHEAw

### DIFF
--- a/bindings/csharp/Attr.cs
+++ b/bindings/csharp/Attr.cs
@@ -78,9 +78,13 @@ namespace iio
         public void write(bool val)
         {
             if (val)
+            {
                 write("1");
+            }
             else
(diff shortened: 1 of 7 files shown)


rgetz:
This is part of #441

PR #468, opened 2020-04-27
Dont ignore sign differences

rgetz:
With all the coverity and codacy updates, there are very few signed/unsigned issues anymore - so fix those up, and remove the ignore flag.

This will need to be re-based on master after #466 is merged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,8 +59,7 @@ if (NOT LOG_LEVEL)
 endif()
 
 if (MSVC)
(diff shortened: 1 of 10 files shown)


rgetz:
I'm not 100% this is necessary, or required - but was on a long conference call this afternoon...

dNechita:
This seems useful.

rgetz:
rebasing on master (now that we have -Werror turning on everywhere)

PR #475, opened 2020-04-30
cmake: Restructure CMake CFLAGS for and ensure they are set

rgetz:
In the past Clang was not set on x86, and now we also print a warning if
we don't know what the compiler is.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,20 +53,19 @@ if (WITH_TESTS)
 	set(NEED_THREADS 1)
 endif()
 
+if (NOT LOG_LEVEL)
+	set(LOG_LEVEL Info CACHE STRING "Log level" FORCE)
+	set_property(CACHE LOG_LEVEL PROPERTY STRINGS NoLog Error Warning Info Debug)
+endif()
+
 if (MSVC)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -88,15 +87,21 @@ if (CMAKE_COMPILER_IS_GNUCC)
...
 		set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
 	endif()
-endif()
 
-if(APPLE)
-	#full Single Unix Standard v3 (SUSv3) conformance (the Unix API)
-	add_definitions(-D_DARWIN_C_SOURCE)
+elseif (CMAKE_C_COMPILER_ID STREQUAL "Clang")
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -pedantic -Wno-unused-parameter -Wno-sign-compare")
+	if(APPLE)
```
dNechita, line 93: If someone tries to use gcc on Apple, this won't work. Not sure if people do that. Maybe it's ok this way.

rgetz: OK will fix.

rgetz:
rebased to master to pick up the appveour fixes.

PR #476, opened 2020-04-30
appveyor.yml: Upgrade packages and pacman before trying to install any dependencies.

AlexandraTrifan:
Otherwise it might fail because pacman can't unzip .zst files or because local packages are outdated.


rgetz:
you mean like it has been doing most of yesterday? :)

Will approve & merge as soon as the appveyour build is complete.

PR #477, opened 2020-05-01
MATLAB bindings cleanup

tfcollins:
PR Removed old mfiles which are depreciated, moves the scan_block functions into iio.h, removes dependent code blocks, and cleans up the CMake. This should simplify the code a bit and prevent users from trying the old MATLAB bindings moving forward.


iio.h
```diff
@@ -253,6 +254,42 @@ __api __check_ret __pure const char * iio_context_info_get_uri(
 		const struct iio_context_info *info);
 
 
+/** @brief Create a scan block
+ * @param backend A NULL-terminated string containing the backend to use for
+ * scanning. If NULL, all the available backends are used.
+ * @param flags Unused for now. Set to 0.
+ * @return on success, a pointer to a iio_scan_block structure
+ * @return On failure, NULL is returned and errno is set appropriately */
+__api struct iio_scan_block * iio_create_scan_block(
```
rgetz, line 263: Can we add a 
```
Introduced in version 0.20. */
```

in the comment for all?

tfcollins: added

rgetz:
Thanks

bindings/matlab/iio-wrapper.h
```diff
@@ -1,46 +1,4 @@
...
-/** @brief Get the iio_context_info for a particular context
-* @param blk A pointer to an iio_scan_block structure
-* @param index The index corresponding to the context.
-* @return A pointer to the iio_context_info for the context
-* @returns On success, a pointer to the specified iio_context_info
-* @returns On failure, NULL is returned and errno is set appropriately
-*/
-__api struct iio_context_info *iio_scan_block_get_info(
-		struct iio_scan_block *blk, unsigned int index);
-
```
dNechita, line 45: Is this header still useful, now that is empty?

rgetz:
can we rebase on master? That might find some additional issues now we have warnings and Werror turned on on the CI?

tfcollins:
@rgetz Rebased.

@dNechita I also asked MathWorks if the header is still necessary.

bindings/matlab/iio-wrapper.h
rgetz: yes - this is removed now.

rgetz:
The ARM builds on CI are failing, (due to CI issues), this is working fine.

Issue #364, opened 2020-02-05
review CSharp bindings


rgetz:
This is a list of functions listed as __api in iio.h, and the number of times they show up in the C# bindings.

```
0 iio_buffer_cancel
0 iio_buffer_first
0 iio_buffer_foreach_sample
0 iio_buffer_get_data
0 iio_buffer_get_device
0 iio_buffer_get_poll_fd
0 iio_buffer_set_blocking_mode
0 iio_buffer_set_data
0 iio_buffer_step
0 iio_channel_attr_read_all
0 iio_channel_attr_read_bool
0 iio_channel_attr_read_double
0 iio_channel_attr_read_longlong
0 iio_channel_attr_write_all
0 iio_channel_attr_write_bool
0 iio_channel_attr_write_double
0 iio_channel_attr_write_longlong
0 iio_channel_attr_write_raw
0 iio_channel_convert
0 iio_channel_convert_inverse
0 iio_channel_find_attr
0 iio_channel_get_device
0 iio_channel_get_index
0 iio_channel_get_modifier
0 iio_channel_get_type
0 iio_channel_set_data
0 iio_context_find_device
0 iio_context_get_attr
0 iio_context_get_attrs_count
0 iio_context_get_attr_value
0 iio_context_info_get_description
0 iio_context_info_get_uri
0 iio_context_info_list_free
0 iio_create_local_context
0 iio_create_scan_context
0 iio_create_xml_context
0 iio_create_xml_context_mem
0 iio_device_attr_read_all
0 iio_device_attr_read_bool
0 iio_device_attr_read_double
0 iio_device_attr_read_longlong
0 iio_device_attr_write_all
0 iio_device_attr_write_bool
0 iio_device_attr_write_double
0 iio_device_attr_write_longlong
0 iio_device_attr_write_raw
0 iio_device_buffer_attr_read
0 iio_device_buffer_attr_read_all
0 iio_device_buffer_attr_read_bool
0 iio_device_buffer_attr_read_double
0 iio_device_buffer_attr_read_longlong
0 iio_device_buffer_attr_write
0 iio_device_buffer_attr_write_all
0 iio_device_buffer_attr_write_bool
0 iio_device_buffer_attr_write_double
0 iio_device_buffer_attr_write_longlong
0 iio_device_buffer_attr_write_raw
0 iio_device_debug_attr_read_all
0 iio_device_debug_attr_read_bool
0 iio_device_debug_attr_read_double
[...]

PR #488, opened 2020-05-11
csharp: added missing bindings.

cristi-iacob:
I have added the missing C# bindings, as discussed in #364.

I have not added bindings for set_data() and get_data() for the Device, Channel and Buffer classes, because they are used only for internal purposes.

iio_create_local_context, iio_create_xml_context, iio_create_xml_context_mem are not needed, because the iio_create_context_from_uri, which is used in the Context constructor, calls all these methods internally. 

Signed-off-by: Cristian Iacob <cristian.iacob@analog.com>


cristi-iacob:
I have also added the bindings for scan context, but they are currently having a bug. I will add a commit for them after I solve that issue.

bindings/csharp/Context.cs
```diff
@@ -135,7 +164,37 @@ private static IntPtr getContextFromString(string str)
             return ptr;
         }
 
-        private Context(IntPtr ctx)
+        private static string getBackendFromUri(string uri)
+        {
```
rgetz, line 168: getContextFromString() -> returns an instance of the iio.Context class ; that's not what this does.

If you need this - you should use the C API - iio_context_get_name(ctx)

bindings/csharp/Channel.cs
```diff
@@ -69,8 +69,99 @@ public override void write(string str)
             }
         }
 
+        /// <summary><see cref="iio.Channel.ChannelModifier"/> class:
+        /// Contains the available channel modifiers.</summary>
+        public enum ChannelModifier
+        {
+            IIO_NO_MOD,
```
rgetz, line 76: Is there a different way to include the C enum? (so we don't need to maintain the list in multiple places) and more important - they are all the same values in the codebase... If someone uses IIO_VOLTAGE, on the C library, and that gets assigned 1, it better be "1" in the C# version as well...

dNechita: A solution to not depend on the values of the enum elements is to create a conversion function between the C type and the C# type (like a lookup table) and of course, the inverse function as well.

bindings/csharp/Context.cs
```diff
@@ -39,11 +39,19 @@ internal Version(uint major, uint minor, string git_tag)
         }
     }
 
+    private class ContextAttr : Attr
+    {
+        private IntPtr ctx;
+
+
```
dNechita, line 46: What's with all these spaces?

bindings/csharp/Device.cs
```diff
@@ -101,7 +102,43 @@ public override void write(string str)
...
+            private static extern int iio_device_buffer_attr_write(IntPtr dev, [In] string name, [In()] string val);
+
+            public DeviceBufferAttr(IntPtr dev, string name) : base(name)
+            {
+                this.dev = dev;
+            }
+
+            public override string read()
+            {
+                StringBuilder builder = new StringBuilder(1024);
```
dNechita, line 122: Why 1024?

bindings/csharp/Context.cs
cristi-iacob: getContextFromString() is returning a pointer to the recently created Context. This pointer is afterwards used in the Context(IntPtr ctx) constructor, to create a new instance of the iio.Context  class.

bindings/csharp/Device.cs
cristi-iacob: It was the same as in the DeviceDebugAttr.read() method, which was already implemented. Should I change these values?

dNechita: Some attributes might be longer than this. Have a look at how this is handled in iio_info.

rgetz: I think we determined the min should be a page size (4096).

cristi-iacob: I have changed the size to the value of BUF_SIZE (16384).

bindings/csharp/Channel.cs
cristi-iacob: There is a problem with including the C enum in the C# code. In C the enums are not allocated at an address. They are just interpreted as constants at compile time. This makes them inaccessible in C# via the libiio.dll.
I have discussed with @dNechita and there seem to be two possible solutions for this:
1. Keep a copy of the C enum in the C# code and update it manually every time the C enum in modified. Optionally this process can be automatized with some tools, like Swig.
2. The enums should be placed in a file and CMake should generate their corresponding C header file. If the WITH_CSHARP_BINDINGS flag is set, it should also generate a corresponding C# file containing the enums.

rgetz: Ok - poking around in the [C# doc](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/enums) _An enum declaration that does not explicitly declare an underlying type has an underlying type of int._ That is what I was missing. It might be better to declare things as `public enum ChannelModifier : int`

It might be just easier to modify the existing `CI/travis/check_kernel.sh` (which checks the difference between the latest kernel and the libiio iio.h files) to also check the python and C# bindings.

I think that would resolve things for me - and might be the fastest/easiest.

rgetz:
Also need to check/review the codacy results:
https://app.codacy.com/gh/analogdevicesinc/libiio/pullRequest?prid=5503379

Some are common to alot of other code in the bindings, but we should be looking at fixing at least some of the things - like the private issues?

cristi-iacob:
I have solved the private issues.

bindings/csharp/Context.cs
rgetz: I'm still missing things - if you want the strings - why not just call iio_context_get_name?

rgetz:
Ok - so this is the current status of this branch (there I think you said you were working on):
- iio_create_scan_context missing
- iio_scan_context_destroy missing
- iio_scan_context_get_info_list missing
- iio_context_info_list_free missing
- iio_context_info_get_uri missing

Did you want to wait and fix/add those before this is merged?


I think these were just added, and might not be necessary?
- iio_create_scan_block missing
- iio_scan_block_destroy missing
- iio_scan_block_scan missing
- iio_scan_block_get_info missing

I think these are not necessary:
- iio_create_xml_context missing
- iio_create_xml_context_mem missing

This is missing:
- iio_context_get_attr_value missing
- iio_buffer_foreach_sample missing

I think these are built in?
- iio_device_set_data missing
- iio_device_get_data missing
- iio_channel_set_data missing
- iio_buffer_set_data missing
- iio_buffer_get_data missing

Thanks

bindings/csharp/Context.cs
cristi-iacob: I have deleted this method and the backend property from the Context class. I didn't know that backend and name are equivalent for Context. Thanks for the suggestion!

cristi-iacob:
I have added the documentation for the buffer_attrs property.
I have also solved some of the new Codacy issues.

About the bindings from this branch:

- iio_create_scan_context missing
- iio_scan_context_destroy missing
- iio_scan_context_get_info_list missing
- iio_context_info_list_free missing
- iio_context_info_get_uri missing
- iio_create_scan_block missing
- iio_scan_block_destroy missing
- iio_scan_block_scan missing
- iio_scan_block_get_info missing

I haven't managed to get access to the context_info structure, but I think that the only new information this one brings is the uri from the scan_context. 

- iio_create_xml_context missing
- iio_create_xml_context_mem missing
- iio_context_get_attr_value missing
- iio_buffer_foreach_sample missing

These are not necessary.

- iio_device_set_data missing
- iio_device_get_data missing
- iio_channel_set_data missing
- iio_buffer_set_data missing
- iio_buffer_get_data missing

Yes, these are built-in.

cristi-iacob:
I have added the missing bindings for context scanning.

PR #489, opened 2020-05-11
tests: add a common function to parse and create iio_contexts

rgetz:
Some (older) implementations of getopt_long) do not protect against internal
buffer overflows, so copy argv to a temp str array, and parse most in a
common function. (-u -h -x -n, and -S).
  -h, --help Show this help and quit.
  -x, --xml  Use the XML backend with the provided XML file.
  -u, --uri  Use the context at the provided URI.
  -S, --scan Scan for available backends.
  -a, --auto Scan for available contexts and if only one is available use it.

The "default" for scan is "S", since "s" is used for many apps as sample
size. This means that iio_info now takes -s and -S as scan.

This also fixes both CWE-120 and CWE-20
https://cwe.mitre.org/data/definitions/120.html
https://cwe.mitre.org/data/definitions/20.html
and almost is a net decrease (348 insertions, 345 deletions over 9
files).

### DIFF
--- a/tests/iio_common.c
+++ b/tests/iio_common.c
@@ -21,6 +21,7 @@
 
 #include <iio.h>
 #include <stdio.h>
+#include <errno.h>
 #include <inttypes.h>
 #include <getopt.h>
 #include <string.h>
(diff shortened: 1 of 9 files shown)


rgetz:
I think the cpp check is a false positive, and the travis-ci did build properly - this is Ok to merge when someone has a chance to review it.

tests/iio_common.c
```diff
@@ -149,6 +150,143 @@ unsigned long int sanitize_clamp(const char *name, const char *argv,
...
+			goto err_dup;
+	}
+
+	return new_argv;
+
+err_dup:
+	i--;
+	for (; i >= 0; i--)
+		free(new_argv[i]);
+
```
dNechita, line 170: Shouldn't `new_argv` be freed as well?

rgetz: yes - fixed.

PR #492, opened 2020-05-13
cppcheck: fix null ptr dereference

rgetz:
Not sure why, but it looks like cppcheck doesn't run on pull requests...

When adding : 336c3c8b884d83049b589ecd8b1482076d40e6e6
I made a mistake, and called xmalloc() with a NULL name, which is
printed when malloc fails. So:
 - fix the calling functions to include a name
 - add a check to make sure name isn't touched when if it is NULL

This should resolve things.

### DIFF
--- a/tests/iio_adi_xflow_check.c
+++ b/tests/iio_adi_xflow_check.c
@@ -166,7 +166,7 @@ int main(int argc, char **argv)
 	char unit;
 	int ret;
 
-	argw = dup_argv(argc, argv);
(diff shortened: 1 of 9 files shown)


dNechita:
Normally argv[0] contains the executable name and MY_NAME definition could be removed. But this is not guaranteed to be always true.
If you think argv[0] isn't reliable enough, then all these changes make sense.

rgetz:
MY_NAME was added by Paul/Lars and I just kept it.

It's a constant, incase the application is re-named.

PR #495, opened 2020-05-14
html: tweak html output to keep track of versions easier.

rgetz:
Add git version to main page. This requires:
  - moving doc/html/index.html -> doc/index.html.in
  - adding cmake vars to new doc/index.html.in
  - tweaking the css, so it prints out on two lines properly
  - processing it in the cmake file

Don't include date in footers, which helps with tracking
changes (don't see doc changes just because you create new doc).
This requires a custom footer, so add a footer template, and use it in
the existing Doxyfile.in files.


rgetz:
Updated html is at : https://rgetz.github.io/libiio/v0.19/index.html

PR #498, opened 2020-05-15
serial: support defining number of stop bits in serial URI

rgetz:
in the past the serial backend was limited to a single stop bit, when
most hardware can support one or two. This allows URIs to define the
number of stop bits in a standard notation (8n1). This means serial uri
are now defined as:

serial:port,baud,config
serial:/dev/ttyUSB0,115200,8n1

This does break backwards compatibility with the previous serial URI
which was:

serial:/dev/ttyUSB0,115200n8

Since this is outside the library, and there are few serial users, and
8n1 is the "standard" way of doing things - it's better to change/update
now.

Update the doc (which mistakenly said we supported this).


dbogdan:
The changes look safe, at least from the libtinyiiod point of view.

PR #499, opened 2020-05-15
python: python examples rewrite.

cristi-iacob:
Rewrote iio_readdev.py and iio_writedev.py in a more pythonic fashion.

### DIFF
--- a/bindings/python/examples/iio_writedev.py
+++ b/bindings/python/examples/iio_writedev.py
@@ -16,208 +16,218 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
 import time
-
-import iio
 import sys
-import signal
 import argparse
-
-parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


bindings/python/examples/iio_writedev.py
```diff
@@ -16,208 +16,218 @@
...
-def main():
-    read_arguments()
+    def write(self):
+        """Push data into the buffer."""
+        app_running = True
+        num_samples = self.arguments.num_samples
 
-    ctx = create_context(scan_for_context, arg_uri, arg_ip)
+        while app_running:
+            bytes_to_read = self.buffer._length if num_samples == 0 \
```
tfcollins, line 199: use len(self.buffer) instead

bindings/python/examples/iio_writedev.py
```diff
@@ -16,208 +16,218 @@
...
+                else min(self.buffer._length, num_samples * self.device.sample_size)
```
tfcollins, line 200: use len(self.buffer) instead

rgetz:
#501 should fix some of the issues codacy issues. It just need to be approved and committed before this is pushed.

rgetz:
This was done as part of #503

Issue #379, opened 2020-02-17
Input/output error (-5) when writing write only attributes using iio_attr


mbancisor:
dump_device_attributes() tries to do a read after each write (https://github.com/analogdevicesinc/libiio/blob/master/tests/iio_attr.c#L207). This fails for the write only attributes, returning:
ERROR: Input/output error (-5)

rgetz:
On the device:
```
--w-------    1 root     root          4096 Jan  1 00:00 ./iio:device1/multichip_sync
```

On a host:
```
iio_attr -a -d ad9361-phy  multichip_sync 0
Using auto-detected IIO context at URI "usb:3.8.5"
dev 'ad9361-phy', attr 'multichip_sync', value :ERROR: Permission denied (-13)
wrote 2 bytes to multichip_sync
dev 'ad9361-phy', attr 'multichip_sync', value :ERROR: Permission denied (-13)
```

That means the read didn't work, but the write did.

What exact device/attribute are you trying?

mhennerich:
ad9361-phy, multichip_sync doesn't implement a read method.

rgetz:
Agree - that is expected. A attribute which is write only - read will fail.

I'm saying that's OK.

dNechita:
Wouldn't it make sense to add an additional flag for cases where only a write is intended.
e.g.
`iio_attr -a -d ad9361-phy -w  multichip_sync 0`
, where -w means write-only.

rgetz:
I'm not sure. The only way that I'm aware to make sure the write worked as expected (or to see what value is actually written to an attribute), is to do a read afterwards.

That is what -q (quiet) does - is don't do the first read, don't print the status of the write, and only do the second one, for that specific purpose (to see what is actually there).

I don't see the harm that the error string provides.

dNechita:
@mbanciso is this behavior causing any inconveniences for you setup?

PR #500, opened 2020-05-17
iio_attr: add error checking

rgetz:
When helping debug someone's shell scripts what used iio_attr on
Friday, we noticed that iio_attr just plain eats too many things,
which makes it hard to debug things when using -q (quiet).

So, make sure to print things when errors occur. Errors are now defined
as attempting to read/write a device/channel/buffer/attribute that
doesn't exist. or when a write error occurs.

Also exit with an exit error code, so that if you set -e in a shell
script, it quits right away.

### DIFF
--- a/tests/iio_attr.c
+++ b/tests/iio_attr.c
@@ -93,10 +93,10 @@ eek:
 	return ret;
 }
 
-static void dump_device_attributes(const struct iio_device *dev,
+static int dump_device_attributes(const struct iio_device *dev,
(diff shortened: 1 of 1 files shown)


rgetz:
This also might help with the issues that were seen in #379 

If a write error occurs - we no longer do a read, and throw an failure error code.
If a write is successful, and a read error occurs afterwards (a write only attribute), we still exit with success.

so `set -e` in most shell scripts should halt on accessing an attribute that doesn't exist, or when there is an error with that attribute.

PR #502, opened 2020-05-18
Flawfinder : remove atol and atof, replace with strtoll and strtof

rgetz:
Flawfinder reminds us that unless checked, the resulting number can exceed
the expected range (CWE-190). If source untrusted, check both minimum and
maximum, even if the input had no minus sign (large numbers can roll over
into negative number; consider saving to an unsigned value if that is
intended).
https://cwe.mitre.org/data/definitions/190.html

replace these library calls with strtoll and strtof, and include more
error checking.

### DIFF
--- a/local.c
+++ b/local.c
@@ -1221,9 +1221,16 @@ static int handle_protected_scan_element_attr(struct iio_channel *chn,
 
 	if (!strcmp(name, "index")) {
 		ret = local_read_dev_attr(dev, path, buf, sizeof(buf), false);
-		if (ret > 0)
-			chn->index = atol(buf);
+		if (ret > 0) {
+			char *end;
+			long long value;
 
+			value = strtoll(buf, &end, 0);
+			if (end == buf || value < 0 || value > LONG_MAX)
+				return -EINVAL;
+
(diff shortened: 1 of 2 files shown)


local.c
```diff
@@ -1221,9 +1221,16 @@ static int handle_protected_scan_element_attr(struct iio_channel *chn,
...
 	if (!strcmp(name, "index")) {
 		ret = local_read_dev_attr(dev, path, buf, sizeof(buf), false);
-		if (ret > 0)
-			chn->index = atol(buf);
+		if (ret > 0) {
+			char *end;
+			long long value;
+
+			value = strtoll(buf, &end, 0);
+			if (end == buf || value > LONG_MAX)
```
dNechita, line 1229: Should we check if 'value' is negative? Since the value is assigned to chn->index and indexes aren't negative. Or they could be?

rgetz: No, they should always be equal to or greater than zero. I can add that before committing.

rgetz: should be fixed now.

rgetz: fixed

PR #503, opened 2020-05-18
rework python examples

rgetz:
Since @cristi-iacob is on vacation this week, take #499 and add a few things to fix all the issues.


rgetz:
image

PR #506, opened 2020-05-19
pylint & pep257: Resolve issues in python coding style.

rgetz:
This resolves most pylint and pep8 and pep257 issues pointed out via
prospector pylint and should be formatted according to black.

prospector --profile .prospector.yml ./bindings/python/examples/iio_info.py
pylint --rcfile=.pylintrc ./bindings/python/examples/iio_info.py
black ./bindings/python/examples/iio_info.py

There are still some issues with the way that globals are handled, but I
think it is OK for now, and makes the generated doc look better as
well.

### DIFF
--- a/.prospector.yml
+++ b/.prospector.yml
@@ -16,15 +16,22 @@ pyroma:
 mypy:
   run: false
 
+mccabe:
+  run: false
+
 pylint:
   run: false
 
 pydocroma:
   run: true
 
+pyflakes:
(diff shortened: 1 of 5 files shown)


rgetz:
image

and then:
image

which _are_ fixed...

rgetz:
The ubuntu <-> travis-ci was broken earlier today, - this does pass all the checks.

PR #508, opened 2020-05-19
Support for DNS-SD on Windows

adisuciu:
This PR adds support for DNS self discovery on Windows using the mdns library - https://github.com/mjansson/mdns


dns_sd_windows.c
```diff
@@ -0,0 +1,262 @@
+// Based on the example provided here: https://github.com/mjansson/mdns
```
rgetz, line 1: We should say which specific file, who was the author, and what license it has (the public domain), to make sure that we can include it in a LGPL project.

dns_sd_windows.c
```diff
@@ -0,0 +1,262 @@
...
+
+#ifdef _WIN32
```
rgetz, line 3: If the CMake is defining things as only Windows, can we remove some of these #ifdef _WIN32? This file will only be compiled on Windows - it just makes is harder to follow.

dns_sd_windows.c
```diff
@@ -0,0 +1,262 @@
...
+#  define _CRT_SECURE_NO_WARNINGS 1
```
rgetz, line 4: We specifically do not do this anymore. Files which are Windows, compile with the secure strings settings

dns_sd_windows.c
```diff
@@ -0,0 +1,262 @@
...
+	} while (num_retries-- > 0);
+
+	if (!adapter_address || (ret != NO_ERROR)) {
+		free(adapter_address);
+		IIO_ERROR("Failed to get network adapter addresses\n");
+		return num_sockets;
+	}
+
+	/*int first_ipv4 = 1;
+	int first_ipv6 = 1;*/
```
rgetz, line 70: Can remove settings that aren't used, or ifdef them based on `HAVE_IPV6` (which is in iio-config.h)

dns_sd_windows.c
```diff
@@ -0,0 +1,262 @@
...
+	WSADATA wsaData;
+	if (WSAStartup(versionWanted, &wsaData)) {
+		printf("Failed to initialize WinSock\n");
+		return -1;
+	}
+
+	char hostname_buffer[128];
+	DWORD hostname_size = (DWORD)sizeof(hostname_buffer);
+	if (GetComputerNameA(hostname_buffer, &hostname_size))
+		hostname = hostname_buffer;
```
rgetz, line 198: `hostname` is set, but never used?

dns_sd_windows.c
```diff
@@ -0,0 +1,262 @@
...
+		MDNS_STRING_FORMAT(srv.name), srv.priority, srv.weight, srv.port);
+
+	// Go to the last element in the list
+	while (dd->next != NULL)
+		dd = dd->next;
+
+	if (srv.name.length > 1)
+	{
+		dd->hostname = malloc(srv.name.length);
+		strncpy(dd->hostname, srv.name.str, srv.name.length);
```
rgetz, line 168: we use iio_strlcpy, rather than strncpy. Need to make sure that srv.name.length includes the null char, or you will need to `+ 1` to it.

dns_sd_windows.c
```diff
@@ -0,0 +1,262 @@
...
+		dd->hostname[srv.name.length - 1] = 0;
+	}
+	strcpy(dd->addr_str, addrbuffer);
```
rgetz, line 171: We use iio_strlcpy, rather than strcpy. Need to make sure that length includes the null char, or you will need to + 1 to it.

dns_sd_windows.c
```diff
@@ -0,0 +1,262 @@
...
+	int num_sockets = open_client_sockets(sockets, sizeof(sockets) / sizeof(sockets[0]));
+	if (num_sockets <= 0) {
+		IIO_ERROR("Failed to open any client sockets\n");
+		return -1;
+	}
+	IIO_DEBUG("Opened %d socket%s for mDNS query\n", num_sockets, num_sockets ? "s" : "");
+
+	IIO_DEBUG("Sending mDNS query: %s\n", service);
+	for (int isock = 0; isock < num_sockets; ++isock) {
+		transaction_id[isock] = mdns_query_send(sockets[isock], MDNS_RECORDTYPE_PTR, service, strlen(service), buffer,
```
rgetz, line 227: you can use `sizeof(service) - 1` rather than `strlen(service)`, and it is safer.

dns_sd_windows.c
```diff
@@ -0,0 +1,262 @@
...
+	IIO_DEBUG("DNS SD: Start service discovery.\n");
+
+	if (new_discovery_data(&d) < 0) {
+		return -ENOMEM;
+	}
+	*ddata = d;
+
+	size_t capacity = 2048;
+	void* buffer = malloc(capacity);
+	const char service[] = "_iio._tcp.local.";
```
rgetz, line 212: In the other implementation we use:
```
dns_sd_avahi.c:			"_iio._tcp", NULL, 0, __avahi_browser_cb, d);
dns_sd_bonjour.c:	CFStringRef type = CFSTR("_iio._tcp.");
./iiod/iiod.c:			0, avahi.name, "_iio._tcp", NULL, NULL, IIOD_PORT, NULL);
```

no fully qualified domain. I'm not sure which is 100% correct - but we should be doing the same for all.

mdns.h
```diff
@@ -0,0 +1,1205 @@
+/* mdns.h  -  mDNS/DNS-SD library  -  Public Domain  -  2017 Mattias Jansson
```
rgetz, line 1: Since it looks like this is under active development upstream, can we put the git hash that we used?

https://github.com/mjansson/mdns/commits/master/mdns.h

dns_sd_windows.c
rgetz: If the plan is to remove avahi and bonjour, and only use this - then that is different.... (Since we can't remove avahi - we use it in `iiod` to advertise - I see this being a Windows only file for a long time.

dns_sd_windows.c
adisuciu: I don't know what the correct version is, but not appending .local here makes it not detect anything.

rgetz: Ok - I guess that is the correct version :) Maybe the .local is appended in side the other libraries.

dns_sd_windows.c
```diff
@@ -0,0 +1,266 @@
...
+	free(d);
+}
+
+static int
+open_client_sockets(int* sockets, int max_sockets) {
+	// When sending, each socket can only send to one network interface
+	// Thus we need to open one socket for each interface and address family
+	int num_sockets = 0;
+
+#ifdef _WIN32
```
rgetz, line 56: This ifdef isn't needed?

dns_sd_windows.c
```diff
@@ -0,0 +1,266 @@
...
+			}
+#endif
+		}
+	}
+
+	free(adapter_address);
+#endif
+
+	for (int isock = 0; isock < num_sockets; ++isock) {
+#ifdef _WIN32
```
rgetz, line 128: same here?

dns_sd_windows.c
```diff
@@ -0,0 +1,266 @@
...
+			i = 0;
+		Sleep(100);
+	}
+
+	free(buffer);
+	for (int isock = 0; isock < num_sockets; ++isock)
+		mdns_socket_close(sockets[isock]);
+	IIO_DEBUG("Closed socket%s\n", num_sockets ? "s" : "");
+
+#ifdef _WIN32
```
rgetz, line 261: Same here

mdns.h
```diff
@@ -0,0 +1,1206 @@
...
+		return parsed;
+	}
+
+#ifdef _WIN32
+#undef strncasecmp
+#endif
+
+#ifdef __cplusplus
+}
+#endif
```
dNechita, line 1206: empty line is missing at EOF

rgetz: we dont control this file - it's an exact copy (plus the git hash) of upstream

dns_sd_windows.c
```diff
@@ -0,0 +1,257 @@
...
+		MDNS_STRING_FORMAT(srv.name), srv.priority, srv.weight, srv.port);
+
+	// Go to the last element in the list
+	while (dd->next != NULL)
+		dd = dd->next;
+
+	if (srv.name.length > 1)
+	{
+		dd->hostname = malloc(srv.name.length);
+		iio_strlcpy(dd->hostname, srv.name.str, srv.name.length);
```
rgetz, line 172: why remove the three lines, and  
```
dd->hostname = strndup(srv.name.str, srv.name.length);
```
With strndup, If s is longer than len, only len bytes are copied, and a terminating null byte  ('\0')  is  added.

but we should also check that it is null, and handle the error if it doesn't work. (the malloc fails)

dns_sd_windows.c
```diff
@@ -0,0 +1,257 @@
...
+
+	IIO_DEBUG("DNS SD: Start service discovery.\n");
+
+	if (new_discovery_data(&d) < 0) {
+		return -ENOMEM;
+	}
+	*ddata = d;
+
+	size_t capacity = 2048;
+	void* buffer = malloc(capacity);
```
rgetz, line 208: buffer is never checked to see if malloc() fails

dns_sd_windows.c
```diff
@@ -0,0 +1,257 @@
...
+	// When sending, each socket can only send to one network interface
+	// Thus we need to open one socket for each interface and address family
+	int num_sockets = 0;
+
+	IP_ADAPTER_ADDRESSES* adapter_address = 0;
+	ULONG address_size = 8000;
+	unsigned int ret;
+	unsigned int num_retries = 4;
+	do {
+		adapter_address = malloc(address_size);
```
rgetz, line 61: malloc isn't checked for failure.

mdns.h
adisuciu: An almost exact copy .. there was one or two warnings that didn't compile with /WX MSVC compiler switch ..

dns_sd_windows.c
adisuciu: MSVC does not have strndup, only strdup which doesn't cut it .. We can use strdup and nullterminate it. I can just leave it like this and add the NULL check.

dns_sd_windows.c
```diff
@@ -0,0 +1,257 @@
...
+ */
+
+#include <stdio.h>
+#include <errno.h>
+#include "iio-private.h"
+#include "mdns.h"
+#include "network.h"
+#include "debug.h"
+
+#  include <iphlpapi.h>
```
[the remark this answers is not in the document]
adisuciu: Including this before "iio-private.h" creates a collision

in combaseapi.h (windows header) which is included somewhere in <iphlpapi.h> 

```
#define __STRUCT__ struct
#define interface __STRUCT__
```

so "interface" becomes a keyword which collides with this line:
https://github.com/analogdevicesinc/libiio/blob/master/iio-private.h#L288

A proper fix would be to refactor the 
struct iio_context * usb_create_context(unsigned int bus, uint16_t address,
		uint16_t interface);

so it won't use the keyword.

dns_sd_windows.c
adisuciu: fixed

dns_sd_windows.c
adisuciu: fixed

dns_sd_windows.c
```diff
@@ -0,0 +1,257 @@
...
+	while (dd->next != NULL)
+		dd = dd->next;
+
+	if (srv.name.length > 1)
+	{
+		dd->hostname = malloc(srv.name.length);
+		iio_strlcpy(dd->hostname, srv.name.str, srv.name.length);
+		dd->hostname[srv.name.length - 1] = 0;
+	}
+	iio_strlcpy(dd->addr_str, addrbuffer, strlen(addrbuffer));
```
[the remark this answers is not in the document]
adisuciu: fixed

dns_sd_windows.c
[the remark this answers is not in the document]
dNechita: Ok, makes sense.

dns_sd_windows.c
rgetz: Ok - we have iio_strdup - I would just use that., but we still need the fail / null checking...

dns_sd_windows.c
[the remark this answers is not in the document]
rgetz: I can fix that on master later today.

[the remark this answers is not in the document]
rgetz: Done as #512 - just need it to build everywhere, and have someone approve it.

dns_sd_windows.c
adisuciu: The string is not null terminated. iio_strdup is bascially a malloc and memcopy which does not guarantee null termination
I used malloc and iio_strlcpy (which guarantees nulltermination). Also added errorcheck on malloc.

I don't think we can get away with one operation here ..

rgetz: Ok - thanks

rgetz:
This looks good to me. 

Did you want to squash some of those commits?

rgetz:
Looks like there are some formatting changes between upstream msdn.h and ours?

lots of:
```
--- ./libiio/mdns.h      2020-05-22 11:00:41.592646263 -0400
+++ ../mdns/mdns.h      2020-05-22 10:59:35.848102185 -0400

-                       }
-                       else {
+               } else {
```

Just makes it harder to compare other differences...

PR #509, opened 2020-05-20
add uri context attribute

rgetz:
The uri context attribute is added as "local:" (which is what is in iiod), and as this is passed to various remote devices, it will be overwritten by the remote method (usb, ip or serial).

-Robin


rgetz:
Updated with a rebase on master to make sure Windows works properly.

PR #510, opened 2020-05-21
coverity: handle strtol and strtoll under|over-flows properly

rgetz:
a recent change that we (I) made to things wasn't handling
under|over-flows properly. The man page for strtol states:

Since strtol() can legitimately return 0, LONG_MAX, or LONG_MIN on both
success and failure, the calling program should set errno to 0 before the
call, and then determine if an error occurred by checking whether errno
has a nonzero value after the call.

So do that - set errno to 0, and see if it is errno is set to ERANGE
when it's finished.

Review the entire code base, and do the same thing everywhere.

### DIFF
--- a/tests/iio_common.c
+++ b/tests/iio_common.c
@@ -133,14 +133,17 @@ unsigned long int sanitize_clamp(const char *name, const char *argv,
 	uint64_t min, uint64_t max)
 {
 	uint64_t val;
-	char buf[20];
+	char buf[20], *end;
 
 	if (!argv) {
 		val = 0;
 	} else {
 		/* sanitized buffer by taking first 20 (or less) char */
 		iio_snprintf(buf, sizeof(buf), "%s", argv);
-		val = strtoul(buf, NULL, 10);
+		errno = 0;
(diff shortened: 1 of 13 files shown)


tests/iio_common.c
```diff
@@ -133,14 +133,16 @@ unsigned long int sanitize_clamp(const char *name, const char *argv,
...
-	char buf[20];
+	char buf[20], *end;
 
 	if (!argv) {
 		val = 0;
 	} else {
 		/* sanitized buffer by taking first 20 (or less) char */
 		iio_snprintf(buf, sizeof(buf), "%s", argv);
-		val = strtoul(buf, NULL, 10);
+		val = strtoul(buf, &end, 10);
```
dNechita, line 143: Shouldn't errno be set to 0 before this call?

rgetz: should be fixed.

rgetz:
Since I added checks for more functions/touched more things - I will let @dNechita review/approve again.

rgetz:
update to master

rgetz:
All fixed:
image

PR #513, opened 2020-05-22
local: add some more info to local scan contexts

rgetz:
In the past, when doing an 'iio_info -s' on machines with local contexts,
you would get something like:
  "Local devices"
which isn't exactly descriptive...

look in sysfs for the machine (DT model for embedded ARM and DRI for x86),
and combine that with a device list so now you get something that looks like:

Available contexts:
	0: (spi1.0, ad9361-phy, cf-ad9361-lpc, spi1.1, xadc, ad7291, cf-ad9361-dds-core-lpc on Xilinx Zynq ZED) [local:]

or

Available contexts:
	0: (ad7124-4 on Raspberry Pi 4 Model B Rev 1.1) [local:]

which makes it easier to identify for end users.

### DIFF
--- a/local.c
+++ b/local.c
@@ -2080,6 +2080,48 @@ err_set_errno:
 	return NULL;
 }
 
+#define BUF_SIZE 128
+
+static char * cat_file(const char *path)
+{
+	char buf[BUF_SIZE];
+	ssize_t ret;
+
+	FILE *f;
+
+	f = fopen(path, "re");
+	if (!f)
+		return NULL;
+
+	ret = fread(buf, 1, sizeof(buf)-1, f);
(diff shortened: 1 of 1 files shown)


local.c
```diff
@@ -2080,6 +2080,48 @@ struct iio_context * local_create_context(void)
...
+#define BUF_SIZE 256
+
+static char * cat_file(const char *path)
+{
+	char buf[BUF_SIZE];
+	ssize_t ret;
+
+	FILE *f;
+
+       	f = fopen(path, "re");
```
dNechita, line 2092: This is not aligned

rgetz: fixed

PR #515, opened 2020-05-22
coverity: fix resource leak in network.c

rgetz:
the buffer 'uri' was malloced, and only free'd on the error case.
free it in the normal case.


rgetz:
This is OK to be merged - appveyor -> github got lost.

rgetz:
Fixed:
image

PR #516, opened 2020-05-22
test memleaks

rgetz:
fix issues pointed out by:
```
valgrind --tool=memcheck --leak-check=full --show-leak-kinds=all iio_util
```

needs some more testing, but wanted to check it on CI as well.


rgetz:
Now that @pcercuei is back - I was wondering if you could have a look at this?

pcercuei:
@rgetz I'm not really fond of the added API function.

Do you need it to be available on every supported platform and compiler?

If it's acceptable to restrain this to GCC, you can do this in the XML backend, and it'd happen automatically:

```
static void __attribute__((destructor)) xml_purge_context(void)
{
	xmlCleanupParser();
	xmlMemoryDump();
}
```

pcercuei:
The equivalent is possible in Visual Studio (just tested it) with some trickery: https://stackoverflow.com/questions/1113409/attribute-constructor-equivalent-in-vc

pcercuei:
See my PR: #679.

rgetz:
looks much better - I will close this, and make a new one for the applications... (which should be easier - since you don't need to purge anything and have one exit anymore).

Issue #519, opened 2020-05-26
xml reserved chars are not escaped 


rgetz:
If a attribute has an xml reserved char, we don't escape it, and it causes the xml parser to blow up.

We get errors like:
```
Entity: line 1: parser error : Unescaped '<' not allowed in attributes values
armv7l" ><context-attribute name="hdl_system_id" value="[fmcomms2] on [zed] git 
                                                                               ^
Entity: line 1: parser error : attributes construct error
armv7l" ><context-attribute name="hdl_system_id" value="[fmcomms2] on [zed] git 
                                                                               ^
Entity: line 1: parser error : Couldn't find end of Start Tag context-attribute line 1
armv7l" ><context-attribute name="hdl_system_id" value="[fmcomms2] on [zed] git 
                                                                               ^
Entity: line 1: parser error : StartTag: invalid element name
rmv7l" ><context-attribute name="hdl_system_id" value="[fmcomms2] on [zed] git <
                                                                               ^
ERROR: Unable to parse XML file
Unable to create IIO context ip:analog.local: Invalid argument
```

OR

```
Entity: line 1: parser error : attributes construct error
rmv7l" ><context-attribute name="hdl_system_id" value="[fmcomms2] on [zed] git "
                                                                               ^
Entity: line 1: parser error : Couldn't find end of Start Tag context-attribute line 1
rmv7l" ><context-attribute name="hdl_system_id" value="[fmcomms2] on [zed] git "
                                                                               ^
ERROR: Unable to parse XML file
Unable to create IIO context ip:analog.local: Invalid argument
```

OR

```
Entity: line 1: parser error : xmlParseEntityRef: no name
tem_id" value="[fmcomms2] on [zed] git 2e4ac278eb09c13471e381459b0da790ebad8373&
                                                                               ^
ERROR: Unable to parse XML file
Unable to create IIO context ip:analog.local: Invalid argument
```

This isn't an issues for most kernel attributes, as they are pure numbers, but it is a problem on some of the new ini file processing, which can contain more free text.

Char | Escape String
--- | ---
`<` | `&lt;`
`>` | `&gt;`
`"` | `&quot;`
`'` | `&apos;`
`&` | `&amp;`

We also don't verify that Element and Attribute names can NOT contain characters `<`, `>`, `"`, `'`, or `&` escaped or otherwise. (per the xml standard).

rgetz:
I think this should only effect:
```
attribute name
buffer-attribute name
channel id
[...]

PR #526, opened 2020-06-01
xml: Add capability to properly encode xml entities in context attrib…

rgetz:
…utes

Per #519, xml entities (<, >, ", ', and &) were not encoded properly, and if
used in the libiio.ini file (which was added in 77568c71ec0d7b0162f0b488f261d6d9874f2e88 )
cause the libiio library to fail.

This fixed that by adding a generic way to encode things in xml.c and uses them in context.c

Right now, we only encode context attributes. In theory this could also effect:
 - attribute name
 - buffer-attribute name
 - channel id
 - context name
 - debug-attribute name
 - device id
 - scan-element index
Since most of those are encoded in file names, we shouldn't have those chars
anyway. Now that we know the issue, and understand the fix, if anyone sees
any issues, please feel free to report.

Tested on Zed + FMCOMMS3, with both 0.19 (with fix) and old libraries without fix.
As long as the the version which is reading the file has the fix - it works with
older libraries (is backwards compatible).

### DIFF
--- a/xml.c
+++ b/xml.c
@@ -23,6 +23,16 @@
 #include <libxml/tree.h>
 #include <string.h>
 
+/* 'input' must be in UTF-8 encoded, null terminated */
+char * encode_xml_ndup(const char * input)
+{
+	char * out;
+
+	out = (char *)xmlEncodeEntitiesReentrant(NULL, (const xmlChar *)input);
+
+	return out;
+}
+
 static int add_attr_to_channel(struct iio_channel *chn, xmlNode *n)
 {
 	xmlAttr *attr;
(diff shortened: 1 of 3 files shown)


xml.c
```diff
@@ -23,6 +23,17 @@
...
 
+/* 'input' must be in UTF-8 encoded, null terminated */
+char * encode_xml_ndup(const char * input)
+{
+	char * out;
+
+	out = (char *)xmlEncodeEntitiesReentrant(NULL, (const xmlChar *)input);
+
+	return out;
+
```
dNechita, line 34: minor thing: Extra line here

rgetz: fixed

PR #534, opened 2020-06-02
fix recent coverity issues

rgetz:
two issues while tracking down others...


rgetz:
This gets things back down to zero.

image

PR #538, opened 2020-06-04
iio_utils: rework/fix command line argument parsing

rgetz:
the rework and moving to common functions for argument parsing broke a
few things, since I thought that you could pass a incomplete optstring
and incomplete longopts to getopt_long(). However, this is not the case.
It broke long flags (--uri vs -u) and broke the position parsing that
happened deep inside getopt_long() somewhere.

In fact - this caused builds on MSVC to crash sometimes. The other thing
that caused MSVC to crash was if an flag that had an optional argument
was the last on on the command line (for example 'iio_info -a') we would
use a index that exceeded the actual memory allocations, so check for
that and don't do it.

Now, make sure to pass complete arguments to things, and fix the bounds
checking which seems to resolve all the issues on Linux and Windows.

### DIFF
--- a/tests/iio_common.c
+++ b/tests/iio_common.c
@@ -196,6 +196,43 @@ static const struct option common_options[] = {
 	{0, 0, 0, 0},
 };
 
+struct option * add_common_options(const struct option * longopts)
+{
+	int i = 0, j = 0;
+	struct option *opts;
+
+	while (longopts[i].name) {
+		i++;
+	}
+	while (common_options[j].name) {
+		j++;
+		i++;
+	}
+	opts = calloc(i + 1, sizeof(struct option));
+	if (!opts) {
+		fprintf(stderr, "Out of memory\n");
+		return NULL;
+	}
+	i = 0, j = 0;
+	while (longopts[i].name) {
+		opts[i].name = longopts[i].name;
+		opts[i].has_arg = longopts[i].has_arg;
+		opts[i].flag = longopts[i].flag;
+		opts[i].val = longopts[i].val;
+		i++;
+	}
+	while (common_options[j].name) {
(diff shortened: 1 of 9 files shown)


tests/iio_common.c
```diff
@@ -196,6 +196,41 @@ static const struct option common_options[] = {
 	{0, 0, 0, 0},
 };
 
+struct option * add_common_options(const struct option * longopts)
+{
+	int i = 0, j = 0;
+	struct option *opts;\
```
dNechita, line 202: The backslash should be removed

tests/iio_common.c
```diff
@@ -207,22 +242,32 @@ static const char *common_options_descriptions[] = {
...
 	opterr = 0;
 	/* start over at first index */
 	optind = 1;
 
-	while ((c = getopt_long(argc, argv, COMMON_OPTIONS,	/* Flawfinder: ignore */
-			common_options, &option_index)) != -1) {
+	iio_snprintf(buf, sizeof(buf), "%s%s", COMMON_OPTIONS, optstring);
+	opts = add_common_options(options);
+	if (!opts) {
+		fprintf(stderr, "Memory error\n");
```
dNechita, line 265: To me, it makes more sense to say here: "failed to add common options" and inside add_common_options to add the message "Memory error" after calloc() fails. Today add_common_option() might fail due to one reason but as the function might change over time, other reasons to fail might get added.

tests/iio_common.c
rgetz: fixed

tests/iio_common.c
rgetz: Makes sense. fixed

PR #540, opened 2020-06-04
Use https for download links in README.md

MartinSpiessl:
Packages are installed by people on their machines and should therefore not be downloaded via insecure http connections. The README.md currently links to the libiio software packages with http links.

The changes in this commit can be reproduced by the following command:

    sed -i "s|\(http\)\(://swdownloads\)|\1s\2|g" README.md

This adds an "s" after http for every occurrence of `http://swdownloads`.

Currently the certificate for `https://swdownloads.analog.com` is invalid (cname only lists `a248.e.akamai.net`), so not much is to be gained by accepting this PR security wise, and users would be irritated by the "insecure connection" warning. I would recommend getting that cert fixed and then merge this PR, or apply the changes yourself with the provided sed command (it is hard to tell that I did not change anything else, since github's diff engine seems to fail in highlighting all changes appropriately).

A different way to fix this might be to link to the releases page under https://github.com/analogdevicesinc/libiio/releases instead, but as far as I can tell the intention was to link to the latest build from the CI, in which case linking to the releases is not an equivalent substitute.


rgetz:
Looks cool - thank for the contribution.

I will poke someone about the cert.

It just needs a signed-off-by in the commit message.

rgetz:
I guess it will take a week or so to get all the certs done - and transition some things on the akamai backend. Will update this when that is done.

MartinSpiessl:
> It just needs a signed-off-by in the commit message.

Sure, I just force-pushed a new version of the commit with signed-off-by information created by calling `git commit --amend --sign-off` (and also fixed a typo in the commit message I discovered afterwards, that's why there are two force-pushs).

rgetz (approved):

PR #542, opened 2020-06-04
iio_attr: change verbosity of output

rgetz:
decrease the output of iio_attr.
To go back to "normal", use the new '-v' flag (verbose).
To quiet everything use the '-q', which is useful for checking error
codes.

tweak the context creation in iio_common for better error reporting

### DIFF
--- a/tests/iio_common.c
+++ b/tests/iio_common.c
@@ -236,11 +236,17 @@ struct option * add_common_options(const struct option * longopts)
 static const char *common_options_descriptions[] = {
 	"Show this help and quit.",
 	"Use the XML backend with the provided XML file.",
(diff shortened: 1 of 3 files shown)


tests/iio_common.c
```diff
@@ -307,14 +307,15 @@ struct iio_context * handle_common_opts(char * name, int argc,
 				fprintf(stderr, "uri options requires a uri\n");
 				return NULL;
 			}
-			backend = IIO_AUTO;
```
dNechita, line 310: It would be nice to have these changes (about context creation) in a separate commit.

rgetz: done. Sorry about that.

PR #544, opened 2020-06-05
libiio.iss.cmakein: Drop the "ignoreversion" flag and use the "replacesameversion" flag instead.

AlexandraTrifan:
From the official documentation: "When this flag is used and the file
already exists on the user's system and it has the same version number
as the file being installed, Setup will compare the files and replace
the existing file if their contents differ.
The default behavior (i.e. when this flag isn't used) is to not replace
an existing file with the same version number."


rgetz:
Thanks - that gets rid of all the warnings that were there...

PR #551, opened 2020-06-08
iiod: Make startup (with Avahi) more robust.

rgetz:
In certain embedded cases, iiod may start before networking or dbus is
fully alive on a system (like on Pluto and M2k).

This change ensures that when we are trying to connect to avahi, and
register on the network, we give things a few moments to stablize (re-trying)
before we fail.

tested on PlutoSDR, and Zedboard + FMCOMMS3

### DIFF
--- a/iiod/iiod.c
+++ b/iiod/iiod.c
@@ -219,9 +219,14 @@ static void __avahi_client_cb(AvahiClient *client,
 			create_services(client);
 			break;
 		case AVAHI_CLIENT_FAILURE:
-			IIO_ERROR("Avahi: Client failure: %s\n",
+			if (avahi_client_errno(client) != AVAHI_ERR_DISCONNECTED) {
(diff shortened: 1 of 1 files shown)


rgetz:
This did build on travis - I'm not sure why it's not showing up....
https://travis-ci.org/github/analogdevicesinc/libiio/builds/696242826

iiod/iiod.c
```diff
@@ -481,7 +534,7 @@ static int main_server(struct iio_context *ctx, bool debug)
 	}
 
 	if (ipv6)
-		IIO_INFO("IPv6 support enabled\n");
+		IIO_ERROR("IPv6 support enabled\n");
```
dNechita, line 537: Why is this an error?

rgetz: This was an accident - I was easier to change things to IIO_ERROR that recompile the entire library, and this never got changed back. Thanks for looking.

PR #563, opened 2020-07-06
cmake: Fix paths in pkg-config file

jtojnar:
It is not generally true that `CMAKE_INSTALL_<dir>` variables are relative paths:

https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files

Absolute paths will cause things like this appear:

    includedir=${prefix}//nix/store/87zhscw1p08nqm73ls2sd82zmr1a8rni-libiio-0.21-dev/include

Signed-off-by: Jan Tojnar <jtojnar@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -379,7 +379,13 @@ endif()
 
 configure_file(libiio.iss.cmakein ${CMAKE_CURRENT_BINARY_DIR}/libiio.iss @ONLY)
 
+include(cmake/JoinPaths.cmake)
+
+join_paths(libdir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_LIBDIR}")
+join_paths(includedir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_INCLUDEDIR}")
+
 set(LIBIIO_PC ${CMAKE_CURRENT_BINARY_DIR}/libiio.pc)
+
 configure_file(libiio.pc.cmakein ${LIBIIO_PC} @ONLY)
 
 if(NOT SKIP_INSTALL_ALL)
--- a/libiio.pc.cmakein
+++ b/libiio.pc.cmakein
@@ -1,7 +1,7 @@
 prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=${prefix}
-libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+libdir=@libdir_for_pc_file@
+includedir=@includedir_for_pc_file@
 
 Name: libiio
 Description: Library for interfacing IIO devices
--- a/cmake/JoinPaths.cmake
+++ b/cmake/JoinPaths.cmake
@@ -1,0 +1,23 @@
+# This module provides function for joining paths
+# known from from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
+# Copyright 2020 Jan Tojnar
+# https://github.com/jtojnar/cmake-snips
+#
+# Modelled after Python’s os.path.join
+# https://docs.python.org/3.7/library/os.path.html#os.path.join
+# Windows not supported
+function(join_paths joined_path first_path_segment)
+    set(temp_path "${first_path_segment}")
+    foreach(current_segment IN LISTS ARGN)
+        if(NOT ("${current_segment}" STREQUAL ""))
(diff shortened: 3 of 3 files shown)


libiio.pc.cmakein
```diff
@@ -1,7 +1,7 @@
 prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=${prefix}
-libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+libdir=@libdir_for_pc_file@
```
commodo, line 3: what if we do:
```
libdir=@CMAKE_INSTALL_LIBDIR@
includedir=@CMAKE_INSTALL_INCLUDEDIR@
```

and make sure that `CMAKE_INSTALL_xxxx` are absolute paths?
and if they aren't absolute maybe we cal do `CMAKE_INSTALL_LIBDIR=CMAKE_INSTALL_PREFIX/CMAKE_INSTALL_LIBDIR`

the overall `join_paths()` looks nice, but it's a bit more logic than is required to solve the actual problem;
and it looks to me that `CMAKE_INSTALL_LIBDIR` (and includedir) is used in CMakeLists.txt as well;
so there can be a minor inconsistency between the values of  `CMAKE_INSTALL_LIBDIR`  use in libiio.pc.cmakein  and the ones used in CMakeLists.txt

so, maybe something like?:
```
if(NOT IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
CMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}
endif
```

CMakeLists.txt
```diff
@@ -379,7 +379,13 @@ endif()
 
 configure_file(libiio.iss.cmakein ${CMAKE_CURRENT_BINARY_DIR}/libiio.iss @ONLY)
 
+include(cmake/JoinPaths.cmake)
+
+join_paths(libdir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_LIBDIR}")
+join_paths(includedir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_INCLUDEDIR}")
```
commodo, line 385: continuing from comment https://github.com/analogdevicesinc/libiio/pull/563/commits/48e1606c30b5ae567aee55ec7bb6765d0e5b79ac#r449998108
this is where we could do:
```
if(NOT IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
CMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}
endif
```

commodo:
thanks for the find :)

libiio.pc.cmakein
jtojnar: > and make sure that `CMAKE_INSTALL_xxxx` are absolute paths?

There are `FULL` variants of the variables, see the [docs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables) so that could be used if we decide to go that way.

The reason why there is `${prefix}` is to make it overridable (at least in the cases when it is not absolute), see https://www.bassi.io/articles/2018/03/15/pkg-config-and-paths/.

commodo: hmm, you are right;
i am just a bit hesitant about having an extra level of indirection for changing from `includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@` to constructing `${prefix}/@CMAKE_INSTALL_INCLUDEDIR@` in CMake via some function;

if ` libiio.pc.cmakein` will only be generated from CMake, then this code seems a bit much;
i'm not fully against it, i just prefer something simpler to read/review

jtojnar: I guess the absolute paths are fine if libiio does not expect other software to ever install to their `libdir` or `includedir`.

commodo: if this would fix your case as well, i would prefer to do it simpler for now;
if we need to add the complexity of `join_paths()` we can add it later;

jtojnar: Depends on whether `libiio` is extensible and the extension points are the installation paths. If not, I was not able to come up with a reason https://github.com/analogdevicesinc/libiio/pull/564 should fail.

commodo: oh; i did not notice the other PR;
too many emails i think

rgetz:
#564 was merged instead of this - so closing.

PR #565, opened 2020-07-07
tests: iio_attr: Add support for controlling the context timeout

mhennerich:

### DIFF
--- a/tests/iio_attr.c
+++ b/tests/iio_attr.c
@@ -312,6 +312,7 @@ static const struct option options[] = {
 	{"quiet", no_argument, 0, 'q'},
 	{"verbose", no_argument, 0, 'v'},
 	{"generate-code", required_argument, 0, 'g'},
+	{"timeout", required_argument, 0, 't'},
 	/* Channel qualifiers */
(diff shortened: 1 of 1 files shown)


rgetz:
shouldn't this be a common thing that is added to all utilities? (which is already in some of them)?

mhennerich:
I think it now in all tools which requires it. 
Maybe we should also use the set value for codegen. There is already: 
void gen_context_timeout(unsigned int timeout_ms); 
which doesn't get used at all.

rgetz:
yeah, its:

```
iio_readdev.c:	  {"timeout", required_argument, 0, 'T'},
iio_writedev.c:	  {"timeout", required_argument, 0, 'T'},
iio_stresstest.c:	{"timeout", required_argument, 0, 't'},
```

so, I think it makes sense to shuffle `iio_stresstest` to use `"T"` and move your patch to the common parts, and like you said - connect the gen_code parts... 

I can do that this afternoon if you want.

mhennerich:
Sounds great - Thanks

rgetz:
Ok - closing this.

PR #567, opened 2020-07-10
dnssd: Use lower case when including Windows headers

dNechita:
This would help the cross compile process where these headers are not
found because their filenames are lowercase. This happens if the host
where the cross compiling is done is case sensitive when it comes to
file paths.


rgetz:
thanks

Issue #569, opened 2020-07-16
cross compile for window on ubuntu  error: no TunnelType member in struct IPADAPTER_ADDRESS


anzyelay:
https://github.com/analogdevicesinc/libiio/blob/740b3f49b82be96b0a10620adcf7741eec3cd4c3/dns_sd_windows.c#L82

when i cross compile it ,i found the definitions _WIN32_WINNT means XP, so the struct  in iptypes.h pointer to an XP struct which has no  member TunnelType.

anzyelay:
i  have resolved it in  temporary as follow:

diff --git a/CMakeLists.txt b/CMakeLists.txt
index c044fe0..121ad48 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,6 +312,7 @@ if(WITH_NETWORK_BACKEND)
                list(APPEND LIBS_TO_LINK ${AVAHI_LIBRARIES})
        elseif(WIN32)
                set(HAVE_DNS_SD ON)
+               set(CMAKE_C_FLAGS "-D_WIN32_WINNT=0x0600 ${CMAKE_C_FLAGS}")
                list(APPEND LIBIIO_CFILES dns_sd_windows.c dns_sd.c)
                if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
                        set_source_files_properties(dns_sd_windows.c PROPERTIES COMPILE_FLAGS "-Wno-unused-function")
diff --git a/mdns.h b/mdns.h
index 58a5bb5..7e164f7 100644
--- a/mdns.h
+++ b/mdns.h
@@ -22,8 +22,8 @@
 
 #include <fcntl.h>
 #ifdef _WIN32
-#include <Winsock2.h>
-#include <Ws2tcpip.h>
+#include <winsock2.h>
+#include <ws2tcpip.h>
 #define strncasecmp _strnicmp
 #else
 #include <unistd.h>

rgetz:
Thanks for the report.

`_WIN32_WINNT` should be set by the [compiler](https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=vs-2019) where `_WIN32_WINNT >= 0x600` means one of:
```
#define _WIN32_WINNT_WIN6                   0x0600 // Windows Vista
#define _WIN32_WINNT_VISTA                  0x0600 // Windows Vista
[...]

PR #570, opened 2020-07-16
errors  occurs in dns_sd_windows.c

anzyelay:
errors  occurs in dns_sd_windows.c when cross compile where ignoring  _WIN32_WINNT.
and lowercases on including files


mdns.h
```diff
@@ -22,8 +22,8 @@
 
 #include <fcntl.h>
 #ifdef _WIN32
-#include <Winsock2.h>
```
dNechita, line 25: There is already a PR in review for the header fix: https://github.com/analogdevicesinc/libiio/pull/567

CMakeLists.txt
```diff
@@ -312,6 +312,7 @@ if(WITH_NETWORK_BACKEND)
 		list(APPEND LIBS_TO_LINK ${AVAHI_LIBRARIES})
 	elseif(WIN32)
 		set(HAVE_DNS_SD ON)
+		set(CMAKE_C_FLAGS "-D_WIN32_WINNT=0x0600 ${CMAKE_C_FLAGS}")
```
dNechita, line 315: I wouldn't want this flag set when building on Windows (not cross compiling) since it freezes the value of D_WIN32_WINNT to 0x0600. On windows this flag gets set automatically. However you could check if the project is being cross-compiled and then set this flag (check if CMAKE_COMPILER_IS_GNUCC or Clang).

rgetz:
yeah - sorry - this is not the right fix, can you follow up on the #569 issue tracker?

rgetz:
I have a fix in #574 (which is similar), but more general purpose - since we should be setting this all the time - not just on one file / if we include dns-sd...

@anzyelay Can you check that out?

rgetz:
I will close this, since the other implementation was merged.

PR #571, opened 2020-07-20
Update python package name

tfcollins:
The name libiio is not available in pypi since Fedora blocked out future
package names that share the same name as their yum/apt names.
Therefore, we switch to using the name pylibiio for pypi. This change
updates the doc and packaging script. v0.21 has already been uploaded
with this name to pypi.


rgetz:
Thanks

Issue #555, opened 2020-06-17
0.21: cmake fails with WITH_MAN=ON


kloczek:
```
+ /usr/bin/cmake -D BUILD_SHARED_LIBS=ON -D CMAKE_AR=/usr/bin/gcc-ar -D CMAKE_BUILD_TYPE=RelWithDebInfo -D CMAKE_C_FLAGS_RELEASE=-DNDEBUG -D CMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -D CMAKE_Fortran_FLAGS_RELEASE=-DNDEBUG -D CMAKE_INSTALL_PREFIX=/usr -D CMAKE_NM=/usr/bin/gcc-nm -D CMAKE_RANLIB=/usr/bin/gcc-ranlib -D CMAKE_VERBOSE_MAKEFILE=ON -D INCLUDE_INSTALL_DIR=/usr/include -D LIB_INSTALL_DIR=/usr/lib64 -D LIB_SUFFIX=64 -D SHARE_INSTALL_PREFIX=/usr/share -D SYSCONF_INSTALL_DIR=/etc . -D INSTALL_UDEV_RULE=ON -D PYTHON_BINDINGS=ON -D WITH_DOC=ON -D WITH_MAN=ON -D WITH_SERIAL_BACKEND=ON -D WITH_SYSTEMD=ON -D WITH_SYSVINIT=OFF -D WITH_TESTS=ON -D WITH_USB_BACKEND=ON -D WITH_XML_BACKEND=ON
-- cmake version: 3.17.3
-- The C compiler identification is GNU 10.1.1
-- Check for working C compiler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc - works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test HAS_WPEDANTIC
-- Performing Test HAS_WPEDANTIC - Success
-- Looking for strdup
-- Looking for strdup - found
-- Looking for strndup
-- Looking for strndup - found
-- Looking for strerror_r
-- Looking for strerror_r - found
-- Looking for newlocale
-- Looking for newlocale - found
-- Looking for pthread_setname_np
-- Looking for pthread_setname_np - found
-- Looking for in6addr_any
-- Looking for in6addr_any - found
-- Looking for libusb-1.0 : Found
-- Looking for libusb_get_version
-- Looking for libusb_get_version - found
[...]

PR #572, opened 2020-07-22
man: allow man pages to be built when not in a ./build dir

rgetz:
as pointed out in #555, we assumed that iio.h was always one directory
up, which is sort of bad form.

Don't make that assumption, and process things with the CMAKE system.


kloczek:
Confirmed. With that patch I was able to build libiio using off-tree build.

Thx :)

PR #575, opened 2020-07-24
local: harden local_close() and use it for local_open() failure

commodo:
There's a bug, when enabling a buffer (with local_buffer_enabled_set(),
i.e. writing to 'buffer/enable'), we can get -EINVAL, because the scan_mask
is invalidated by the kernel.
The issue seems to be that the local_open() exit-on-failure path is
un-balanced. Some things don't seem to be cleaned-up on exit-on-failure.

This is seen with iiod trying to open a high-speed buffer device, where the
exit-on-failure path does not call munmap() on the buffers, nor does it
free the memory allocated for the buffer (in libiio). On the next
retry/open the buffer device is locked and -EBUSY is returned.
Essentially, this leaks resources (a file-descriptor and some memory for
the buffers).
Closing and re-opening iiod fixes the issue.

Also, the local_close() function should not exit on the first fail in
finds. It should continue with disabling everything it can. The only error
that might make sense, is if the main FD is not initialized.

This change simplifies the local_close() a bit: we return -EBADF if the
main FD is -1. Other than that, we continue disabling/cleaning everything
else (with the proper NULL/fd checks in place).

This makes it re-usable for the local_open() exit-on-failure path.
This could be addressed [also] by balancing the initialization in the
reverse order of init. But it looks like something might be omitted later
(as they were up until now) in the exit-on-failure path.
So maybe a more future-proof method is to just re-use the local_open()
routine where possible.

### DIFF
--- a/local.c
+++ b/local.c
@@ -998,10 +998,14 @@ static int local_close(const struct iio_device *dev)
 {
 	struct iio_device_pdata *pdata = dev->pdata;
 	unsigned int i;
+	char err_str[32];
+	int ret, ret1;
 
 	if (pdata->fd == -1)
 		return -EBADF;
 
+	ret = 0;
+	ret1 = 0;
 	if (pdata->is_high_speed) {
 		unsigned int i;
 		if (pdata->addrs) {
@@ -1009,7 +1013,12 @@ static int local_close(const struct iio_device *dev)
 				munmap(pdata->addrs[i], pdata->blocks[i].size);
 		}
 		if (pdata->fd > -1)
-			ioctl_nointr(pdata->fd, BLOCK_FREE_IOCTL, 0);
+			ret = ioctl_nointr(pdata->fd, BLOCK_FREE_IOCTL, 0);
+		if (ret) {
+			ret = -errno;
+			iio_strerror(errno, err_str, sizeof(err_str));
+			IIO_ERROR("Error during ioctl(): %s\n", err_str);
+		}
 		pdata->allocated_nb_blocks = 0;
 		free(pdata->addrs);
 		pdata->addrs = NULL;
@@ -1017,24 +1026,59 @@ static int local_close(const struct iio_device *dev)
 		pdata->blocks = NULL;
 	}
 
-	close(pdata->fd);
+	ret1 = close(pdata->fd);
+	if (ret1) {
+		ret1 = -errno;
+		iio_strerror(errno, err_str, sizeof(err_str));
+		IIO_ERROR("Error during close() of main FD: %s\n", err_str);
+		if (ret == 0)
+			ret = ret1;
+	}
+
 	pdata->fd = -1;
 
 	if (pdata->cancel_fd > -1) {
 		close(pdata->cancel_fd);
 		pdata->cancel_fd = -1;
+
+		if (ret1) {
+			ret1 = -errno;
+			iio_strerror(errno, err_str, sizeof(err_str));
+			IIO_ERROR("Error during close() of cancel FD): %s\n",
+				  err_str);
+			if (ret == 0)
+				ret = ret1;
+		}
 	}
 
-	local_buffer_enabled_set(dev, false);
+	ret1 = local_buffer_enabled_set(dev, false);
+	if (ret1) {
+		ret1 = -errno;
+		iio_strerror(errno, err_str, sizeof(err_str));
+		IIO_ERROR("Error during buffer disable: %s\n", err_str);
+		if (ret == 0)
+			ret = ret1;
+	}
 
 	for (i = 0; i < dev->nb_channels; i++) {
 		struct iio_channel *chn = dev->channels[i];
 
-		if (chn->pdata->enable_fn)
-			channel_write_state(chn, false);
+		if (!chn->pdata->enable_fn)
+			continue;
+
+		ret1 = channel_write_state(chn, false);
+		if (ret1 == 0)
+			continue;
+
+		ret1 = -errno;
+		iio_strerror(errno, err_str, sizeof(err_str));
+		IIO_ERROR("Error during channel[%u] disable: %s\n",
+			  i, err_str);
+		if (ret == 0)
+			ret = ret1;
 	}
 
-	return 0;
+	return ret;
 }
 
 static int local_get_fd(const struct iio_device *dev)
--- a/local.c
+++ b/local.c
@@ -97,7 +97,7 @@ struct iio_device_pdata {
 	struct block *blocks;
 	void **addrs;
 	int last_dequeued;
-	bool is_high_speed, cyclic, cyclic_buffer_enqueued, buffer_enabled;
+	bool is_high_speed, cyclic, cyclic_buffer_enqueued;
 
 	int cancel_fd;
 };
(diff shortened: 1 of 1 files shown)


rgetz:
I don't know if @pcercuei or @larsclausen have time to have a peak at this one?

larsclausen:
> There's a bug...

What are the effects of the bug? Does it crash?

commodo:
> > There's a bug...
> 
> What are the effects of the bug? Does it crash?

Hmm, maybe I should describe it a bit better in the commit desc.
Also, I'll admit, I may need to dig this into the kernel a bit.
I was a bit happy to just use this hardened `local_open()` vs going deeper.

So, issue is: `local_open()` creates FD, `local_buffer_enabled_set()` fails [with -EINVAL], then on the next `local_open()` call, it returns -EBUSY (busy flag in the IIO device is still left as set), the IIO device is stuck as busy;
The issue is with iiod. Since it's a process.
Restarting iiod fixes the issue.

Some resources seem to stay locked/leaked after `local_buffer_enabled_set()` fails and some cleanup isn't happening.

I'll dig quick into the kernel and come back.

commodo:
> > > There's a bug...
> > 
> > 
> > What are the effects of the bug? Does it crash?
> 
> Hmm, maybe I should describe it a bit better in the commit desc.
> Also, I'll admit, I may need to dig this into the kernel a bit.
> I was a bit happy to just use this hardened `local_open()` vs going deeper.
> 
> So, issue is: `local_open()` creates FD, `local_buffer_enabled_set()` fails [with -EINVAL], then on the next `local_open()` call, it returns -EBUSY (busy flag in the IIO device is still left as set), the IIO device is stuck as busy;
> The issue is with iiod. Since it's a process.
> Restarting iiod fixes the issue.
> 
> Some resources seem to stay locked/leaked after `local_buffer_enabled_set()` fails and some cleanup isn't happening.
> 
> I'll dig quick into the kernel and come back.

So, after some quick debug-prints into the kernel, what seems to be happening is that this code does not run:
https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/industrialio-core.c#L1619
Even though, iiod is calling close() on that FD [via exit-on-failure path in local_open()]; this `close()` that iiod is doing doesn't return an error though:
https://github.com/analogdevicesinc/libiio/blob/master/local.c#L1001

When I add this patchset, the chrdev_release seems to be.

I assumed that it may be a race-condition, and I added some sleep(1000)/wait [after the close()], it doesn't seem to be helping.
Maybe I forgot some pthread basics.

The local_close() though, seems to be doing some `munmap()` calls, which look like a good idea to be added in the exit-on-failure path.
I don't know if those are the ones helping, but there's also the matter of free-ing some malloc-ed data for high-speed buffers.

larsclausen:
> The local_close() though, seems to be doing some `munmap()` calls, which look like a good idea to be added in the exit-on-failure path.
> I don't know if those are the ones helping, but there's also the matter of free-ing some malloc-ed data for high-speed buffers.

A mapping on a file will keep a reference to that file. So this is probably the issue then.

rgetz:
so, in local, the only mmap is in `enable_high_speed()`: [link](https://github.com/analogdevicesinc/libiio/blob/master/local.c#L872-L894)

which should be released as part of `local_close()` [link](https://github.com/analogdevicesinc/libiio/blob/master/local.c#L1018-L1028) in the normal case.

when something calls `local_open()`, which eventually calls `enable_high_speed()`, there are some exit/error paths taken afterwards [here](https://github.com/analogdevicesinc/libiio/blob/master/local.c#L982) and [here](https://github.com/analogdevicesinc/libiio/blob/master/local.c#L992) (those two should not matter, since those are  only executed when `!pdata->is_high_speed`, but the latter [here](https://github.com/analogdevicesinc/libiio/blob/master/local.c#L997) that do not `munmap` (if `local_enable_buffer()` fails).

commodo:
Changelog v1 -> v2:
* added more info about iiod issue in commit comment

rgetz:
there are a few `close()` calls which now ignore the return code. From the man page:
`A careful programmer will check the return value of close(). a failure return should be used only for diagnostic purposes`

commodo:
> there are a few `close()` calls which now ignore the return code. From the man page:
> `A careful programmer will check the return value of close(). a failure return should be used only for diagnostic purposes`

yeah; i also thought about that for a bit when doing the initial change;
exiting there/early may prevent more cleanup from happening; so not exiting there, is a good idea;

i was wondering if printing the error code from `close()` makes sense; or is useful, or is just noise;
tbh, i don't know; i don't have any strong opinions about it;
any suggestion is welcome;
an error code doesn't seem to be coming through, even when close() doesn't actually free the FD [due to munmap not being called]

rgetz:
I think passing it up, and printing something might be worth it - then we can decide (based on user feedback) if it is a real problem or not. Won't be noisy, since I'm planning on starting the error message revamp in the next couple of weeks.

commodo:
> I think passing it up, and printing something might be worth it - then we can decide (based on user feedback) if it is a real problem or not. Won't be noisy, since I'm planning on starting the error message revamp in the next couple of weeks.

alright;
will return the error code of the first error; we can't print messages for other error codes;
there 2 close() calls, 1 ioctl() x number-of-buffers, 1 local_buffer_enabled_set() call, 1 channel_write_state() x  number-of-channels

commodo:
I added a separate commit for the error handling in local_close().

I am not sure, the code has much added value, vs it's quantity.
I think that a `void close()` would have simplified things [vs `int local_close()`]; the int return kind of makes you want to return something [other than 0], but I am not sure it's really useful for close().
A lot of the times, close() is ignored in many places i've seen it used [other than libiio].

Also, since many places could return an error code in `local_close()`, the whole logic of keeping only the first error code [when other places could fail as well], is ¯\_(ツ)_/¯
Also, returning the last error, is also not great.
I'm mostly feeling that we have other bigger bugs, issues and design flaws that we should address than this.

rgetz:
This looks good to me - thanks

as long as @larsclausen doesn't have any comments...

-Robin

larsclausen:
:+1:

rgetz:
this caused a warning in coverity...

1049 CID 361216 (#1 of 1): Logically dead code (DEADCODE)
dead_error_line: Execution cannot reach this statement

I think because, you are missing a `ret1 = ' on:
https://github.com/analogdevicesinc/libiio/blob/master/local.c#L1041

commodo:
> this caused a warning in coverity...
> 
> 1049 CID 361216 (#1 of 1): Logically dead code (DEADCODE)
> dead_error_line: Execution cannot reach this statement
> 
> I think because, you are missing a `ret1 =` on:
> https://github.com/analogdevicesinc/libiio/blob/master/local.c#L1041

yeah; it's a good catch on the side of coverity;
but it's also one reason why i did not really want to do patch  4b0a7a26182d86c97bd6de70d1530987e2416774
too much added complexity with not much added value; and difficult to review apparently

rgetz:
> but it's also one reason why i did not really want to do patch [4b0a7a2](https://github.com/analogdevicesinc/libiio/commit/4b0a7a26182d86c97bd6de70d1530987e2416774)
> too much added complexity with not much added value; and difficult to review apparently

The reason I asked for it - is there were error conditions on close() failing, that your previous patches removed. That is all. It's basically keeping those failing paths as they were - not adding more.

Issue #573, opened 2020-07-22
iio_attr cannot take negative numbers


tfcollins:
```
root@analog:~# iio_attr -o -c adrv9002-phy voltage0 hardwaregain
-20.000000 dB
root@analog:~# iio_attr -o -c adrv9002-phy voltage0 hardwaregain -10
iio_attr: invalid option -- '1'
Unknown argument '?'
root@analog:~# iio_attr -o -c adrv9002-phy voltage0 hardwaregain "-10"
iio_attr: invalid option -- '1'
Unknown argument '?'
root@analog:~# iio_attr -o -c adrv9002-phy voltage0 hardwaregain 0
0.000000 dB
root@analog:~# 

```
It does not know when to stop looking for flags.

Workaround:

```
root@analog:~# iio_attr -o -c -- adrv9002-phy voltage0 hardwaregain -10
-10.000000 dB
```

rgetz:
confirmed - and I have a patch I'm testing locally...

rgetz:
I made a PR #577 - so closing this for now.

PR #577, opened 2020-07-28
iio_attr: accept negative numbers as data to write to

rgetz:
getopt doesn't understand the difference between the '-' before an
option and the '-' before a negative number. It can not - they are both
the same, and in some applications option zero (-0) might be valid.

in iio_attr this causes problems when we try to write negative numbers

root@analog:~# iio_attr -o -c adrv9002-phy voltage0 hardwaregain -10
iio_attr: invalid option -- '1'

Since the only time we do this in in iio_attr, handle the last option
differently (if the last option is a negative number).

This does mean we still don't pass -foo to an attribute to be written,
since iio_attr thinks that is an unknown "-f" option.

./tests/iio_attr -v -a usb -o -c ad9361-phy voltage0 hardwaregain -foo
Using auto-detected IIO context at URI "usb:3.39.5"
./tests/iio_attr: invalid option -- 'f'

but this does fix the reported bug #573

./tests/iio_attr -v -a usb -o -c ad9361-phy voltage0 hardwaregain -1
Using auto-detected IIO context at URI "usb:3.39.5"
dev 'ad9361-phy', channel 'voltage0' (output), attr 'hardwaregain', value '-10.000000 dB'
wrote 3 bytes to hardwaregain
dev 'ad9361-phy', channel 'voltage0' (output), attr 'hardwaregain', value '-1.000000 dB'

### DIFF
--- a/tests/iio_attr.c
+++ b/tests/iio_attr.c
@@ -353,7 +353,7 @@ int main(int argc, char **argv)
 {
 	char **argw;
 	struct iio_context *ctx;
-	int c;
+	int c, argd = argc;
 	int device_index = 0, channel_index = 0, attr_index = 0;
 	const char *gen_file = NULL;
 	bool search_device = false, ignore_case = false,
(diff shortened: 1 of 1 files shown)


tests/iio_attr.c
```diff
@@ -371,13 +371,24 @@ int main(int argc, char **argv)
...
+	 * The only time we should see a negative number is the last argument during a write,
+	 * so if there is one, we skip that argument during getopt processing
+	 * look for "-" followed by a number.
+	 */
+	if (strnlen(argv[argc - 1], 2) >= 2 && argv[argc - 1][0] == '-' && 
+			(argv[argc - 1][1] >= '0' && argv[argc - 1][1] <= '9')) {
+		argd--;
+	}
+
+	ctx = handle_common_opts(MY_NAME, argd, argw, MY_OPTS, options, options_descriptions);
```
tfcollins, line 385: I guess the main assumption here is that no flags start with numbers?

rgetz:
this was a simple rebase on master, to make sure that the windows version works.

PR #580, opened 2020-07-31
tests/iio_common: Fix HEX value parsing in sanitize_clamp()

mhennerich:
Right now iio_reg and possibly other tools are broken.
Registers and values are typically provided in hex values.
strtoul() with base 10 returns error when hex values are parsed.
So iio_reg will always clear register 0 which in most cases will
reset the device.


rgetz:
strtoul with base 0 will accept decimal, hex (with leading `0x`) and octal (with leading `0).

```
printf("%lu\n", strtoul("014",NULL,0));
12
```

as long as that we what we want - that works. Just need to add that to the man pages. (and I agree - adding a `-v` option to `iio_reg` print out what it is doing would also be a good idea).

PR #589, opened 2020-09-16
network: Add consume to client ops

nunojsa:
The goal of this new callback is to consume all the __possible__ stalled data in the buffers before doing a new transaction. The problem can occur if a remote device is occupied doing something when a new transaction is started (eg: read an attribute). In this case, most likely, our read will timeout and, due to the nature of stream sockets, the next attribute we try to read will get the stalled value from the previous, failed, transaction.

The problem came up when using the adrv9002 plugin. Doing a manually update of the device profile while having an osc session opened lead to the plugin starting to display wrong values... And while this is a very specific error and is totally acceptable for the first timeout error, that should not mean that the following operations should get invalid data.

Although the solution has windows in mind it was not tested under it (but confirmed that it is also an issue there). It would be nice if someone that can easily build libiio and osc for windows could try this...

### DIFF
--- a/network.c
+++ b/network.c
@@ -1193,6 +1193,38 @@ static ssize_t network_read_data(struct iio_context_pdata *pdata,
 	return network_recv(io_ctx, dst, len, 0);
 }
 
+static int network_consume(struct iio_context_pdata *pdata)
+{
+	int ret;
+	char dummy[1024];
+
+	set_blocking_mode(pdata->io_ctx.fd, false);
+
+	do {
+		ret = recv(pdata->io_ctx.fd, dummy, sizeof(dummy), 0);
+		if (!ret) {
+			ret = -EPIPE;
+			break;
+		} else if (ret < 0) {
+			int err;
+
+			err = network_get_error();
+			if (network_is_interrupted(err))
+				continue;
+
+			/* if we get this it means there's no stalled data so we can proceed */
+			if (network_should_retry(err))
+				ret = 0;
+
+			break;
+		}
+	} while (1);
+
+	set_blocking_mode(pdata->io_ctx.fd, true);
+
+	return ret;
+}
+
 static ssize_t network_read_line(struct iio_context_pdata *pdata,
 		void *io_data, char *dst, size_t len)
 {
@@ -1265,6 +1297,7 @@ static const struct iiod_client_ops network_iiod_client_ops = {
 	.write = network_write_data,
 	.read = network_read_data,
 	.read_line = network_read_line,
+	.consume = network_consume,
 };
 
 #ifdef __linux__
--- a/iiod-client.c
+++ b/iiod-client.c
@@ -75,6 +75,12 @@ static int iiod_client_exec_command(struct iiod_client *client,
 	int resp;
 	ssize_t ret;
 
+	if (client->ops->consume) {
+		ret = client->ops->consume(client->pdata);
+		if (ret)
+			return (int) ret;
+	}
+
 	ret = client->ops->write(client->pdata, desc, cmd, strlen(cmd));
 	if (ret < 0)
 		return (int) ret;
--- a/iiod-client.h
+++ b/iiod-client.h
@@ -32,6 +32,7 @@ struct iiod_client_ops {
 			void *desc, char *dst, size_t len);
 	ssize_t (*read_line)(struct iio_context_pdata *pdata,
 			void *desc, char *dst, size_t len);
+	int (*consume)(struct iio_context_pdata *pdata);
 };
 
 struct iiod_client * iiod_client_new(struct iio_context_pdata *pdata,


network.c
```diff
@@ -1193,6 +1193,38 @@ static ssize_t network_read_data(struct iio_context_pdata *pdata,
 	return network_recv(io_ctx, dst, len, 0);
 }
 
+static int network_consume(struct iio_context_pdata *pdata)
+{
+	int ret;
+	char dummy[1024];
+
+	set_blocking_mode(pdata->io_ctx.fd, false);
```
commodo, line 1201: i know there is some precedence in network.c for this `set_blocking_mode(false) + set_blocking_mode(true)`, but it is a bit weird;

network.c
```diff
@@ -1193,6 +1193,38 @@ static ssize_t network_read_data(struct iio_context_pdata *pdata,
 	return network_recv(io_ctx, dst, len, 0);
 }
 
+static int network_consume(struct iio_context_pdata *pdata)
+{
+	int ret;
+	char dummy[1024];
```
commodo, line 1199: maybe we could reduce the buffer size to 128 or 64 bytes?
it means more reads, but 1k on the stack can be a lot;
some libcs limit the pthread stack size to 128k, so 1k can be a lot when writing an app with libiio;

network.c
nunojsa: I think the main reason for this is to also work with windows. Initially, I was just passing `MSG_DONTWAIT` to `recv` but then I looked for this in windows and it seems to be not supported. `set_blocking_mode`will actually have another implementation on windows...

commodo:
> The problem came up when using the adrv9002 plugin. Doing a manually update of the device profile while having an osc session opened lead to the plugin starting to display wrong values... And while this is a very specific error and is totally acceptable for the first timeout error, that should not mean that the following operations should get invalid data.

libiio has a few design limitations; especially with high-speed ADCs/DACs; we kind of know this;
we keep adding faster and more complicated transceiver on top of a library (libiio) which didn't get any polishes/design-updates to handle these;
one idea was to add new mechanism for these fast converters; something like: start-stream, stop-stream; this would go in parallel than the control sockets;
currently, libiio is limited to a ping-pong of "give me 1000 samples, iiod provides 1000 samples, and so on", which is expensive over the network [obviously];

so, this feels like a workaround;
if this issue is less notice-able on systems with a higher system/CPU speed, or faster networks, then it also has some race-condition connotations;

if we could push the issue to the 9002 plugin for now, it would be great;
that is, until we improve libiio to better handle this;

network.c
nunojsa: I guess I can reduce it for 512 but nowadays 1K should be more than acceptable. For instance on my system I have 8192k for the stack size by default. Even 32bit kernel (where stack occupation is really more "dangerous") have 1024 set for `Wframe-larger-than`

commodo: glibc allocates 2M, 4M or 8M;
musl-libc 128k
uclibc-ng - don't know

it very much depends on which libc is used

nunojsa:
> if we could push the issue to the 9002 plugin for now, it would be great;

I have to disagree here. Fixing the plugin feels like a workaround :smile: 

> currently, libiio is limited to a ping-pong of "give me 1000 samples, iiod provides 1000 samples, and so on", which is expensive over the network [obviously];

This problem don't show up only when you have a capture on. It just messes with your displayed values if it happens (the ones that you need to keep updating). Also this is nothing to do with internet speed... This is just how stream sockets work. Let me try to explain what is going on better. Imagine the following

1. On the remote end, start some time consuming operation on the device (on our transceivers this would be updating a profile). This will hold the device mutex most likely;
2. Osc, on your local machine is continuously updating some values on a plugin (on adrv9002 these could be RSSI or decimated power);
3. At some point it will timeout because it wont get any response since iiod is probably waiting for the mutex to be released on the remote side;
4. After the mutex is released on the remote machine, iiod will send the response back to you.
5. Next time osc tries to read an attr and calls `recv`, it will get this stalled data from the previous transaction...
6. At this point, the plugin values that are continuously updated will show wrong values and if I click the "Reload Settings" button, then It will even apply just some wrong/random values to the device...

So in the end, the problem is just that you go "out of sync" with your "server". Having some kind of protocol on libiio with some sequence counter or something like could also be something to think but It adds more complexity and it feels to much.

The idea in the patch is just to make sure that we don't have any data left to consume before starting a new `command` and so, even if we go out of sync, eventually we stabilize the connection again...

network.c
nunojsa: yes, I know... But even 128k is more than enough I guess. If an app is using the other 127K then I would argue that it's more on the app side than on the libiio. This feels like being really really cautious :smile: . But no strong feeling, I can reduce it to 256 as a compromise...

commodo: well, the discussion is more like: someone will write a bad-app [anyway, it's pretty unavoidable], which uses plenty of stack, and it blows up in libiio and it ends up being on EZ [because that's where the stack blew up];
while, there are chances of happening with 256 / 128 / 64 bytes on the stack, i think it would be a higher chance with 1k than with less;

nunojsa: Eheheh, you want to avoid EZ at all cost :smile:. I will jut give a bit more time to check if some other requests come up and then I will send a v2 with 256

rgetz:
> libiio has a few design limitations; especially with high-speed ADCs/DACs; we kind of know this;

And 9002 is actually pretty slow. This doesn't have anything to do with speed - this is - if you poke the wrong attribute the device basically locks up for 30 seconds (or longer) and in transactions, or new attribute reads fail...

It's 9002 specific - devices shouldn't be unresponsive that long, but it is. change the sample rate - wait 30 seconds. change bandwidth - wait 30 seconds. that is how the device is architected. if a different thread is doing a read during this time, or async to this time - it's going to fail.

commodo:
> > libiio has a few design limitations; especially with high-speed ADCs/DACs; we kind of know this;
> 
> And 9002 is actually pretty slow. This doesn't have anything to do with speed - this is - if you poke the wrong attribute the device basically locks up for 30 seconds (or longer) and in transactions, or new attribute reads fail...
> 
> It's 9002 specific - devices shouldn't be unresponsive that long, but it is. change the sample rate - wait 30 seconds. change bandwidth - wait 30 seconds. that is how the device is architected. if a different thread is doing a read during this time, or async to this time - it's going to fail.

ok, but anyway you cut it, doing a dummy-read and throwing away the data looks fishy;
it shows that there may be a design limitation in iiod or libiio, or osc; and that this looks more like a band-aid;
it would be like a web-server not sending you parts of a web-page because something locked-up;

now one way to do it may be with implementing asynchronous-sockets using `select()` ;
this works on Windows/Mac/Linux reasonably well, even with all the minor interoperability issues

rgetz:
> Although the solution has windows in mind it was not tested under it (but confirmed that it is also an issue there). It would be nice if someone that can easily build libiio and osc for windows could try this...

You should be able to install libiio into the osc directory, and it will use that one.
https://ci.appveyor.com/project/analogdevicesinc/libiio/builds/35237722/artifacts

rgetz:
> Doing a manually update of the device profile while having an osc session opened lead to the plugin starting to display wrong values...

I'm not sure what you mean by "manual update". 

For devices that have profiles, I think there is a certain order of operations that needs to be respected by the person between the chair and the keyboard, and libiio - I don't think it will work without... @mhennerich can comment more.

mhennerich:
> > Doing a manually update of the device profile while having an osc session opened lead to the plugin starting to display wrong values...
> 
> I'm not sure what you mean by "manual update".
> 
> For devices that have profiles, I think there is a certain order of operations that needs to be respected by the person between the chair and the keyboard, and libiio - I don't think it will work without... @mhennerich can comment more.

I think manual update means clicking the refresh button in the plugin. The issue is for sure caused by the long blocking write for the profile_config load. While this is going on the plugin keeps polling some attributes. This in return is queuing up, and causes iiod to get out of sync. The workaround is to flush the sockets. We should avoid this condition, by pausing the poll thread in osc.
However it's a known deficiency of libiio and having a recovery mechanism is a good thing to have.

nunojsa:
So, the manual update means doing a profile update on the device (using serial or ssh) with `echo some_profile > profile_config`. This will held the device locked up for some time while osc is polling to get readings about RSSI and decimated power. This will trigger the issue as @mhennerich explains:

>... While this is going on the plugin keeps polling some attributes. This in return is queuing up, and causes iiod to get out of sync."

So, again, this special error has nothing to do with internet speed. We can have the fastest link but as a long as the device mutex is held, that won't really matter...

I know that this is a very specific corner case and as I said, getting a timeout because the device is doing something else is not really a bug, my point is that the following `commands` (reading an attribute for example) should not get stalled/wrong data...

nunojsa:
> ok, but anyway you cut it, doing a dummy-read and throwing away the data looks fishy;
it shows that there may be a design limitation in iiod or libiio, or osc; and that this looks more like a band-aid;

I get your point here. One other solution would be to have some kind of protocol implementation where we could make sure that the reply we got is really for the command we sent. But that also feels to much of overhead.

AFAICT, as long as this send command -> get data is synchronous (and it looks it is since we held the client mutex), we should be fine in assuming that if we have any data on the socket before starting a new command, then it's stalled data.

But I also don't really know the internals of libiio, so if someone sees that this can lead to some race where we discard data that we should not...

rgetz:
> I get your point here. One other solution would be to have some kind of protocol implementation where we could make sure that the reply we got is really for the command we sent. But that also feels to much of overhead.

Yes - there is no reset, or sync, to recover after a timeout. (I would be curious how this manifest itself on usb, or the other backends)...

nunojsa:
> I would be curious how this manifest itself on usb, or the other backends

That also came to my mind. I would suspect that this can also manifest in serial unless libserial is doing something special...

nunojsa:
@pcercuei, I guess this one can be closed as it might be already (or be in the future) addressed in the dev branch...

pcercuei:
@nunojsa It is not yet addressed in the dev branch, but it is addressed  by the async protocol which I'm working on right now.
I'll close then.

PR #590, opened 2020-09-16
utilities: Fix getenv function

nunojsa:
This patch fixes a wrong comparison leading to a false failure. We want to
return error if the env var size is bigger than max size and not the other
way around.

Signed-off-by: Nuno Sá <nuno.sa@analog.com>

### DIFF
--- a/utilities.c
+++ b/utilities.c
@@ -301,7 +301,7 @@ char * iio_getenv (char * envvar)
 	len = strnlen(hostname, tmp);
 
 	/* Should be smaller than max length */
-	if (len <= tmp)
+	if (len == tmp)
 		goto wrong_str;
 
 	/* should be more than "usb:" or "ip:" */


utilities.c
```diff
@@ -301,7 +301,7 @@ char * iio_getenv (char * envvar)
 	len = strnlen(hostname, tmp);
 
 	/* Should be smaller than max length */
-	if (len <= tmp)
+	if (len > tmp)
```
commodo, line 304: should this be `len >= tmp`?
`len > tmp` should be an impossible condition

nunojsa: Oh yeah, you are right. It could even be `len == tmp` since it should never be `>` than `tmp`

commodo: yep; `len == tmp` also works

commodo:
only the commit comment requires a bit of update now;
rest looks good

commodo (approved):

nunojsa:
Changes in v2:

- Make the comparison `len == tmp`, since `strnlen` already guarantees that `len > tmp` is not possible

dNechita (approved):

rgetz (approved):

PR #595, opened 2020-10-06
CMake: fix build with cmake 2.8.7

DanielGuramulta:
Building libiio with cmake 2.8.7(minimum version required) would fail due to cmake(2.8.7) not being able to copy a list of files separated by ";" to a destination. A solution to this problem is to create a list of commands containing a copy command for each iio test target file. With this approach the minimum required CMake version is not touched.

Tested with CMake version 2.8.7 and 3.18.2

Signed-off-by: DanielGuramulta <Daniel.Guramulta@analog.com>


dNechita (approved):
This seems to be ok. Thanks!

rgetz (approved):

PR #596, opened 2020-10-07
local.c: when target buffer is too small, fail

### LINKED ISSUE #357: Different behaviour (local vs remote) when an iio attribute read causes buffer overflow
Suppose one wants to read an iio attribute but doesn't provide enough space for the buffer where the content of the attribute will be read:
 - when local backend is used, the read operation is successful and the content of the attribute is truncated to fit the size of the buffer
 - when a backend that communicates with IIOD is used (e.g. usb, ethernet), the read operation fails with ERROR: Input/output error (-5).

This inconsistency has been found while investigating bug: #268

rgetz:
Fix #357

If a target buffer is too small, right now, we try to put a partial fill
in the buffer, and return sucess (which is wrong).

Now we will return EFBIG, and leave the target buffer empty. This will
make the local backends the same as the otherbackends.

### DIFF
--- a/local.c
+++ b/local.c
@@ -706,10 +706,16 @@ static ssize_t local_read_dev_attr(const struct iio_device *dev,
 		return -errno;
 
 	ret = fread(dst, 1, len, f);
+
+	/* if we didn't read the entire file, fail */
+	if (!feof(f))
+		ret = -EFBIG;
+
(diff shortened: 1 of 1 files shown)


dNechita (approved):
Looks good.

PR #597, opened 2020-10-08
iio: context: start hiding the iio_context object inside context.c

commodo:
This PR is trying to reboot this PR https://github.com/analogdevicesinc/libiio/pull/9
Which is essentially a modularization of libiio to be able to add more backends in a more dynamic fashion.

The intent is to be able to add more backends (like GPIO, HWmon, and other stuff).
But to get there, ideally we'd need to make libiio a bit more modular, and first convert all current backends to be made modular (independent of internal structures like iio_context, iio_device, etc).
We don't need to split libiio.so in smaller libs [if we don't want to]; we can implement this option to allow libiio to be built as a single library or as a smaller set of libraries, one being libiio-core.so and the others being libiio-<backend>.so (which would be loaded by libiio-core.so).
But if adding more backends, at least the internal structures/types need to be made completely opaque to the IIO backends, to make the scaling less error-prone.

We've seen various attempts in Linux at squeezing stuff in the IIO kernel framework. 
The gatekeeper there is upstream, so it takes a while to be able to do these things.
The reason seems to be network communication aspect, which other libraries don't seem to have (hwmon has lm-sensors, libiio has some libgpio, [all] with no network support).
In libiio we have way more flexibility to start integrating various backends.

### DIFF
--- a/context.c
+++ b/context.c
@@ -443,6 +443,22 @@ const char * iio_context_get_attr_value(
 	return NULL;
 }
 
+int iio_context_add_device(struct iio_context *ctx, struct iio_device *dev)
+{
+	struct iio_device **devices = realloc(ctx->devices,
+			(ctx->nb_devices + 1) * sizeof(struct iio_device *));
+
+	if (!devices) {
+		IIO_ERROR("Unable to allocate memory\n");
+		return -ENOMEM;
+	}
+
+	devices[ctx->nb_devices++] = dev;
+	ctx->devices = devices;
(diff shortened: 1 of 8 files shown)


dNechita (approved):
These changes seem harmless. Also, they remove some duplicate code. And yes, makes things easier for the idea of modular backends.

tfcollins (approved):
Agreed with @dNechita , this is mostly some housekeeping ahead of the flood.

mirceacaprioru (approved):

Issue #591, opened 2020-09-17
 Reproducible Builds


rgetz:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=970498

tfcollins:
Anyway for us to test this?

rgetz:
make two builds - make sure they are identical ?

mhennerich:
@dNechita  can you apply the patch from here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=970498

and test reproducibility?

dNechita:
Yes, I will test.

dNechita:
I've tested with the changes in the patch
 What the patch does is that allows setting the date during the cmake step. So, if you want to reproduce an identical build with a yesterday build, all you need is to set an environment variable with the yesterday date and then do the build.
Then, the man pages will be identical.

dNechita:
The idea is to add the changes in the patch to libiio, right?
If so, I'm not sure how to apply the patch (which seems to be a debian diff sort of thing) in order the preserve the original author. Any suggestions?

rgetz:
I think it is just : 
`git commit --author="Author Name <email@address.com>"`

but you need their permission first. The DCO is per author, and we won't accept anything without a signed-off-by by the original author.

-Robin

lamby:
Added PR. :+1:

rgetz:
Thanks for the PR

PR #598, opened 2020-10-10
Make the buld reproducible

lamby:
Whilst working on the [Reproducible Builds effort](https://reproducible-builds.org/) we noticed that `libiio` could not be built reproducibly. This is because it uses the current build time in the generated manpages. Patch attached that uses [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/). This PR addresses #591.


commodo:
Thanks for the patch

2 things:
1. We’d need a Signed-off-by to the commit
2. Typo buld -> build

lamby:
Updated.

commodo:
Many thanks :)

commodo (approved):

PR #603, opened 2020-10-14
iio: start splitting iio_backend.h (with iio_backend objects)

commodo:
This changeset starts to split some bits away from libiio core to add support for more modular backends.
The changes are more cleanups at this point, but they pave the way to having current IIO backends more independent of the IIO core logic.


context.c
```diff
@@ -182,6 +182,51 @@ char * iio_context_create_xml(const struct iio_context *ctx)
...
+			goto err_free_pdata;
+	}
+
+	ctx->name = backend->name;
+	ctx->ops = backend->ops;
+
+	return ctx;
+
+err_free_pdata:
+	if (ctx->pdata)
```
nunojsa, line 222: nit: Do we need the `if()` here? I believe `free()` will just be be a `no op` if the ptr is NULL

nunojsa (approved):

context.c
commodo: i don't know;
there used to be a time when free(NULL) wasn't covered by all libc implementations;
[ i have memories of doing free(NULL) and crashing ]
i don't know how the case is today;

a lot of libiio free()s seem to be NULL checked;
so   ¯\_(ツ)_/¯

nunojsa: I would be surprised if it's still the case that `free(NULL)` crashes. But I have no idea how that works on windows for example...

> a lot of libiio free()s seem to be NULL checked;

fair enough...

commodo: there's also the case that you can't predict what libc will be used [for libiio];
some may use something proprietary libcs [they may be rare];
that's why kernel is nice; the allocation and everything is self contained;

PR #605, opened 2020-10-14
iio: backend: add get_description() backend op & hook it into backends

commodo:
This doesn't get rid of the 'ctx->description' object; that will still be
around and cache the description.

The get_description() splits the creation logic of the description into a
backend op. That way, we can think of splitting the creation of an IIO
context into an 'alloc()' and 'init()' part where:
* an IIO context is allocated first in libiio core
* it's backend init() routine is called
* it's backend get_description() routine is called

The idea of the 'get_description()' op is to return a pointer, that libiio
core will manage/free. Hence the return is 'char *' vs 'const char *'.

### DIFF
--- a/iio-backend.h
+++ b/iio-backend.h
@@ -70,6 +70,8 @@ struct iio_backend_ops {
 
 	void (*shutdown)(struct iio_context *ctx);
 
+	char * (*get_description)(const struct iio_context *ctx);
+
 	int (*get_version)(const struct iio_context *ctx, unsigned int *major,
 			unsigned int *minor, char git_tag[8]);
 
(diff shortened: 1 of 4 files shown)


commodo:
oh, this actually needed PR #604 to be merged first;

dNechita (approved):

commodo:
The `uri_len = strlen(description)` that Codacy throws is a bit annoying to fix at this point in time.
We could think of a better way to do it.

But, there are plenty of strlen() uses.

In any case, I've re-organized the commits to be 4 [instead of 5].
The final code is the same as before.

PR #608, opened 2020-10-17
ci,travis: remove lsb package requirement

commodo:
To get the lsb_release command working some packages need to be
pre-installed.
The docker images come pretty bare, so they need to be installed before.

This can complicate things, because we need to make sure that all builds
down the pipeline install this package.

The utilities that wrap lsb_release can be re-implemented with simple shell
commands and parsing of /etc/os-release.
This change does that.


dNechita:
Will this re-implementation be as generic as using the lsb_release tool? (will it work on all Linux distributions that we need it to work and be consistent on the format of the result?)

commodo:
My understanding is that lsb_release will process /etc/os-release and /etc/lsb-release

But os-release is the standard 
https://stackoverflow.com/questions/47838800/etc-lsb-release-vs-etc-os-release

dNechita (approved):

PR #610, opened 2020-10-19
.travis.yml: unify EXTRA_SSH env var

commodo:
This has mostly cleanup value. Since it's duplicated everywhere, it may
make sense to just move the EXTRA_SSH to a global place, and not duplicate
it anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python: "3.6"
 
 env:
   global:
+  - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
   # SSHHOST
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -23,13 +24,11 @@ matrix:
...
-        - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
     - compiler: "gcc"
       dist: xenial
       env:
         - ARCH=arm
         - OS_VERSION=stretch
-        - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
     - compiler: "gcc"
       os: linux
       env:
```
dNechita, line 34: EXTRA_SSH will be set for this configuration. And the next one (with clang). Before these 2 configurations did not have EXTRA_SSH set to anything, now it will be set. Just wondering if this might cause a problem.

commodo: it won't [as discussed offline]

dNechita (approved):

Issue #611, opened 2020-10-19
Eliminate variable shadowing in local.c 


msporcich:
In local.c, there are two instances of variable shadowing which cause the build to fail when compiling with flag 
`-Werror=shadow 
`

In local_close()
`  error: declaration of ‘i’ shadows a previous local [-Werror=shadow]
`
In get_short_attr_name()
`error: declaration of ‘len’ shadows a previous local [-Werror=shadow]
`

msporcich:
Forked with a proposed resolution.  I'm interested in getting any comments and then signoff so that I may issue a pull request.  Thanks.

commodo:
the change looks reasonable to me
feel free do a PR targeting our master;

thanks
Alex

rgetz:
Only question is should we add `-Werror=shadow` all the time? (for gcc)

Nominally it's `-Wall -Wextra -Wno-unused-parameter -Wpedantic -std=c99`

rgetz:
closing this.

PR #612, opened 2020-10-20
#611:Eliminate variable shadowing in local.c

msporcich:
Resolves build problems when compiling with -Werror=shadow.

Signed-off-by: [Marc Sporcich] <marc@epiqsolutions.com>

### DIFF
--- a/local.c
+++ b/local.c
@@ -1012,7 +1012,6 @@ static int local_close(const struct iio_device *dev)
 	ret = 0;
 	ret1 = 0;
 	if (pdata->is_high_speed) {
-		unsigned int i;
 		if (pdata->addrs) {
 			for (i = 0; i < pdata->allocated_nb_blocks; i++)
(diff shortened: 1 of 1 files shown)


local.c
```diff
@@ -1198,7 +1197,7 @@ static char * get_short_attr_name(struct iio_channel *chn, const char *attr)
 		ptr += len + 1;
 
 	if (chn->name) {
-		size_t len = strlen(chn->name);
+		len = strlen(chn->name);
```
commodo, line 1200: the commit also needs a Signed-off-by tag;

commodo: and preferably the same description as in the PR;
Github PRs don't track too well in git history; so the information in the PR description is sightly more useful in the git history to be able to view it with `git log` on the command line

msporcich: Ah yes, understood.

commodo (approved):

commodo:
Thanks :)

PR #614, opened 2020-10-22
Dev/enhance network context creation

nunojsa:
The first patch is a simple leak fix on the iio mutex. The second patch improves the network context creation by adding a fallback mechanism to resolve a hostname. Basically, if the given hostname is an avahi typical one (eg: analog.local), `getaddrinfo()` will only work if `nss-mdns` is installed on the host and `/etc/nsswitch.conf` is correctly configured. While that might be true for most linux distros, it is still something that we should not rely on. Hence, with this patch, if `getaddrinfo()` fails, we still try to resolve the host by directly using libavahi.

I was actually caught by this issue when trying to use pyadi-iio in my archlinux machine.

Also note that this is only targeting builds with `HAVE_AVAHI` since I have no idea if this is actually an issue on Windows or Mac (most likely it isn't)

### DIFF
--- a/dns_sd_avahi.c
+++ b/dns_sd_avahi.c
@@ -131,6 +131,28 @@ static void __avahi_resolver_cb(AvahiServiceResolver *resolver,
 	avahi_service_resolver_free(resolver);
 }
 
+static void avahi_host_resolver(AvahiHostNameResolver *resolver, __notused AvahiIfIndex iface,
+				__notused AvahiProtocol proto, AvahiResolverEvent event,
+				const char *host_name, const AvahiAddress *address,
+				__notused AvahiLookupResultFlags flags, void *d)
+{
+	struct dns_sd_discovery_data *ddata = (struct dns_sd_discovery_data *) d;
+
+	switch(event) {
+	case AVAHI_RESOLVER_FAILURE:
+		IIO_ERROR("Avahi Resolver: Failed to resolve host '%s' : %s\n", host_name,
+			avahi_strerror(avahi_client_errno(
+				avahi_host_name_resolver_get_client(resolver))));
+		break;
+	case AVAHI_RESOLVER_FOUND:
+		avahi_process_resolved(ddata, address, host_name, IIOD_PORT);
+		break;
+	}
+
+	avahi_host_name_resolver_free(resolver);
+	avahi_simple_poll_quit(ddata->poll);
+}
+
 static void __avahi_browser_cb(AvahiServiceBrowser *browser,
 		AvahiIfIndex iface, AvahiProtocol proto,
 		AvahiBrowserEvent event, const char *name,
@@ -256,5 +278,90 @@ err_free_poll:
 	iio_mutex_destroy(d->lock);
 	*ddata = d;
 
+	return ret;
+}
+
+static void avahi_resolve_host(struct dns_sd_discovery_data *d, const char *hostname,
+			       const AvahiProtocol proto)
+{
+	AvahiClient *client;
+	AvahiHostNameResolver *resolver;
+	int ret;
+
+	d->poll = avahi_simple_poll_new();
+	if (!d->poll)
+		return;
+
+	client = avahi_client_new(avahi_simple_poll_get(d->poll), 0, NULL, NULL, &ret);
+	if (!client) {
+		IIO_ERROR("Unable to create Avahi DNS-SD client :%s\n", avahi_strerror(ret));
+		goto err_free_poll;
+	}
+
+	resolver = avahi_host_name_resolver_new(client, AVAHI_IF_UNSPEC, proto, hostname, proto, 0,
+						avahi_host_resolver, d);
+	if (!resolver) {
+		ret = avahi_client_errno(client);
+		IIO_ERROR("Unable to create Avahi DNS-SD browser: %s\n", avahi_strerror(ret));
+		goto err_free_client;
+	}
+
+	IIO_DEBUG("Trying to resolve host: %s\n, proto: %d", hostname, proto);
+	avahi_simple_poll_loop(d->poll);
+
+err_free_client:
+	avahi_client_free(client);
+err_free_poll:
+	avahi_simple_poll_free(d->poll);
+}
+
+int dnssd_resolve_host(const char *hostname, char *ip_addr, const int addr_len)
+{
+	struct dns_sd_discovery_data *d;
+	int ret = 0;
+
+	if (!hostname || hostname[0] == '\0')
+		return -EINVAL;
+
+	if (new_discovery_data(&d) < 0)
+		return -ENOMEM;
+
+	d->lock = iio_mutex_create();
+	if (!d->lock) {
+		ret = -ENOMEM;
+		goto err_free_data;
+	}
+	/*
+	 * The reason not to use AVAHI_PROTO_UNSPEC is that avahi sometimes resolves the host
+	 * to an ipv6 link local address which is not suitable to be used by connect. In fact,
+	 * `port_knock_discovery_data()` would discard this entry. On the other hand, some users
+	 * might really want to use ipv6 and have their environment correctly configured. Hence,
+	 * we try to resolve both in ipv4 and ipv6...
+	 */
+	avahi_resolve_host(d, hostname, AVAHI_PROTO_INET);
+#ifdef HAVE_IPV6
+	avahi_resolve_host(d, hostname, AVAHI_PROTO_INET6);
+#endif
+
+	if (d->resolved) {
+		port_knock_discovery_data(&d);
+		remove_dup_discovery_data(&d);
+	} else {
+		ret = -ENXIO;
+		goto err_mutex_destroy;
+	}
+
+	/* If next is null it means that d is empty */
+	if (!d->next) {
+		ret = -ENXIO;
+		goto err_mutex_destroy;
+	}
+
+	iio_strlcpy(ip_addr, d->addr_str, addr_len);
+
+err_mutex_destroy:
+	iio_mutex_destroy(d->lock);
+err_free_data:
+	dnssd_free_all_discovery_data(d);
 	return ret;
 }
--- a/dns_sd_avahi.c
+++ b/dns_sd_avahi.c
@@ -211,6 +211,7 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 
 	d->poll = avahi_simple_poll_new();
 	if (!d->poll) {
+		iio_mutex_destroy(d->lock);
 		dnssd_free_all_discovery_data(d);
 		return -ENOMEM;
 	}
--- a/dns_sd_avahi.c
+++ b/dns_sd_avahi.c
@@ -58,6 +58,38 @@ void dnssd_free_discovery_data(struct dns_sd_discovery_data *d)
 	free(d);
 }
 
+static void avahi_process_resolved(struct dns_sd_discovery_data *ddata, const AvahiAddress *addr,
+				   const char *host_name, const uint16_t port)
+{
+       /* Avahi is multi-threaded, so lock the list */
+       iio_mutex_lock(ddata->lock);
+       ddata->resolved++;
+
+       /* Find empty data to store things*/
+       while (ddata->next)
+               ddata = ddata->next;
+
+       /* link a new placeholder to the list */
+       avahi_address_snprint(ddata->addr_str,
+                       sizeof(ddata->addr_str), addr);
+       memcpy(ddata->address, addr, sizeof(*ddata->address)); /* Flawfinder: ignore */
+       ddata->port = port;
+       ddata->hostname = strdup(host_name);
+       ddata->resolved = true;
+       /* link a new, empty placeholder to the list */
+       if (!new_discovery_data(&ddata->next)) {
+               /* duplicate poll & lock info,
+                * since we don't know which might be discarded */
+               ddata->next->poll = ddata->poll;
+               ddata->next->lock = ddata->lock;
+       } else {
+               IIO_ERROR("Avahi Resolver : memory failure\n");
+       }
+       iio_mutex_unlock(ddata->lock);
+
+       IIO_DEBUG("\t\t%s:%u (%s)\n", host_name, port, ddata->addr_str);
+}
+
 /*
  * libavahi callbacks for browser and resolver
  * for more info, check out libavahi docs at:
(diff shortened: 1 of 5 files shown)


rgetz:
a scan context always includes both the name and the IP number - broken distributions can use IP numbers from the scan context.

This is what we do inside osc (specifically for windows - since it has a similar issue).
https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/dialogs.c#L391-L402

network.c
```diff
@@ -1402,6 +1402,30 @@ struct iio_context * network_create_context(const char *host)
 #endif
 	{
 		ret = getaddrinfo(host, IIOD_PORT_STR, &hints, &res);
+#ifdef HAVE_AVAHI
+		/*
```
rgetz, line 1406: to keep network.c free of #ifdef HAVE_... should we just add a dnssd_resolve_host() function for all dnssd implementations (which is null is some) and just make everything common in network.c?

dns_sd_avahi.c
```diff
@@ -60,6 +60,39 @@ void dnssd_free_discovery_data(struct dns_sd_discovery_data *d)
...
+       ddata->resolved++;
+
+       /* Find empty data to store things*/
+       while (ddata->next)
+               ddata = ddata->next;
+
+       /* link a new placeholder to the list */
+       avahi_address_snprint(ddata->addr_str,
+                       sizeof(ddata->addr_str), addr);
+       memcpy(ddata->address, addr, sizeof(*addr));
```
rgetz, line 77: This throws a warning in https://app.codacy.com/gh/analogdevicesinc/libiio/pullRequest?prid=6384953 since we no longer know the length of these...

dns_sd_avahi.c
```diff
@@ -213,6 +241,7 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 
 	d->poll = avahi_simple_poll_new();
 	if (!d->poll) {
+		iio_mutex_destroy(d->lock);
```
rgetz, line 244: Thanks for fix.

dns_sd_avahi.c
nunojsa: Will look at this. Honestly that code was just copy past

network.c
nunojsa: That looks good to me...

nunojsa:
> a scan context always includes both the name and the IP number - broken distributions can use IP numbers from the scan context.

So pyadi-iio ends up calling `iio_create_context_from_uri()` with a hostname like "analog.local" which is fine. It just depends on your distribution. Mainly if you the right configuration in nsswitch and the right package installed. And I actually have to confirm this, but I'm almost positive the automatic scan in osc was also broken in my machine and now works without me having to install  `nss-mdns`...

dns_sd_avahi.c
nunojsa: Hmm, maybe I'm missing something oblivious, but I'm not getting this warning. `const AvahiAddress *addr` and `ddata->address` are pointers to the same data type so I'm not really seeing the issue here... I mean, I get the warning can be useful to double check this line (and potentially find a bug) but I'm not really seeing any issue with this particular line of code... Maybe if I change the pattern to `memcpy(ddata->address, addr, sizeof(*ddata->address)`, it shuts up the warning...

nunojsa: So, this warning is still coming from codacy and I don't think there's much we can do about it (as the code stands). Alternatively, we can either just get rid of this line and the `address` variable (it looks is not really used) or make `address` a plain variable rather than a pointer...

nunojsa:
Changes in V2:

1. Add implementations of `dnssd_resolve_host` in all libiio dnssd alternatives so that we keep `network.c` without nee ifdefery...

dns_sd_avahi.c
```diff
@@ -60,6 +60,39 @@ void dnssd_free_discovery_data(struct dns_sd_discovery_data *d)
...
+               ddata->next->lock = ddata->lock;
+       } else {
+               IIO_ERROR("Avahi Resolver : memory failure\n");
+       }
+       iio_mutex_unlock(ddata->lock);
+
+       IIO_DEBUG("\t\t%s:%u (%s)\n", host_name, port, ddata->addr_str);
+}
+
+
```
commodo, line 95: nitpick: extra line

dns_sd_bonjour.c
```diff
@@ -78,12 +78,12 @@ static void __cfnet_browser_cb (
 		IIO_DEBUG("DNS SD: Net service is null.\n");
 		goto verify_flags;
 	}
-	
+
```
commodo, line 81: whitespace change could be it's own commit

dns_sd_avahi.c
```diff
@@ -92,42 +125,37 @@ static void __avahi_resolver_cb(AvahiServiceResolver *resolver,
...
-		if (!new_discovery_data(&ddata->next)) {
-			/* duplicate poll & lock info,
-			 * since we don't know which might be discarded */
-			ddata->next->poll = ddata->poll;
-			ddata->next->lock = ddata->lock;
-		} else {
-			IIO_ERROR("Avahi Resolver : memory failure\n");
-		}
-		iio_mutex_unlock(ddata->lock);
-
```
commodo, line 120: this move looks like it's own commit

network.c
```diff
@@ -1402,6 +1402,29 @@ struct iio_context * network_create_context(const char *host)
...
+		 * as a last resort, let's try to resolve the host with avahi...
+		 */
+		if (ret) {
+			char addr_str[DNS_SD_ADDRESS_STR_MAX];
+
+			IIO_DEBUG("'getaddrinfo()' failed: %s. Trying dnssd as a last resort...\n",
+				  gai_strerror(ret));
+			ret = dnssd_resolve_host(host, addr_str, sizeof(addr_str));
+			if (ret) {
+				char buf[512];
```
commodo, line 1418: 512 bytes seems a lot for an error message; 128 or 256?

dns_sd_avahi.c
```diff
@@ -255,3 +283,88 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
...
+
+int dnssd_resolve_host(const char *hostname, char *ip_addr, const int addr_len)
+{
+	struct dns_sd_discovery_data *d;
+	int ret = 0;
+
+	if (!hostname || hostname[0] == '\0')
+		return -EINVAL;
+
+	if (new_discovery_data(&d) < 0)
```
commodo, line 329: i didn't follow the rest of the flow yet; but does `new_discovery_data()` require to be under lock?
or is this in a safe context to let it be unlocked?

commodo: ah wait, stuff gets allocated there;
nevermind

dns_sd_bonjour.c
nunojsa: well, that is my editor doing this automatically. If I have to come with a v3 I will make this changes in vim so I don't get this automatically done

network.c
nunojsa: I'm already reducing it... it's 1024 some lines above this. But yeah, most likely we don't really need 512 bytes

dns_sd_avahi.c
nunojsa: Don't think so...

dns_sd_avahi.c
nunojsa: ack

dns_sd_avahi.c
```diff
@@ -60,6 +60,39 @@ void dnssd_free_discovery_data(struct dns_sd_discovery_data *d)
 	free(d);
 }
 
+static void __avahi_process_resolved(struct dns_sd_discovery_data *ddata, const AvahiAddress *addr,
```
dNechita, line 63: I suggest avoiding the use of double underscore at the beginning of name.

dNechita: Such naming seems to be reserved by gnu compilers: https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html

dns_sd_windows.c
```diff
@@ -263,3 +263,8 @@ int dnssd_find_hosts(struct dns_sd_discovery_data** ddata)
 
 	return 0;
 }
+
+int dnssd_resolve_host(const char *hostname, char *ip_addr, const int addr_len)
+{
+	return -ENOENT;
+}
```
dNechita, line 270: new line is missing

dns_sd_avahi.c
nunojsa: Oh, didn't new this. Nice catch!

dns_sd_windows.c
nunojsa: ack

nunojsa:
Changes in v3:

1. Added/removed newlines where appropriate;
2. Reduce error msg size to 256;
3. Removed double underscore  at the beginning of functions;
4. Made `avahi_resolve_host` static.

dns_sd_avahi.c
rgetz: >So, this warning is still coming from codacy and I don't think there's much we can do about it 

I think it's that inside that function, it doesn't know the lengths. Since humans do, just do something like:

https://github.com/analogdevicesinc/libiio/blob/master/tests/iio_common.c#L66

nunojsa: Ahh nice! I could have figured that we could somehow ignore the warning 🤦...

nunojsa:
Changes in v4:
1. Add `/* Flawfinder: ignore */`.

rgetz:
CI is failing, you need to fix the issues on MAC.
```
In file included from /Users/travis/build/analogdevicesinc/libiio/utilities.c:24:
4865/Users/travis/build/analogdevicesinc/libiio/network.h:79:16: error: use of undeclared identifier 'DNS_SD_ADDRESS_STR_MAX'
4866        char addr_str[DNS_SD_ADDRESS_STR_MAX];
4867                      ^
48681 error generated.
4869make[2]: *** [CMakeFiles/iio.dir/utilities.c.o] Error 1
4870make[1]: *** [CMakeFiles/iio.dir/all] Error 2
4871make: *** [all] Error 2
```

nunojsa:
Changes in v5:
1. Restore `DNS_SD_ADDRESS_STR_MAX` define for MAC build.

nunojsa:
So, is there anything pending here? While I really feel that both patches are worth for inclusion, patch 1 is definitely needed... Just making sure that this is not forgotten

commodo:
> So, is there anything pending here? While I really feel that both patches are worth for inclusion, patch 1 is definitely needed... Just making sure that this is not forgotten

Well, I don't agree with the entire patch `network: Improve context creation robustness ` ; so, I'd let someone else ack that.
Or have it split up in smaller patches.
I'd merge patch `network: Fix leak in `dnssd_find_hosts()`  separately for now.

nunojsa:
Also bringing this into @pcercuei attention. This is sitting here for a long time already. I can re-spin this PR if the intention is to merge it otherwise, let's just pick the first patch...

network.c
```diff
@@ -1402,6 +1402,29 @@ struct iio_context * network_create_context(const char *host)
 #endif
 	{
 		ret = getaddrinfo(host, IIOD_PORT_STR, &hints, &res);
+		/*
+		 * It might be an avahi hostname which means that getaddrinfo() will only work if
+		 * nss-mdns is installed on the host and /etc/nsswitch.conf is correctly configured
+		 * which might be not the case for some minimalist distros. In this case,
+		 * as a last resort, let's try to resolve the host with avahi...
+		 */
+		if (ret) {
```
pcercuei, line 1411: With the latest master, you can write: `if (HAVE_DNS_SD && ret)`. Then you can remove the static inline  `dnssd_resolve_host` fallback in network.h.

dns_sd_avahi.c
pcercuei: Agreed.

network.c
nunojsa: Yeah I know :). If we are going to take this, I definitely have to re-apply all of this on top of the new master. But I'm only going to have that work if this is to be merged :smile:

dns_sd_avahi.c
nunojsa: Well, I still think this is simple enough that it could go in the same patch but it's already 2 complaining about this so I have to resign myself and do it in it's own patch :sweat_smile:

pcercuei: It's a good habit to create many small commits with atomic changes rather than a big one. Especially if you have people (maintainers) review your code. Here in particular, it's not obvious at first sight that this code didn't change and was just moved, and it makes the actual functionality change appear much bigger than it should be.

network.c
pcercuei: Well, I don't see why I wouldn't merge it, if it's done properly and has been tested. Go for it!

nunojsa:
Changes in v6: 

1.  Rebased on top of master;
2.  Avahi host process function moved to it's own commit.

To test this I created a simple test program:

```
#include <stdio.h>
#include <iio.h>
  
int main(void)
{
         struct iio_context *ctx;
 
         ctx = iio_create_network_context("analog.local");
         if (!ctx) {
                 printf("Unable to find iio context\n");
                 return -1;
         }
 
         printf("Found context with %d devices\n", iio_context_get_devices_count(ctx));
 
         return 0;
 }
``` 

Running it on my archlinux machine I have (without and with this patches):

```
[nsa@nsa libiio]$ ./net
ERROR: Unable to find host: Name or service not known
Unable to find iio context

[nsa@nsa libiio]$ ./net
Found context with 4 devices
```

network.c
```diff
@@ -1082,6 +1082,30 @@ struct iio_context * network_create_context(const char *host)
 		ret = getaddrinfo(addr_str, port_str, &hints, &res);
 	} else {
 		ret = getaddrinfo(host, IIOD_PORT_STR, &hints, &res);
```
pcercuei, line 1084: Just a quick question. Would it be better to resolve first with `dnssd_resolve_host`, instead of calling it on fallback?

nunojsa: Hmm, never thought about that as I just wanted to be the less disruptive possible. I would assume that most of the more popular linux distros already come with `nss-mdns` which means `getaddrinfo` is capable of resolving the name [and so we should really get into the fallback phase] and most likely faster than this implementation. OTOH, using `dnssd_resolve_host` first should ensure that things always work without the need of a fallback... I have no preference here...

pcercuei: Could you try if `dnssd_resolve_host` work with an IP and with `localhost`? If it does, then we can call it first, then fall back to getaddrinfo.

network.c
nunojsa: As it stands, it does not work...

```
[nsa@nsa libiio]$ ./net
ERROR: Unable to create Avahi DNS-SD relsover: Invalid host name
ERROR: Unable to create Avahi DNS-SD relsover: Invalid host name
```

However, I can tell that `avahi-resolve` has an address option that just uses another [implementation](https://github.com/rockchip-android/external-avahi/blob/e67ae65eb3c48969c16751feccfa9daf74b0f406/avahi-utils/avahi-resolve.c#L306). So, in theory, we could support it but we would need more complexity [basically just checking if we have an IP]. For the localhost, I'm getting the same error and I'm not sure if there's a way for it. So, IMO I would just stick with the fallback otherwise we might be breaking some corner case that we are not seeing...

nunojsa: Maybe we would not even need to do the IP check... but the localhost is still a question mark

nunojsa: Ups, just realized now that this is the wrong thread :facepalm:

pcercuei: Alright, green light for me then.

dns_sd_avahi.c
```diff
@@ -58,6 +58,38 @@ void dnssd_free_discovery_data(struct dns_sd_discovery_data *d)
...
+       /* Find empty data to store things*/
+       while (ddata->next)
+               ddata = ddata->next;
+
+       /* link a new placeholder to the list */
+       avahi_address_snprint(ddata->addr_str,
+                       sizeof(ddata->addr_str), addr);
+       memcpy(ddata->address, addr, sizeof(*ddata->address)); /* Flawfinder: ignore */
+       ddata->port = port;
+       ddata->hostname = strdup(host_name);
```
pcercuei, line 77: I believe this should be iio_strdup, but since you didn't modify that code, it can be done in a future commit.

pcercuei (approved):

pcercuei:
Merged. Thanks!

PR #617, opened 2020-11-02
CI/travis/lib.sh: change Travis-CI API endpoint to api.travis-ci.com

commodo:
We've recently been seeing issues with travis-ci.org; particularly builds
would hang/not-start.
This is bad, since we can no longer be comfortable validating that a change
builds with no errors. So, this would be a bit of a regression in process.

At this point in time, all ADI Github repos have been migrated to
travis-ci.com.
This migration would happen anyway until 31st of December 2020.
The plan we're using is still for Open-Source projects, so no extra cost
should be involved.

This change, switches the API endpoint to 'api.travis-ci.com', so that a
change in libiio propagates to the libad9361-iio, gr-iio, iio-oscilloscope
repos.
Without this, libiio would build, but would not be validated (build-wise)
against these repos.

Some refs:
https://travis-ci.community/t/builds-queuing-for-hours-with-public-git-repo/10351
https://docs.travis-ci.com/user/migrate/open-source-repository-migration#frequently-asked-questions


dNechita:
The badges/shields on main readme should be modified as well

commodo:
good catch
Changelog v1 -> v2:
* update badge URLs to travis-ci.com

dNechita (approved):

PR #619, opened 2020-11-06
.github/actions: add initial build.yml file

commodo:
Currently just does a build, an only on PRs to save time/resources.
Deployments will be re-added in the next commit series.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,81 @@
+name: Build the libiio matrix
+
+on:
+  push:
+    branches:
+      - 'staging/**'
+  pull_request:
+    branches:
+      - main
+      - master
+      - '20[1-9][0-9]_R[1-9]'
+
+jobs:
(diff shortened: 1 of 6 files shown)


commodo:
This is ready for review now.
Notes:
* we don't have any deployments yet with this, but we at least get sure that things still build;
* currently running builds on PRs
* branches get built **only** if they are prefixed with `staging/` ; this means branch building is voluntary/opt-in
* dropped CentOS 6 builds during migration; we can re-add it; but it's old enough
* the only MacOS X images on GH actions are version 10.5 and 11.0
* suffixed builds with CPU architecture `_x86_64` ; we may want to add `_arm` or `_arm64` later;
* will try to migrate Windows here as well

rgetz:
It's unlikely that we would use github actions (it cost too much for what we need). We are discussing, and will get back next week.

commodo:
Closing this; will spin-up an Azure Pipelines build.

Issue #593, opened 2020-10-01
Document iiod


tfcollins:
Need basic doc on iiod and its configuration. There is some handy functionality in iiod that users should be aware of like context attributes.

rgetz:
I think this is done in #959 now ?

PR #621, opened 2020-11-10
iiod/init: add a script which creates /etc/libiio.ini and bolt it in

rgetz:
This add a small script /usr/sbin/iiod_create_ini which can create a
/etc/libiio.ini file. This is connected to the startup scripts, and run
before iiod, so that things are populated on time.

Re-work a little of the CMake so that this is included during an
install.

Update the doc (the README_BUILD.md) so these options are clear for
people.

This is the first stage of #593


tfcollins (commented):
Have we been able to test all 3?

rgetz:
I have tested the ini creation on x86, Zynq and Pi.
I have tested the sysvinit (Zynq) and systemd (x86 & Pi).

Upstart didn't change much from the existing code.

mhennerich:
https://github.com/analogdevicesinc/linux_image_ADI-scripts/blob/master/iiod_usbd.init#L84

We also handle UDC (usb_gadget) and IIO USB functionfs support.
Do we want that to include here as well? It matters since IIOD needs to be started differently.

rgetz:
Can do - that makes sense.

I would have a uniq script to set up the gadget device (which is common, and will be less copy/paste across the different init systems). basically what is in here: https://github.com/analogdevicesinc/linux_image_ADI-scripts/blob/master/iiod_usbd.init#L84-L112

mirceacaprioru:
I was working on some systemd services for UDC. Systemd has the possibility to trigger services based on folder updates like the /sys/class/udc.

 I have not succeeded  in testing this services using the ZCU102. But if we add them to libiio that would be great from my side too since they are tied.

rgetz:
> But if we add them to libiio that would be great from my side too since they are tied.

What specifically do you need?

mirceacaprioru:
It is basically just splitting this script that you also pointed out before: https://github.com/analogdevicesinc/linux_image_ADI-scripts/blob/master/iiod_usbd.init#L84-L130

rgetz:
I have something that I think works - but I need to test a lot more on a few different systems before it's merged.

rgetz:
getting back to this - has anyone else tested this but me? I will update to latest master if no one has any comments.

mhennerich:
Please see here: 

https://github.com/analogdevicesinc/linux_image_ADI-scripts/pull/27

This does effectively the same and can live completely outside libiio

rgetz:
I think this was replaced with https://github.com/analogdevicesinc/linux_image_ADI-scripts/pull/27

PR #622, opened 2020-11-20
tests: iio_readdev: Set minimal buffer size to 2

gwendalcr:
Currenly set at 64, it is not optimal when a smaller number of samples
are requested. For instance, if 4 samples are requested, we still have
to wait for up to 64 samples to come in to return.

### DIFF
--- a/tests/iio_readdev.c
+++ b/tests/iio_readdev.c
@@ -236,7 +236,7 @@ int main(int argc, char **argv)
 				fprintf(stderr, "Buffersize requires an argument\n");
 				return EXIT_FAILURE;
 			}
-			buffer_size = sanitize_clamp("buffer size", optarg, 64, 4 * 1024 * 1024);
+			buffer_size = sanitize_clamp("buffer size", optarg, 1, SIZE_MAX);
 			break;
 		case 's':
 			if (!optarg) {
--- a/tests/iio_writedev.c
+++ b/tests/iio_writedev.c
@@ -247,7 +247,7 @@ int main(int argc, char **argv)
 				fprintf(stderr, "Buffer Size requires argument\n");
 				return EXIT_FAILURE;
 			}
-			buffer_size = sanitize_clamp("buffer size", optarg, 64, 4 * 1024 * 1024);
+			buffer_size = sanitize_clamp("buffer size", optarg, 1, SIZE_MAX);
 			break;
 		case 's':
 			if (!optarg) {


rgetz:
I think this was set to 64 since the limit was a cache line when doing things with local & DMA.

I would need to get @mhennerich 's feedback how this might impact some of the drivers...

mhennerich:
IMHO we shouldn't hardcode constrains here.
The actual behavior depends on the underlaying buffer implementation.
For example low speed buffers shouldn't have such limitations at all.

For high speed buffers even when the buffer size is below dmaengine_buffer->align, things will succeed.
However samples will be lost, since the IIO dmabuf implementation will round to the alignment.
And when the user request a buffer smaller, he won't be able to consume all samples actually captured.

For this reason Lars introduced: length_align_bytes
https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio-dma-buffer

Please see this pathwork for more information:

https://lore.kernel.org/patchwork/patch/1165432/

-Michael

rgetz:
@mhennerich : thanks for feedback.

So - what should be the limit to readdev/writedev? the number needs to be clamped to something (1 to INT_MAX)?

right now it's 64 to 4 Meg.

This patch makes it 2 to 4 Meg.

Suggestions?

And then - it looks like we should be checking the new attribute inside the libiio core...

mhennerich:
The proper range would be 1 to limit of `size_t' type. Typically a 32 or 64 unsigned.
I wouldn't do any further checking.

**buffer_size is in samples not bytes!**

The real buffer size in bytes is buffer_size * sample_size where sample_size is computed from the number of enabled channels and each of their storage sizes. [sample_size = iio_device_get_sample_size(dev);]

Let the kernel decide if buffer_size is valid and supported or not.

Prior to this 503fa9bf9cc863295ab1842ac320c9d0985e009f the buffer_size was limited to int. Which would have only caused issues with negative numbers.

rgetz:
Thanks.

`SIZE_MAX (C99) maximum value of object of size_t type`

it is then.

@gwendalcr did you want to update your patch to that? (or did you want me to do it?) (and at the same time, make the same change to 'iio_writedev')

gwendalcr:
Patch updated, allowed range is now 1..SIZE_MAX.

mhennerich (approved):

dNechita (approved):

rgetz:
thanks for update.

PR #629, opened 2020-12-11
ci,azure-pipelines: add coverity job

commodo:
This change adds job for the Coverity scan project. There doesn't seem to
be an environment variable (at first glance) for the repo slug.
So, the name of the repo was added directly in the yaml file.

This doesn't run yet. Some more setup is required. It does a port of the
environment variables from the .travis.yml file.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,6 +13,23 @@ pr:
 - 20*
 
 jobs:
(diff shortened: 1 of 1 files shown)


rgetz:
We have a different branch that seems to be working. Need to see if from works - but otherwise I will close this.

PR #634, opened 2020-12-16
ci,azure-pipelines: add windows builds

commodo:
This change adds a minimal configuration to build some Windows artifacts
for libiio.
These are stored after the pipeline is done.

Adapted from Travis Collins' work.
Added into PowerShell scripts. Ideally we should not add any script-code
into yaml files, as that makes it harder to scale, and even harder to
migrate (if ever will be needed again).

Publishing artifacts, the generated DLL, static library, and the iio.h
header, which will be used in the next jobs.

### DIFF
--- a/CI/build_win.ps1
+++ b/CI/build_win.ps1
@@ -1,0 +1,11 @@
+
+$COMPILER=$Env:COMPILER
+$ARCH=$Env:ARCH
+
+$src_dir=$pwd
+
+mkdir build
+cd build
+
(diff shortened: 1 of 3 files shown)


commodo:
closing & re-opening to see whether appveyor builds;
i changed it to build only `no_branch` branches

since we're migrating to Azure Pipelines, might as well avoid building on Appveyor as it fails almost always;

mhennerich (approved):

PR #635, opened 2020-12-16
[RFC] ci,linux: reduce python pip deps and use distro's python3

commodo:
All Linux distros [except for Ubuntu 16.04] support Python 3.6 or newer.
For this version it might make sense to no longer build docs, since it's getting old (5 years in a few months).

With this change we switch to using the distribution's Python3 package.
This unclutters some Python setup.

Since we're moving to Azure, it is a good opportunity to do this cleanup.

### DIFF
--- a/CI/travis/before_install_linux
+++ b/CI/travis/before_install_linux
@@ -13,26 +13,6 @@ install_sphinx() {
 	fi
 }
 
-install_pyenv() {
-	if ! command_exists pyenv ; then
-		echo installing pyenv
-		git clone git://github.com/yyuu/pyenv.git $HOME/.pyenv
-		export PATH="$HOME/.pyenv/bin:$PATH"
-		export PYENV_ROOT="$HOME/.pyenv"
-	fi
-}
-
-install_python() {
-	echo "### installing python"
-	command -v pyenv
-	add_python_path
-	pyenv install --list | grep "^[[:space:]]*[0-9]"
-	pyenv install 3.6.3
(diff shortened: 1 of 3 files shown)


tfcollins:
Should we just drop 16.04?

-Travis

commodo:
Fine with me to drop.
Just need to check (first) if Matlab still needs 16.04
If they don’t need it, I see no reason not to drop.

Though at this point as the entire CI is resetting, it is a bit more
difficult to support some of the old matrix, while also moving forward.

tfcollins:
Support ends in April. Should probably remove it then. 

-Travis

tfcollins:
Looks fine. There is a weird print here: https://dev.azure.com/AnalogDevices/OpenSource/_build/results?buildId=287&view=logs&j=a459469c-8d72-520b-630e-f50592a3038d&t=6328fd15-4213-5932-7fbc-42817ed20acd&l=100

I don't think that's related though.

tfcollins (approved):

commodo:
> Looks fine. There is a weird print here: https://dev.azure.com/AnalogDevices/OpenSource/_build/results?buildId=287&view=logs&j=a459469c-8d72-520b-630e-f50592a3038d&t=6328fd15-4213-5932-7fbc-42817ed20acd&l=100
> 
> I don't think that's related though.

yeah, that's from CMake:
https://github.com/analogdevicesinc/libiio/blob/master/bindings/python/CMakeLists.txt#L25

PR #639, opened 2020-12-18
[RFC] ci,travis: remove some old build elements

commodo:
The deploy stuff is being removed. This isn't being used anymore in Azure.
The deploy key is especially important to remove, since this can be
downloaded from Github via zip.

The QEMU elements have been replaced with multiarch docker images. Docker
images for ARM/ARM64 (or other architectures) can be run.


rgetz:
There was some specific mangling/mac things in files you removed - are those somewhere else now? They will still be needed for mac builds - or we will just end up with 5 year or user complaints...

There was alot of post-build, pre-deploy pieces - where did those go?

commodo:
> There was some specific mangling/mac things in files you removed - are those somewhere else now? They will still be needed for mac builds - or we will just end up with 5 year or user complaints...
> 
> There was alot of post-build, pre-deploy pieces - where did those go?

Oh. So that Mac mangling should definitely go into the `make_darwin` ; thanks for the heads-up.
It looks like it should have been there from the start.

The other pieces are redundant though.

commodo:
Closing this
swdownloads will still be in yse

PR #642, opened 2020-12-22
[RFC] ci,linux: don't install sphinx docs on ARM{64}

commodo:
Since the environment for building on ARM/ARM64 is emulated, execution is a
bit slow.
The whole build with docs takes 13 minutes on ARM and 17 minutes on ARM64.
So, for now we don't build docs on these platforms.


commodo:
So, the ARM builds take about 9-11 minutes each; down from 13-17.
I'd say it's a good time-reduction.

rgetz:
yeah, I think we only built the doc on one platform, and didn't install it (or the man pages) in any of the deb/rpms.

rgetz (approved):

commodo:
> yeah, I think we only built the doc on one platform, and didn't install it (or the man pages) in any of the deb/rpms.

we could switch this to a build flag (env-var) opt-in like `BUILD_SPHINX_DOCS=1` and then add it to the build matrix where we want it;
that offers a bit more granular/explicit control, which sounds like we have a better grasp/control of our builds;

the `build docs if not-arm` is a bit misleading;
for once, we don't build docs on CentOS [we could and it works, I tried it];
if we do `BUILD_SPHINX_DOCS=1`, then we can also add an error check to fail the build if the docs aren't there;

commodo:
Changelog v1 -> v2:
* implemented opt-in flag `BUILD_SPHINX_DOCS=1 ` ; this way if sphinx docs can't be built, the build will error out

rgetz:
We just need to differentiate the CI build vars, and the CMake build vars.

I got confused when reading the description - we have Cmake `WITH_DOC`, and `WITH_MAN` - but the code is clear...

Should we namespace things with `CI_` or something? (CMAKE already uses '_BUILD_` for some of it's internal things).

commodo:
> We just need to differentiate the CI build vars, and the CMake build vars.
> 
> I got confused when reading the description - we have Cmake `WITH_DOC`, and `WITH_MAN` - but the code is clear...
> 
> Should we namespace things with `CI_` or something? (CMAKE already uses '_BUILD_` for some of it's internal things).

sure ; will update to   `CI_BUILD_SPHINX_DOCS=1`   if that's ok

commodo:
Changelog v2 -> v3:
* renamed `BUILD_SPHINX_DOCS` -> `CI_BUILD_SPHINX_DOCS`

commodo:
Will merge this next week if no objections.

rgetz:
looks great - thanks

PR #645, opened 2021-01-20
serial: Detect timeout when reading a byte from serial port

dNechita:
The descrpition of sp_blocking_read_next() specifies that 0 is returned on timeout. This commit handles this.
An issue has been observed when trying to create a context for the wrong serial port (where no tiny-iiod exists on the other end) which causes a really long wait time because sp_blocking_read_next() was being called for at least 1024 times and each time it timedout after 1 second.

This fixes https://github.com/analogdevicesinc/iio-oscilloscope/issues/280

### DIFF
--- a/serial.c
+++ b/serial.c
@@ -281,6 +281,12 @@ static ssize_t serial_read_data(struct iio_context_pdata *pdata,
 				pdata->port, buf, len, pdata->timeout_ms));
 
 	IIO_DEBUG("Read returned %li: %.*s\n", (long) ret, (int) ret, buf);
+
+	if (ret == 0) {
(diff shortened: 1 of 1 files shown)


commodo (approved):

commodo:
This looks good, but it also points out that we should maybe do a review of the libserial timeouts.
I'm thinking other operations would also timeout.

nunojsa (approved):

dNechita:
> This looks good, but it also points out that we should maybe do a review of the libserial timeouts.
> I'm thinking other operations would also timeout.

That is a good idea. I've created an issue to keep track of your suggestion. Thanks!

PR #647, opened 2021-01-26
serial: Detect timeout when writing to serial port

dNechita:
This handles #645

### DIFF
--- a/serial.c
+++ b/serial.c
@@ -271,6 +271,12 @@ static ssize_t serial_write_data(struct iio_context_pdata *pdata,
 				pdata->port, data, len, pdata->timeout_ms));
 
 	IIO_DEBUG("Write returned %li: %s\n", (long) ret, data);
+
+	if (ret == 0) {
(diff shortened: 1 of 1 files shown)


commodo (approved):

commodo:
maybe a bit of a comment in the commit would be a good idea;
rest looks good;

rgetz:
There is a problem in this PR.

https://www.sigrok.org/api/libserialport/unstable/a00026.html#gaf765f17269cd1ea2e22c0b9e61d523ae
```
The number of bytes written on success, or a negative error code.
```

so, if you send an empty buffer, you will get zero back; and it's not a timeout.

```
If the number of bytes returned is less than that requested, the timeout was reached before the 
requested number of bytes was written.
```

so, its not a `ret == 0` test; it should be  `ret < len` test.

dNechita:
Yes, you are right. I will fix this by the EOD.
Thanks for pointing this out!

PR #649, opened 2021-02-10
iio: rework entire xml generation

commodo:
The entire XML generation (the one done manually) is pretty difficult to
change, so if someone wants to add some attributes (because some kernel ABI
got updated), it's easy to change and introduce leaks.

This code as-is already has some potential leaks (and faults).
One example is in in iio_device_get_xml() (on the error path), where:
```
err_free_debug_attrs:
	while (k--)
		free(debug_attrs[k]);
	free(debug_attrs);
err_free_debug_attrs_len:
	free(debug_attrs_len);
err_free_buffer_attrs:
	while (k--)
		free(buffer_attrs[k]);
	free(buffer_attrs);
```

Now, it's obvious that such errors can be fixed, but a better idea is to do
a single allocation and use the 'snprintf(NULL, 0, ...)' to count the
amount of memory we need for the XML string.

This change does that. And all the allocations (that were needed before)
are reduced to a single one (vs 10+ ... I stopped counting).

This obviously simplifies error paths as the risk of leaks is dramatically
reduced. A lot of code is also reduced, because we don't need to track any
pointers ourselves.
Admittedly, snprintf() may do some allocations under the hood, it's still
better to offload the pointer tracking to snprintf(), than keep doing it
ourselves.

Unfortunately, this change has to be done a single go. A bit more effort
could be invested into splitting this into smaller commits. But the way the
XML string generation works, once you start pulling on this thread, it
tends to unwind in multiple places.

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -177,168 +177,100 @@ void iio_channel_init_finalize(struct iio_channel *chn)
 	}
 }
 
-static char *get_attr_xml(struct iio_channel_attr *attr, size_t *length)
+static ssize_t iio_snprintf_chan_attr_xml(char *str, ssize_t len,
+					  struct iio_channel_attr *attr)
 {
-	char *str;
-	size_t len;
+	if (!attr->filename)
+		return iio_snprintf(str, len, "<attribute name=\"%s\" />", attr->name);
 
-	len = strnlen(attr->name, MAX_ATTR_NAME);
-	len += sizeof("<attribute name=\"\" />") - 1;
-
-	if (attr->filename) {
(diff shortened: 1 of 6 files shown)


mhennerich (approved):

mhennerich:
Looks good to me, but that's have some more review on this.

nunojsa (approved):

dNechita (approved):

rgetz:
Michael's suggestion was "Maybe merge right after we have done the next release."

What happened to that?

PR #651, opened 2021-02-16
local: introduce support for new iio buffer interfaces

commodo:
This change introduces support for the new IIO multibuffer support (i.e
multiple IIO buffers per one IIO device). The way this works is that
instead of using the FD for /dev/iio:deviceX to read data from it, we
request a new FD from which we read data.

This means that we can close the old FD (for /dev/iio:deviceX) and use the
new FD (from the ioctl()).

In the kernel, the new ioctls for the high-speed mmap interface should also
be at around the same time with this new multibuffer support. The ioctls
for the mmap interface should be done on the FD returned by the GETFD
ioctl.

For now, this interface is still used to access just a single IIO buffer
per device. This patch just does the transition to the new interface.

### DIFF
--- a/local.c
+++ b/local.c
@@ -48,24 +48,6 @@
 
 #define NB_BLOCKS 4
 
-#define BLOCK_ALLOC_IOCTL   _IOWR('i', 0xa0, struct block_alloc_req)
-#define BLOCK_FREE_IOCTL      _IO('i', 0xa1)
-#define BLOCK_QUERY_IOCTL   _IOWR('i', 0xa2, struct block)
-#define BLOCK_ENQUEUE_IOCTL _IOWR('i', 0xa3, struct block)
-#define BLOCK_DEQUEUE_IOCTL _IOWR('i', 0xa4, struct block)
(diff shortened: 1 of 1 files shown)


commodo:
This PR still requires that the high-speed buffer interface be upstreamed.
The multibuffer support was accepted already.

This was tested on an AD9467 with kernel 5.11

commodo:
the new ioctls have been accepted:

https://lore.kernel.org/linux-iio/20210215143234.3248-5-alexandru.ardelean@analog.com/T/#m3b9c96739a5ffb0c99f0778e2d8cbba555a0f12a

however, i'm sending a v4 to fix some issues found in some build configurations;

commodo:
The new mmap support is merged.
This change should be good to go in.

I'd recommend some re-testing this. I've only tested this briefly on kernel 5.11.

pcercuei:
Closing this PR, as the mmap interface was not merged upstream, and libiio will move towards using a DMABUF based API.

PR #652, opened 2021-02-22
Fix timeout handling serial write

dNechita:


rgetz:
Don't we need to handle the error? Just printing something, and allowing a partial write to occur will confuse the system.,,

dNechita:
> Don't we need to handle the error? Just printing something, and allowing a partial write to occur will confuse the system.,,

When the timeout occurs, yes we print something but then we immediately return and provide the error code for the timeout.
Did you have something in mind about what needs to be done to handle the timeout?

rgetz:
just returning in the middle - will leave the client/server in an unknown state (although that looks like what usb does); so maybe it is OK. 

when I test this, I get different results based on the random devices:
```
rgetz@brain:~/github/libiio-master/build$ time ./tests/iio_info -u serial:/dev/ttyUSB0
Library version: 0.21 (git tag: 9d79757)
ERROR: sp_blocking_read_next has timedoutERROR: READ LINE: -110
Unable to create IIO context serial:/dev/ttyUSB0: Connection timed out (110)

real	0m31.020s
user	0m0.114s
sys	0m0.114s
```

31 seconds it pretty long to time out?

```
rgetz@brain:~/github/libiio-master/build$ time ./tests/iio_info -u serial:/dev/ttyACM0
Library version: 0.21 (git tag: 9d79757)
ERROR: sp_blocking_read_next has timedoutERROR: READ LINE: -110
Unable to create IIO context serial:/dev/ttyACM0: Connection timed out (110)

real	0m1.038s
user	0m0.009s
sys	0m0.000s
```
seems more reasonable.

rgetz:
I dug into this a little bit... This seems to be an issue in the FTDI kernel driver.

https://stackoverflow.com/questions/14255314/close-function-on-serial-port-descriptor-blocks-in-linux

there is no way to set the `closing_wait` function that I can find on this driver, it's set to 30 seconds... 

only suggestion was to have the close will probably need to be called in a separate thread... (which I'm not sure is worth the problem?) If we want to make serial scan work - it will be necessary....

rgetz (approved):

dNechita:
Thanks for investigating things. I'm going to merge this. Also I've created an issue to track the 30 seconds timeout issue.

PR #660, opened 2021-04-01
tests: Print context creation messages to stderr

### LINKED ISSUE #583: iio_readdev corrupts the data when using auto detection
Run iio_readdev in the auto detection mode (i.e. "-a"):

```
#: ./iio_readdev -a -s 2 cf-ad9361-lpc
Using auto-detected IIO context at URI "usb:20.10.5"
??????%
```

All data produced to stdout. Thus the message "Using auto-detected IIO context at URI "usb:20.10.5"" corrupts the output.

pcercuei:
When using iio_readdev, stdout is reserved for the data output; so
information messages about the context creation must be sent to stderr
so that the data stream is not polluted.

Fixes #583.

### DIFF
--- a/tests/iio_common.c
+++ b/tests/iio_common.c
@@ -95,11 +95,11 @@ struct iio_context * autodetect_context(bool rtn, const char * name, const char 
 	}
 
 	if (ret == 0) {
-		printf("No IIO context found.\n");
+		fprintf(stderr, "No IIO context found.\n");
 		goto err_free_info_list;
 	}
 	if (rtn && ret == 1) {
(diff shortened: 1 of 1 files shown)


mhennerich (approved):

rgetz:
Did we just want to print this while "-v" (verbose)? So in the normal case, people don't need to redirect stderr?

pcercuei:
I don't think people need to redirect stderr, since they probably redirect stdout already - stderr would just end up in the terminal.
There is no "-v" option yet anyway, is there?

pcercuei:
I'll merge this now as a quick & easy fix, we can still rework it later with a "verbose" option.

PR #661, opened 2021-04-01
Support for device labels

pcercuei:
**Completely untested.**

The label of an IIO device is an optional attribute that can be specified by the OS (generally through Device Tree), to differenciate IIO devices that would otherwise have exactly the same name.

This PR adds support for reading a IIO device's label using a new API function `iio_device_get_label()`.

The `iio_context_find_device` has been augmented to now also search by label.

The device labels are now printed as part of the XML generated by `iio_context_create_xml`. Using this XML to create a context (through the network, USB or UART) using an older version of libiio should still work, but will print warnings about the unrecognized "label" attribute. Using an XML string printed before this change with a new version of libiio will simply cause `iio_device_get_label()` to return NULL, without any side effect.


local.c
```diff
@@ -1221,6 +1221,22 @@ static int read_device_name(struct iio_device *dev)
 		return 0;
 }
 
+static int read_device_label(struct iio_device *dev)
+{
+	char buf[1024];
```
commodo, line 1226: 1024 bytes on the stack is a bit much
maybe 128/256?

pcercuei: It's been 1024 bytes everywhere in local.c, I'm just being consistent :)

commodo: Yeah I know.
It’s also fine to leave it like this for now.
I wanted to maybe reduce it everywhere.

I just never got to it.

I’m usually paranoid with some libc stacks like musl that used to have small pthread stack sizes (~128k)
If you combine several libs that use 1k buffers like this, it increases the chances of stack overflow.

context.c
```diff
@@ -237,7 +237,8 @@ struct iio_device * iio_context_find_device(const struct iio_context *ctx,
 	for (i = 0; i < ctx->nb_devices; i++) {
 		struct iio_device *dev = ctx->devices[i];
 		if (!strcmp(dev->id, name) ||
-				(dev->name && !strcmp(dev->name, name)))
+				(dev->name && !strcmp(dev->name, name)) ||
+				(dev->label && !strcmp(dev->label, name)))
```
nunojsa, line 241: I'm not sure about the device id. But between the label and the name, I would first match against the label. If a label is present, most likely it is how the device wants to be found 😄

nunojsa (commented):
I'll be back now only Tuesday (Easter holidays). If by then, you still don't have any HW to test, I can do some testing (unless someone else does it)

context.c
mirceacaprioru: I would disagree here. The name is still the main way to identify a device. A label comes into play when you have multiple devices with the same name and you would like to know which is which. That is just the way I used labels.

nunojsa: Yes, exactly. And for me that means that if someone defines a label in DTS for a device, he wants it to be primarily identified by that label. Image the following scenario where someone has two devices with the part name foo. It would be a valid option to just use a label in the second device like foo_1, leaving the first device just with the name. With this, I guess `iio_context_find_device()` could always return the same device instead of foo and foo_1...

pcercuei: @mirceacaprioru if you have multiple devices with the same name, then the label is the differenciating factor. Therefore it has a higher priority than the name, and @nunojsa is right.

mirceacaprioru: ack. Thanks

local.c
rgetz: there isn't anything in the kernel which actually limits this - it (in theory) should be bigger (should be page size); although from a practical perspective - it's fine with 1k.

rgetz:
https://app.codacy.com/gh/analogdevicesinc/libiio/pullRequest?prid=7196263 should be fixed before merge.

local.c
nunojsa: I think @commodo means threads created with pthreads where the stack size is not really controlled by the kernel but from libc. And some implementations as musl are more [conservatives](https://wiki.musl-libc.org/functional-differences-from-glibc.html#:~:text=Thread%20stack%20size,-The%20default%20stack&text=musl%20provides%20a%20default%20stack,21). Anyways, I'm not so paranoid here and think that 128k is more than enough and if some app uses that much in the stack, then arguably the __real__ problem is not really in this 1k...

rgetz: right - what I'm saying is that the actual label - can be up to a PAGESIZE in length - this is controlled by the kernel - libiio can truncate it - but it shouldn't; since it becomes less unique - which is the whole point.

nunojsa: Ahh I see. Yeah, that might happen even though It's probably unlikely for someone to define a label bigger than 1k [but who knows]...

We could also just dynamically allocate a buffer with PAGE_SIZE len to be completely "correct". Anyways, I'm also ok with the 1k stack buffer...

nunojsa:
Anyone got the chance to test this?

local.c
pcercuei: If someone picks a IIO device label bigger than 1024 characters, I think he deserves some bugs here and there... Even UUIDs are only 36 bytes. As for the page size, hardcoding it to 4 KiB is wrong; it is generally that size, but that's not a safe thing to assume. It can also be much bigger...

nunojsa: > If someone picks a IIO device label bigger than 1024 characters, I think he deserves some bugs here and there...

Definitely...

> As for the page size, hardcoding it to 4 KiB is wrong...

Also true. We would need to do something like `sysconf(_SC_PAGESIZE)`to get it at runtime and then allocate a buffer for the label. But this surely feels like a lot of handling already!

FWIW, 1k is more than ok in my opinion...

pcercuei:
I just tested it on hardware. It works as expected.

With this in my device tree: `&adc { label = "my-custom-iio-device"; };`

I get:
```
$ ./iio_info 
...
IIO context has 2 devices:
	iio:device0: jz-adc (label: my-custom-iio-device)
...
```

nunojsa (approved):
LGTM... Should we keep the CI happy? If you don't really want to change the code, I think there are ways to make the CI to ignore it (adding some comments)...

pcercuei:
I think it's more important to keep the API consistent than to please the static analyzer.

nunojsa:
> I think it's more important to keep the API consistent than to please the static analyzer.

Agreed. My point was just that, most likely, you can add some comment on the code so that the analyzer does not report any error... Anyways, really up to you (for me the PR is good as is)

PR #662, opened 2021-04-06
Use SPDX license headers

pcercuei:
Use simple SPDX license strings instead of having paragraphs of text about the license in the header of every single source file of the project.

Also I'm pretty sure @cristi-iacob's first name is not Paul.

### DIFF
--- a/bindings/csharp/IioLib.cs
+++ b/bindings/csharp/IioLib.cs
@@ -1,20 +1,10 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
 /*
  * libiio - Library for interfacing industrial I/O (IIO) devices
  *
  * Copyright (C) 2020 Analog Devices, Inc.
  * Author: Paul Cristian Iacob <cristian.iacob@analog.com>
- *
- * This library is free software; you can redistribute it and/or
- * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation; either
- * version 2.1 of the License, or (at your option) any later version.
- *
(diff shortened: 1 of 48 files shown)


bindings/csharp/IioLib.cs
```diff
@@ -2,7 +2,7 @@
  * libiio - Library for interfacing industrial I/O (IIO) devices
  *
  * Copyright (C) 2020 Analog Devices, Inc.
- * Author: Paul Cristian Iacob <cristian.iacob@analog.com>
+ * Author: Cristian Iacob <cristian.iacob@analog.com>
```
mirceacaprioru, line 6: actually his name really is Paul Cristian Iacob.

pcercuei: Ha :D
All other headers just had "Cristian Iacob", so I thought it was a bad copy-paste.

mirceacaprioru:
just one nitpick some places have 
```
/* SPDX-License-Identifier: LGPL-2.1-or-later */
```
while others
`// SPDX-License-Identifier: LGPL-2.1-or-later`

But this is not a strong request. Just an observation

mirceacaprioru (approved):

pcercuei:
> just one nitpick some places have
> 
> ```
> /* SPDX-License-Identifier: LGPL-2.1-or-later */
> ```
> 
> while others
> `// SPDX-License-Identifier: LGPL-2.1-or-later`
> 
> But this is not a strong request. Just an observation

Yes, it does look strange, but this is the convention - /* */ in headers and // in C source files.

PR #663, opened 2021-04-07
Sanitize context attributes for XML

pcercuei:
This set of commits work on sanitizing the context attributes, so that they can be used in XML.

I did move `iio_snprintf` to being its own function. See the corresponding commit for details. `vsnprintf` is listed as supported in Microsoft's docs, so I expect it to build under MSVC with no problems.


pcercuei:
Rebased on top of master.

rgetz (approved):

PR #664, opened 2021-04-07
ERR_PTR

pcercuei:
If you are familiar with the Linux kernel, you will know how useful these functions are.

The first commit adds `IS_ERR()`, `ERR_TO_PTR()` and `PTR_TO_ERR()` macros, which can be used to encode or retrieve error codes into pointers.

The second commit shows how it can be used to make sure the error codes are properly cascaded back to the calling process, by updating the `iio_context_create_xml()` function.

### DIFF
--- a/context.c
+++ b/context.c
@@ -100,29 +100,24 @@ static ssize_t iio_snprintf_context_xml(char *ptr, ssize_t len,
 }
 
 /* Returns a string containing the XML representation of this context */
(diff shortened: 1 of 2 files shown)


nunojsa (approved):
LGTM. `ERR_CAST()` might also be an useful helper...

pcercuei:
@nunojsa maybe; I'd prefer to add it when we actually use it.

PR #666, opened 2021-04-08
treewide: Avoid conditional compilation + DNS-SD fixes

pcercuei:
Conditional compilation is a problem, and should be avoided as much as
possible.

Consider the following:

```
    #ifndef _RANDOM_FEATURE
    static int bar() {
    	bug bug bug
    }
    #endif

    int foo() {
    #ifndef _RANDOM_FEATURE
    	bar();
    #endif
    }
```

This code will compile just fine as long as _RANDOM_FEATURE is not
defined. This is a big problem, as the compiler will detect this bug
only when _RANDOM_FEATURE is enabled, which may only happen in some
corner cases. This is problematic especially with autobuilders, which
won't build every possible combination of configuration settings.

The solution is to define _RANDOM_FEATURE unconditionally, to 0 or 1
according to whether or not the setting is enabled, and use it outside
the #ifdef blocks, like this:

```
    static int bar() {
        bug bug bug
    }

    int foo() {
        if (_RANDOM_FEATURE)
	    bar();
    }
```

Here, the bug in bar() will be detected independently of the value of
_RANDOM_FEATURE. Should this option be disabled, the bar() function will
be properly detected as dead code, and will automatically be deleted
from the object by the compiler.

This is obviously only possible when the code protected by the
_RANDOM_FEATURE option is compilable even with the option disabled.


nunojsa (approved):
Even though I don't really have big problems with conditional compilation (as long as it's mainly done in headers; not really nice to see source code full of #ifdefs), the whole series LGTM...

PR #671, opened 2021-04-15
Xml backend cleanups

pcercuei:
A set of cleanups to the XML backend, to make sure that the error codes are carried all the way from the deeply nested functions back to the caller function.


rgetz (approved):
looks good. thanks

PR #673, opened 2021-04-21
DNS-SD cleanups

pcercuei:
A set of cleanups to the DNS-SD code, with no functional changes.


rgetz:
In general - I think most is fine - but we try to keep the windows version in sync with upstream - https://github.com/mjansson/mdns/blob/master/mdns.c (which there were some changes last week we should look at).

Did you want to work out the "cleanup" there, and then merge the final here?

Thanks

dns_sd_windows.c
```diff
@@ -19,42 +19,60 @@
...
-	// When sending, each socket can only send to one network interface
-	// Thus we need to open one socket for each interface and address family
-	int num_sockets = 0;
+static inline bool is_localhost4(struct sockaddr_in *saddr)
+{
+	return saddr->sin_addr.S_un.S_un_b.s_b1 == 127 &&
+		saddr->sin_addr.S_un.S_un_b.s_b2 == 0 &&
+		saddr->sin_addr.S_un.S_un_b.s_b3 == 0 &&
+		saddr->sin_addr.S_un.S_un_b.s_b4 == 1;
+}
```
nunojsa, line 47: maybe make *saddr `const`?

dns_sd_windows.c
```diff
@@ -19,42 +19,60 @@
...
+static const unsigned char localhost[] = {
+	0, 0, 0, 0, 0, 0, 0, 0,
+	0, 0, 0, 0, 0, 0, 0, 1
+};
+static const unsigned char localhost_mapped[] = {
+	0, 0, 0,    0,    0,    0, 0, 0,
+	0, 0, 0xff, 0xff, 0x7f, 0, 0, 1
+};
+
+static inline bool is_localhost6(struct sockaddr_in6 *saddr6)
```
nunojsa, line 32: ditto for the ``const?

More like a nitpick, in general I'm not a big fan of the `inline` (even being it a static function that only appears to be used once) as the compiler is smart enough to do it already without the hint. It can even be more "smarter" and ignore the keyword at all :smile: (which I don't think it would do on this case)...

Anyways, just my 2 cents

nunojsa (commented):
in general lgtm

dns_sd_windows.c
pcercuei: I feel like Codacity should report stuff like this :(

dns_sd_windows.c
pcercuei: The thing with `static inline` is that you can have it in a header, the compiler won't complain about it being defined everywhere, and won't complain if it's not used in a particular source file. I'm just too used to write `static inline` when I write functions that could be macros, there's not more to it.

pcercuei:
> In general - I think most is fine - but we try to keep the windows version in sync with upstream - https://github.com/mjansson/mdns/blob/master/mdns.c (which there were some changes last week we should look at).
> 
> Did you want to work out the "cleanup" there, and then merge the final here?

I don't know, our version looks so different now, we have a lot of local changes already that makes it near-impossible to stay in sync with the upstream version. I think it would be easier to manually cherry-pick the fixes from upstream.

Ideally, you're right though - but if we really want to go this route, then I'll need to contact the author, see what he thinks about the code cleanup (not point in doing this if he isn't on board), and then we can use the mdns code as a submodule.

dns_sd_windows.c
nunojsa: > The thing with static inline is that you can have it in a header

In headers sure... no way around it

dns_sd_windows.c
nunojsa: Agreed. It spills so much stuff that you do not really care and this is something easy for us to forget but very important to have...

rgetz:
> and then we can use the mdns code as a submodule.

yeah, I didn't see we were that far apart - but most is removing functions we don't need. I will open a issue to start https://github.com/mjansson/mdns/issues/43 and see what the opinion is.

I also don't think a submodule would work, since the upstream uses printf everywhere, which we moved over to IIO_DEBUG; and remembering to use ` -recursive` on git clone all the time sucks, and just generates support questions...

pcercuei:
Submodules aren't a problem, have a look at `git subtree` (it's an extension)

mhennerich (approved):

pcercuei:
@rgetz No feedback, I think we should merge this...

dns_sd.c
```diff
@@ -32,6 +32,13 @@
 #define close(s) closesocket(s)
 #endif
 
+static void dnssd_free_discovery_data(struct dns_sd_discovery_data *d)
```
rgetz, line 35: Did you test this on a Mac? 

While this is the same as the existing Linux code, it's different than the bonjour code. (which is why it was separate before).

pcercuei: `d->address` is NULL everywhere except on Linux, and `free(NULL)` is defined by the C standard to be a NOP, so this shouldn't break anything.

I did not test it on Mac, maybe @AlexandraTrifan can verify that it still works fine?

pcercuei: Alexandra tested it, and it still works fine.

PR #674, opened 2021-04-23
Fix warnings under MinGW and Clang

pcercuei:
Use the 'printf' format on all GNU compilers, which includes the
MinGW-flavoured GCC and Clang (which is not GNU, but says it is).

### DIFF
--- a/iio-private.h
+++ b/iio-private.h
@@ -37,10 +37,8 @@
 #define iio_sscanf sscanf
 #endif
 
-#if defined(__MINGW32__)
-#   define __iio_printf __attribute__((__format__(ms_printf, 3, 4)))
-#elif defined(__GNUC__)
-#   define __iio_printf __attribute__((__format__(gnu_printf, 3, 4)))
+#if defined(__GNUC__)
+#   define __iio_printf __attribute__((__format__(printf, 3, 4)))
 #else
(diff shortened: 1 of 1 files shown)


AlexandraTrifan (approved):
Tested on a mac, looks good.
The previous version raised some warnings at compile time.

pcercuei:
Thanks @AlexandraTrifan. @dNechita had the same problem but on MinGW, so I'll wait for his ACK as well.

dNechita (approved):

dNechita:
Hi! I'm still having those warnings. I haven't checked this properly in the first place, sorry about that.

dNechita:
the compiler I'm using is: x86_64-w64-mingw32-gcc-7.3-win32 (GCC) 7.3-win32 20180506

pcercuei:
@dNechita Can you paste the warnings?
From what I can see the "format" GCC function attribute exists since 4.x, so you should definitely have it. I tried to replicate it, but my mingw/msys2 installation uses GCC 10 already, and everything builds just fine.

dNechita:
[libiio-compile-warnings.txt](https://github.com/analogdevicesinc/libiio/files/6376841/libiio-compile-warnings.txt)

PR #677, opened 2021-04-26
Use gnu_printf format with GCC

pcercuei:
We don't actually *need* any of the GNU printf format specifiers.

However, older versions of MinGW will complain about C99-compliant
specifiers like %lld unless 'gnu_printf' is used.

Since this is used only with GCC, which is the "GNU C Compiler", it is
pretty safe to enable the 'gnu_printf' format unconditionally.

### DIFF
--- a/iio-private.h
+++ b/iio-private.h
@@ -37,7 +37,9 @@
 #define iio_sscanf sscanf
 #endif
 
-#if defined(__GNUC__)
+#if defined(__MINGW32__)
+#   define __iio_printf __attribute__((__format__(gnu_printf, 3, 4)))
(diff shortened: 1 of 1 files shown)


dNechita (approved):
Looks good. This time I've checked the build :)

dNechita:
Will it work for Clang?

pcercuei:
I guess we'll see ;)

pcercuei:
I tested locally, and Clang does not like it. Meh...

pcercuei:
PR updated. Now the gnu_printf format is only used on MinGW, and everybody should be happy.

Issue #373, opened 2020-02-11
incorrect channel type (in, out, in_out?)


rgetz:
I'm not sure if this is kernel issue or a libiio one.

```
iio_attr -a -c ad9361-phy out
Using auto-detected IIO context at URI "usb:3.20.5"
dev 'ad9361-phy', channel 'out' (input), attr 'voltage_filter_fir_en', value '0'
```

an input channel called `out`?

we have:
```
in_voltage_filter_fir_en
out_voltage_filter_fir_en
```

which follow the normal constructs for global voltage[n] attributes, but then we also have:

```
in_out_voltage_filter_fir_en
```

which I think effects both inputs and output channels, but might break the kernel ABI? or is libiio just not handling `in_out` channels?

?

mhennerich:
This convention is used for things that are both directions. It's only used once. It's not mainline and thus libiio doesn't handle it. I would have reverted it but I didn't want to break the API.
We have some workaround here:
https://github.com/analogdevicesinc/libad9361-iio/blob/master/ad9361_baseband_auto_rate.c#L64

rgetz:
OK - I will keep this open until we decide that the "in_out" (bidirectional attributes) are going upstream (and we should extend libiio) or if they aren't (and we should just leave it alone).

-Robin

rgetz:
Ok - now iio_info gives:

```
                        out:  (input, WARN:iio_channel_get_type()=UNKNOWN)
                        1 channel-specific attributes found:
                                attr  0: voltage_filter_fir_en value: 0
```

as a polite reminder...

rgetz:
closing - since this is a kernel driver issue.

PR #678, opened 2021-04-26
IIOD: Drop dependency on private headers

pcercuei:
With this patchset, IIOD does not depend anymore on any of the private functions and structures of libiio, and only accesses everything through the public API.

One "functionality" of IIOD has been dropped in the process: the support for IIO devices that are both input and output. I think this case never happened and will never happen, so it should be OK.

### DIFF
--- a/iio-private.h
+++ b/iio-private.h
@@ -11,21 +11,7 @@
 
 /* Include public interface */
 #include "iio.h"
-
-#ifdef _WIN32
-#   ifdef LIBIIO_EXPORTS
-#	define __api __declspec(dllexport)
-#   else
-#	define __api __declspec(dllimport)
-#   endif
-#elif __GNUC__ >= 4
-#   define __api __attribute__((visibility ("default")))
-#else
-#   define __api
-#endif
-
 #include "iio-backend.h"
-
 #include "iio-config.h"
 
 #include <stdbool.h>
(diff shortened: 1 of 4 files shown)


pcercuei:
Good to know, thanks.

rgetz:
> One "functionality" of IIOD has been dropped in the process: the support for IIO devices that are both input and output. I think this case never happened and will never happen, so it should be OK.

I think there are devices like this on AD936x and described in #373 
`in_out_voltage_filter_fir_en`

commodo:
> > One "functionality" of IIOD has been dropped in the process: the support for IIO devices that are both input and output. I think this case never happened and will never happen, so it should be OK.
> 
> @commodo mainlined multiple buffer support. This was done in preparation for the output buffer support. So there can/will be iio devices that are both input and output.

yes, Jonathan wanted multi-buffer support before output-buffer support;
so, we had to do multi-buffer first;

I may need to talk to Lars about his status on the output-buffer stuff; he said he wanted to do some other approach like dmabuf.
But there hasn't been any update on that.

pcercuei:
libiio / IIOD has never been tested with such device, as far as I know, was it?

This code I'm removing uses the *same* iio_buffer for both input and output, I believe when we do officially support it, we'd have one iio_buffer for input and one for output.

rgetz:
> libiio / IIOD has never been tested with such device, as far as I know, was it?

AD9361 is such a device - isn't it? It uses a in_out type. (but it is for attributes, not buffers)

pcercuei:
> > libiio / IIOD has never been tested with such device, as far as I know, was it?
> 
> AD9361 is such a device - isn't it? It uses a in_out type. (but it is for attributes, not buffers)

Here by "such device" I am talking about a iio_device that does support two-way buffers, yes. The code I am removing is _supposed_ to support a iio_buffer used  both for transmit and receive, which we never have seen in the wild, and could never test with.

mhennerich (approved):

PR #679, opened 2021-04-26
xml: Cleanup libxml2 parser when exiting library

pcercuei:
When libiio is unloaded (when the client program exits), cleanup libxml2
so that memory analyzer tools like Valgrind won't detect a memory leak.


rgetz (approved):

rgetz:
thanks - this looks much cleaner than what I was suggesting...

PR #680, opened 2021-05-05
CMake scripts cleanups

pcercuei:
A set of patches to cleanup the CMake scripts.

This change is a bit more invasive than usual, as it changes the behaviour of the script. For that reason, it is possible that the autobuilders will fail, as their scripts may have to be enabled.

Instead of auto-enabling functions according to the environment (available libraries etc), the options are now never auto-selected, and the environment is probed according to the enabled options.

The goal of this is to make the builds reproductible: now, using the same configuration on two systems will result in the same features being enabled in the library.

The scripts have also been updated to use the 3.x CMake features.


rgetz:
> The scripts have also been updated to use the 3.x CMake features.

yeah, the still shipping Zynq Images use old cmake - this would break too many boards in the field.

The new Kuiper images that we are moving to use 3.13; I don't think that's the default image yet.
```
root@analog:~# cmake --version
cmake version 3.13.4
```

plus it causes all CI builds to fail. Centos like this:
```
CMake Error at cmake/Install.cmake:177 (include):
  include could not find load file:

    LinuxPackaging.cmake
  CMakeLists.txt:425 (include)
```

nunojsa:
> yeah, the still shipping Zynq Images use old cmake - this would break too many boards in the field.

FWIW, cmake 3.20.2 is already complaining about the used version...

```
CMake Deprecation Warning at tests/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

iio-config.h.cmakein
```diff
@@ -17,9 +17,9 @@
 #cmakedefine01 WITH_NETWORK_EVENTFD
 #cmakedefine01 WITH_IIOD_USBD
 #cmakedefine01 WITH_LOCAL_CONFIG
+#cmakedefine01 WITH_AIO
```
nunojsa, line 20: Just a comment which has noting to do with this... We should definitely thing about replacing aio for io_uring and liburing  for libiio2 :smile:

iiod/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ find_library(LIBAIO_LIBRARIES aio)
 find_path(LIBAIO_INCLUDE_DIR libaio.h)
 
 if (LIBAIO_LIBRARIES AND LIBAIO_INCLUDE_DIR)
-	option(ENABLE_AIO "Build IIOD with async. I/O support" ON)
+	option(WITH_AIO "Build IIOD with async. I/O support" ON)
```
nunojsa: Does this mean that we only provide the option if the libraries are installed in the system? If so, I'm not sure if it would not be better to always provide the option and look for the libraries if the option is enabled. Then make it explicit that dependencies are needed (cmake failure)... Otherwise, users might not even be aware that this feature is possible...

CMakeLists.txt
```diff
@@ -145,31 +138,32 @@ if (ENABLE_IPV6)
...
+option(WITH_USB_BACKEND "Enable the libusb backend" ON)
+if (WITH_USB_BACKEND)
+	#Handle FreeBSD libusb and Linux libusb-1.0 libraries
+	find_library(LIBUSB_LIBRARIES NAMES usb-1.0 usb)
+	find_path(LIBUSB_INCLUDE_DIR libusb.h PATH_SUFFIXES libusb-1.0)
+	if (NOT LIBUSB_LIBRARIES OR NOT LIBUSB_INCLUDE_DIR)
+		message(SEND_ERROR "Unable to find libusb-1.0 dependency.\n"
+			"If you want to disable the USB backend, set WITH_USB_BACKEND=OFF.")
+	endif()
+
```
nunojsa, line 128: Hmm, you can disregard my previous comment :+1:

nunojsa (approved):
Apart from the cmake version bump question, looks a very good cleanup to me

iio-config.h.cmakein
pcercuei: Yes, I will definitely have a look at io_uring. I want to see if it can help us with zero-copy.

pcercuei:
> > The scripts have also been updated to use the 3.x CMake features.
> 
> yeah, the still shipping Zynq Images use old cmake - this would break too many boards in the field.

Isn't there a "backports" feature like on Ubuntu/Debian?

pcercuei:
@rgetz Now would maybe be a good time to branch libiio2. Then we can up the CMake requirements in the libiio2 branch.

iio-config.h.cmakein
nunojsa: It has both ring buffers shared between kernel and userspace so that's something already in the right way... Another neat thing is that we can potential reduce a lot the number of syscalls.

pcercuei: But I assume the IIO subsystem needs to support it (does it?).

rgetz:
> @rgetz Now would maybe be a good time to branch libiio2. Then we can up the CMake requirements in the libiio2 branch.

that makes total sense to me.

iio-config.h.cmakein
nunojsa: Don't think so. At least for normal reads/writes on sysfs stuff. Now for ioctl (not sure if that stuff was merged; and yes, it seems io_uring is also allowing to re-implement ioctl), the last time I checked I think IIO would have to support it...

nunojsa: https://lwn.net/Articles/844875/ - not idea the outcome of the RFC

rgetz:
but the CI failures still should be fixed, or you won't have much test coverage going forward...

pcercuei:
@rgetz yes. Don't worry ;)

pcercuei:
Alright, it should be ready for merging now.

Note that USB support in IIOD is disabled in CentOS 7, and UART support and ZSTD compression are disabled on Ubuntu 16.04, as the corresponding dependencies are too old on these OSes.

PR #684, opened 2021-05-21
CMakeLists.txt: Disable WITH_SERIAL_BACKEND by default

adisuciu:
https://ci.appveyor.com/project/analogdevicesinc/scopy/builds/39249068/job/yh1k2u4t425vsxp5

We got this error when building libiio. We could of course set cmake with -DWITH_SERIAL_BACKEND=OFF and it would work. However in my opinion the serial backend should not be enabled by default, but rather be specifically-enabled by the user (the user will know he wants to use the serial backend)

Signed-off-by: Adrian Suciu <adrian.suciu@analog.com>

### DIFF
--- a/CI/travis/make_linux
+++ b/CI/travis/make_linux
@@ -48,7 +48,7 @@ handle_default() {
 	if [ "${OS_VERSION}" = xenial ] ; then
 		FLAGS="${FLAGS} -DWITH_SERIAL_BACKEND=OFF"
 	else
-		FLAGS="${FLAGS} -DWITH_ZSTD=ON"
+		FLAGS="${FLAGS} -DWITH_SERIAL_BACKEND=ON -DWITH_ZSTD=ON"
 	fi
 
 	echo "### cmake ${FLAGS}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,13 +250,12 @@ if(WITH_LOCAL_BACKEND)
 	endif()
 endif()
 
-option(WITH_SERIAL_BACKEND "Enable the serial backend" ON)
+option(WITH_SERIAL_BACKEND "Enable the serial backend" OFF)
 if (WITH_SERIAL_BACKEND)
 	find_library(LIBSERIALPORT_LIBRARIES serialport)
 	find_path(LIBSERIALPORT_INCLUDE_DIR libserialport.h)
 	if (NOT LIBSERIALPORT_LIBRARIES OR NOT LIBSERIALPORT_INCLUDE_DIR)
-		message(SEND_ERROR "Unable to find libserialport dependency.\n"
-			"If you want to disable the serial backend, set WITH_SERIAL_BACKEND=OFF.")
+		message(SEND_ERROR "Unable to find libserialport dependency.\n")
 	else()
 		message(STATUS "Looking for libserialport : Found")
 


pcercuei:
I tend to agree, any other opinions on this?

nunojsa:
> I tend to agree, any other opinions on this?

Me either... I actually came across the same issue in the libiio yocto recipe. I just pushed a [fix](https://github.com/openembedded/meta-openembedded/pull/359/commits/a9d61725f0d14159ce03765c8a167f523fd503c7) where the serial_backend is added and enabled by default.

nunojsa (approved):

pcercuei (approved):

dNechita:
"the serial backend should not be enabled by default, but rather be specifically-enabled by the user (the user will know he wants to use the serial backend)"
Shouldn't this logic be applied to the rest of the backends? On which criteria is serial backend treated differently? Is is less common?

nunojsa:
> Shouldn't this logic be applied to the rest of the backends? On which criteria is serial backend treated differently? Is is less common?

I actually thought about that for the USB backend. I guess the network one makes sense to be enabled by the default. My naive assumption is that the network one is the most commonly used. Anyways, worst case scenario, I think at least one backend should be enabled by default...

pcercuei:
@dNechita the default configuration should cover the typical use-cases, so (I think) local / USB / network backends.

dNechita:
> @dNechita the default configuration should cover the typical use-cases, so (I think) local / USB / network backends.

I guess it makes sense.
I'm ok with these changes.

mhennerich (approved):

buha:
So I just came across this because I noticed building latest master doesn't include serial backend by default.
The build issue should've been fixed by fixing the code that creates the issue, not by removing serial from defaults.
Removing usb or serial from defaults is arbitrary - There are benefits of having them all enabled (ease of use) but I can't think of any meaningful benefit of removing either one from defaults... IMHO

Even if the argument is that serial is not that much used, there is a counterargument for that too - no-OS project relies exclusively on serial backend to interact with iio clients.

nunojsa:
I see your point but I have to agree with this:

> the default configuration should cover the typical use-cases, so (I think) local / USB / network backends.

That is what I would expect from any project rather than have everything enabled by default. Just to have the build complaining about a lot of missing dependencies just so I have to disable all these features manually... So yes, I actually think this is key:

>  Even if the argument is that serial is not that much used...

Now the question would be. Is no-Os such an important user of libiio that justifies having serial enabled by default? Not my question to answer  :).

buha:
I mean, this is turning up into a philosophical debate about the importance of no-OS or serial backend so I'll try to come back to the original topic.

Does this PR fix the original problem ? No
Does it create a workaround for it ? Yes

pcercuei:
The PR isn't so much a workaround for a build issue, but the build issue made us consider the change, if that makes sense.

nunojsa:
> I mean, this is turning up into a philosophical debate about the importance of no-OS...

Don't take me wrong here. I did not want to minimize no_Os  at all. It was an honest question without any meaning or opinion of my own :)

> Does this PR fix the original problem ? No
> Does it create a workaround for it ? Yes

Well, and what would be the fix for the original problem? I guess either  -DWITH_SERIAL_BACKEND=OFF or installing libserial which is exactly what @adisuciu did not want to do... I think the PR description is very clear on why this option was taken which is definitely not a workaround but a decision that was made (that can obviously be reverted at any time).

buha:
-DWITH_SERIAL_BACKEND=OFF only on scopy whan doing cmake would've been a much better choice.

This would've kept the whole discussion in the scopy world.

Now, there is the whole Cluj and Phillipines team working with no-OS that has to document and adapt to the fact that if they need to  -DWITH_SERIAL_BACKEND=ON whenever building libiio. There are possible implications on Kuiper too.

dNechita:
The windows builds of IIO-Oscilloscope use libiio with serial backend enabled.
Just mentioning this in case it helps getting a better understanding where libiio with serial backend is also used.

nunojsa:
> -DWITH_SERIAL_BACKEND=OFF only on scopy whan doing cmake would've been a much better choice.

That is arguable. And it looks like the reviewers of this PR actually agreed with what was purposed. I just want to clarify that this is not a workaround. It was a very aware decision (admittedly not remembering no-OS; at least me).

> -DWITH_SERIAL_BACKEND=OFF only on scopy whan doing cmake would've been a much better choice.
> This would've kept the whole discussion in the scopy world.

I'm not seeing the issue here... it's pretty much what you're suggesting for scopy just on the reverse order :). Anyways, I don't really mind if serial is enabled or not by default. I was just arguing that it actually maters that only the most used features should be enabled by default...

I guess @pcercuei is the one you have to convince to have this back...

pcercuei:
This is not about Scopy, it's about having sane defaults.

Keep in mind that most people simply use binary packages (on Linux and of course Windows) and won't compile from sources. Binary packages will still provide the serial backend.

PR #686, opened 2021-05-24
network: Include headers for the errors that are used

dNechita:
Because of the missing headers, build failed when compiling on Ubuntu1804
image supported by Appveyor, with MinGW compiler.

### DIFF
--- a/dns_sd.c
+++ b/dns_sd.c
@@ -16,11 +16,11 @@
 #include "iio-lock.h"
 #include "iio-private.h"
 
+#include <errno.h>
 #ifdef _WIN32
 #include <winsock2.h>
 #include <ws2tcpip.h>
 #else
-#include <errno.h>
 #include <netdb.h>
 #include <string.h>
 #include <sys/socket.h>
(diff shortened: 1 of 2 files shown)


pcercuei:
@dNechita the move of `errno.h` in dns_sd.c is needed too?

dNechita:
> @dNechita the move of `errno.h` in dns_sd.c is needed too?

Yes

pcercuei:
@dNechita alright, strange. My assumption was that `winsock2.h` and `ws2tcpip.h` included all the headers they needed. Looks like a problem on MinGW's side.

`errno.h` is a standard POSIX header, MSVC should have it too.

dNechita:
> @dNechita alright, strange. My assumption was that `winsock2.h` and `ws2tcpip.h` included all the headers they needed. Looks like a problem on MinGW's side.
> 
> `errno.h` is a standard POSIX header, MSVC should have it too.

Looks like MSVC2019 build has passed. This means error.h is available for MSVC compiler. I guess these changes are ok as they are now. Right?

pcercuei (approved):

PR #690, opened 2021-05-26
Push artifacts to SW Downloads

ccraluca:
This Azure Pipeline was divided into two stages:
- first stage is for builds
- second stage runs at the end of the first stage for download and push the artifacts into SW Downloads in the 'azure_builds' directory

Artifacts are overwritten in SW Downloads only when Azure Pipeline is running for the master branch.

Signed-off-by: Raluca Chis <raluca.chis@analog.com>

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,78 +9,80 @@ pr:
 - master
 - 20*
 
-jobs:
-- job: LinuxBuilds
-  strategy:
-    matrix:
-      centos_7_x86_64:
-        imageName: 'ubuntu-latest'
-        OS_TYPE: 'centos_docker'
-        OS_VERSION: centos7
-        artifactName: 'Linux-CentOS-7-x86_64'
-      centos_8_x86_64:
-        imageName: 'ubuntu-latest'
-        OS_TYPE: 'centos_docker'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -9,148 +9,169 @@ pr:
...
+    steps:
+      - task: DownloadPipelineArtifact@2
+        inputs:
+          path: /home/vsts/work/1/a
+      - task: DownloadSecureFile@1
+        name: key
+        displayName: 'Download rsa key'
+        inputs:
+          secureFile: 'id_rsa'
+      - bash: chmod 600 $(key.secureFilePath) ; scp -2 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o HostKeyAlgorithms=+ssh-dss -vv -i $(key.secureFilePath) -r /home/vsts/work/1/a/* sshacs@analog.ssh.upload.akamai.com:/336936/cse/azure_builds
```
tfcollins, line 175: Can you make the server address a secret as well.

ccraluca: Sure, I set the address as a secret variable.

tfcollins (approved):

PR #691, opened 2021-05-26
Add support for dynamically loaded backends

pcercuei:
Hi,

This PR adds support for loading backends as external modules loaded from the disk.

Using modules instead of built-in backends has two big benefits:
- it makes it possible for third-parties to easily implement a backend;
- it makes it possible for packagers to separate backends into their own packages, and for users to selectively enable what they need, without having to install the dependencies of backends they don't need.

The backends to be exported as modules must only depend on the public API of libiio, which consists in `iio.h`, and `iio/iio-backend.h`, `iio/iio-lock.h` and `iio/iiod-client.h` for the plugin-facing part.

To enable support for modules, the configuration option `WITH_MODULES` must be set to ON (it is OFF by default). Then, individual backends can be switched between built-in and module. For now, only the serial backend supports being installed as a module, with the option `WITH_SERIAL_BACKEND_DYNAMIC` (ON by default). The other backends still depend heavily on the private API and need to be updated.


pcercuei:
I updated the PR. The dynamic backend can now be enabled on Windows as well.

pcercuei:
Feedback on this?

mhennerich (approved):

PR #702, opened 2021-06-15
Network/usb modules

pcercuei:
Cleanup the network and USB backends so that they don't depend on the private libiio headers, and update both backends so that they can now be built as external modules.


CMakeLists.txt
```diff
@@ -376,11 +376,25 @@ if(WITH_NETWORK_BACKEND)
 		endif()
 	endif()
 
-	target_sources(iio PRIVATE network.c)
```
mchindri, line 363: Commint description wasn't changes. Is the one for usb.

pcercuei: Looks like a bad copy/paste in the commit message.

pcercuei: Updated.

mchindri (approved):

PR #703, opened 2021-06-15
Improve xml sanitization

mchindri:
Create function for repeated code and sanitize where needed.

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -170,12 +170,23 @@ void iio_channel_init_finalize(struct iio_channel *chn)
 static ssize_t iio_snprintf_chan_attr_xml(char *str, ssize_t len,
 					  struct iio_channel_attr *attr)
 {
+	ssize_t ret, alen = 0;
+
 	if (!attr->filename)
 		return iio_snprintf(str, len, "<attribute name=\"%s\" />", attr->name);
 
-	return iio_snprintf(str, len,
-			    "<attribute name=\"%s\" filename=\"%s\" />",
-			    attr->name, attr->filename);
+	ret = iio_snprintf(str, len, "<attribute name=\"%s\" ", attr->name);
+	if (ret < 0)
+		return ret;
+
+	iio_update_xml_indexes(ret, &str, &len, &alen);
+
+	ret = iio_xml_print_and_sanitized_param(str, len, "filename=\"",
+						attr->filename, "\" />");
+	if (ret < 0)
+		return ret;
+
+	return alen + ret;
 }
 
 static ssize_t iio_snprintf_scan_element_xml(char *str, ssize_t len,
@@ -204,7 +215,9 @@ ssize_t iio_snprintf_channel_xml(char *ptr, ssize_t len,
 	ssize_t ret, alen = 0;
 	unsigned int i;
 
-	ret = iio_snprintf(ptr, len, "<channel id=\"%s\"", chn->id);
+
+	ret = iio_xml_print_and_sanitized_param(ptr, len, "<channel id=\"",
+						chn->id, "\"");
 	if (ret < 0)
 		return ret;
 	iio_update_xml_indexes(ret, &ptr, &len, &alen);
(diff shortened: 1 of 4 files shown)


pcercuei:
You can use `sanitize_xml` from context.c.

mchindri:
v2:
 - use sanitize_xml

channel.c
```diff
@@ -167,15 +167,26 @@ void iio_channel_init_finalize(struct iio_channel *chn)
 	}
 }
 
+ssize_t sanitize_xml(char *ptr, ssize_t len, const char *str);
```
pcercuei, line 170: Move this to iio-private.h.

pcercuei (changes requested):

channel.c
```diff
@@ -203,8 +214,14 @@ ssize_t iio_snprintf_channel_xml(char *ptr, ssize_t len,
...
 	unsigned int i;
+	char	buf[100];
 
-	ret = iio_snprintf(ptr, len, "<channel id=\"%s\"", chn->id);
+	if (strchr(chn->id, '&')) {
+		sanitize_xml(buf, 100, chn->id);
+		ret = iio_snprintf(ptr, len, "<channel id=\"%s\"", buf);
+	} else {
+		ret = iio_snprintf(ptr, len, "<channel id=\"%s\"", chn->id);
+	}
```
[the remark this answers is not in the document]
mchindri: I did a little refactor with taking in consideration this suggestion.

mchindri:
v3:
 - Update PR name and description
 - Refactor PR and some iiod code in order to be easy to implement requested changes and future ones.

context.c
```diff
@@ -34,6 +34,15 @@ static const char xml_header[] = "<?xml version=\"1.0\" encoding=\"utf-8\"?>"
...
 "]>";
 
+void iio_update_xml_indexes(ssize_t ret, char **ptr, ssize_t *len, ssize_t *alen)
+{
+	if (*ptr) {
+		*ptr += ret;
+		*len -= ret;
+	}
+	*alen += ret;
+}
```
pcercuei, line 44: I don't like this. The fact that it takes arguments as pointers and update them.

... But I like the +++/--- diff ;)

Maybe you could:
- have it as a static inline in iio-private.h, since this is really just a tiny helper;
- "alen" could be dropped, if the functions keep the original value of ptr as e.g. "buf" and `return (uintptr_t) ptr - (uintptr_t) buf;`. I have a small preference for that, but not a deal-breaker either.

iio-private.h
```diff
@@ -271,6 +271,10 @@ int add_iio_dev_attr(struct iio_dev_attrs *attrs, const char *attr,
 
 ssize_t __iio_printf iio_snprintf(char *buf, size_t len, const char *fmt, ...);
 
+ssize_t iio_xml_print_and_sanitized_param(char *ptr, ssize_t len,
+					  const char *before, char *param,
+					  const char *after);
+
```
pcercuei, line 277: Move this to the 2nd commit, since it's where the function is added.

context.c
```diff
@@ -100,18 +126,15 @@ static ssize_t iio_snprintf_context_xml(char *ptr, ssize_t len,
 
 	for (i = 0; i < ctx->nb_attrs; i++) {
 		ret = iio_snprintf(ptr, len,
-				   "<context-attribute name=\"%s\" value=\"",
+				   "<context-attribute name=\"%s\" ",
```
pcercuei, line 129: You could keep it here, no? Then your iio_xml_print_and_sanitized_param wouldn't need a prefix.

pcercuei (changes requested):

context.c
mchindri: I could, but I though is better next to where the value is set.
I mean the first printf set's the name and the next the value.

mchindri:
v4:
 - move iio_xml_print_and_sanitized_param from iio-private.h in previous commit
 - make iio_update_xml_indexes static inline in iio_private.h

context.c
mchindri: I don't like it neither updating variables in functions, but is the best solution that came to my mind to remove that code duplication. 
Regarding your second suggestion, ptr the first time is null so it wont work. I think that's why alen is used.

pcercuei (approved):

pcercuei:
Merged, thanks!

PR #704, opened 2021-06-18
examples: add example for adrv9002

nunojsa:
This is analogous to adrv9009-iiostream. Just serves as an example on how
to stream data on adrv9002 using libiio.

Signed-off-by: Nuno Sá <nuno.sa@analog.com>

### DIFF
--- a/examples/adrv9002-iiostream.c
+++ b/examples/adrv9002-iiostream.c
@@ -1,0 +1,308 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * libiio - Library for interfacing industrial I/O (IIO) devices
+ *
+ * Copyright (C) 2021 Analog Devices, Inc.
+ * Author: Nuno Sá <nuno.sa@analog.com>
+ */
+#include <iio.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <signal.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include <stdbool.h>
+
+#define ARGS(fmt, ...)	__VA_ARGS__
+#define FMT(fmt, ...)	fmt
+#define error(...) \
+	printf("%s, %d: ERROR: " FMT(__VA_ARGS__, 0)"%s", __func__, __LINE__, ARGS(__VA_ARGS__, ""))
+
+#define info(...) \
+	printf("%s, %d: INFO: " FMT(__VA_ARGS__, 0)"%s", __func__, __LINE__, ARGS(__VA_ARGS__, ""))
+
+/* helper macros */
+#define GHZ(x) ((long long)(x * 1000000000.0 + .5))
+
+static bool stop = false;
+static struct iio_context *ctx = NULL;
+static struct iio_buffer *rxbuf = NULL;
+static struct iio_buffer *txbuf = NULL;
+static struct iio_channel *rx_chan[2] = { NULL, NULL };
+static struct iio_channel *tx_chan[2] = { NULL, NULL };
+
+enum {
+	I_CHAN,
+	Q_CHAN
+};
+
+#ifdef _WIN32
+#include <windows.h>
+#include <io.h>
+#include <fcntl.h>
+
+BOOL WINAPI sig_handler(DWORD dwCtrlType)
+{
+	/* Runs in its own thread */
+	switch (dwCtrlType) {
+	case CTRL_C_EVENT:
+	case CTRL_CLOSE_EVENT:
+		stop = true;
+		return true;
+	default:
+		return false;
+	}
+}
+
+static int register_signals(void)
+{
+	if (!SetConsoleCtrlHandler(sig_handler, TRUE))
+		return -1;
+
+	return 0;
+}
+#else
+static void sig_handler(int signum)
+{
+	if (signum == SIGINT || signum == SIGTERM) {
+		info("Exit....\n");
+		stop = true;
+	}
+}
+
+static int register_signals(void)
+{
+	struct sigaction sa = {0};
+	sigset_t mask = {0};
+
+	sa.sa_handler = sig_handler;
+	sigemptyset(&sa.sa_mask);
+	sigemptyset(&mask);
+
+	if (sigaction(SIGTERM, &sa, NULL) < 0) {
+		error("sigaction: %s\n", strerror(errno));
+		return -1;
+	}
+
+	if (sigaction(SIGINT, &sa, NULL) < 0) {
+		error("sigaction: %s\n", strerror(errno));
+		return -1;
+	}
+
+	sigaddset(&mask, SIGINT);
+	sigaddset(&mask, SIGTERM);
+	/* make sure these signals are unblocked */
+	if (sigprocmask(SIG_UNBLOCK, &mask, NULL)) {
+		error("sigprocmask: %s", strerror(errno));
+		return -1;
+	}
+
+	return 0;
+}
+#endif
+
+static int configure_tx_lo(void)
+{
+	struct iio_device *phy;
+	struct iio_channel *chan;
+	int ret;
+	long long val;
+
+	phy = iio_context_find_device(ctx, "adrv9002-phy");
+	if (!phy) {
+		error("Could not find adrv9002_phy\n");
+		return -ENODEV;
+	}
+
+	chan = iio_device_find_channel(phy, "voltage0", true);
+	if (!chan) {
+		error("Could not find TX voltage0 channel\n");
+		return -ENODEV;
+	}
+
+	/* printout some useful info */
+	ret = iio_channel_attr_read_longlong(chan, "rf_bandwidth", &val);
+	if (ret)
+		return ret;
+
+	info("adrv9002 bandwidth: %lld\n", val);
+
+	ret = iio_channel_attr_read_longlong(chan, "sampling_frequency", &val);
+	if (ret)
+		return ret;
+
+	info("adrv9002 sampling_frequency: %lld\n", val);
+
+	/* set the LO to 2.5GHz */
+	val = GHZ(2.5);
+	chan = iio_device_find_channel(phy, "altvoltage2", true);
+	if (!chan) {
+		error("Could not find TX LO channel\n");
+		return -ENODEV;
+	}
+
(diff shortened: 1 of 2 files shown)


examples/adrv9002-iiostream.c
```diff
@@ -0,0 +1,322 @@
...
+			return;
+		}
+
+		/* READ: Get pointers to RX buf and read IQ from RX buf port 0 */
+		p_inc = iio_buffer_step(rx);
+		p_end = iio_buffer_end(rx);
+		for (p_dat = iio_buffer_first(rx, rx_i_chan); p_dat < p_end; p_dat += p_inc) {
+			/* Example: swap I and Q */
+			int16_t i = get_unaligned(p_dat);
+			int16_t q = get_unaligned(&p_dat[2]);
```
pcercuei, line 225: Why the unaligned access?

nunojsa: I took this from [here](https://github.com/analogdevicesinc/libiio/blob/master/examples/adrv9009-iiostream.c#L269). On that code we might clearly have some misaligned access since we have no guarantees that p_dat sits on a address aligned with a short.  The compiler will also complain about that code (using `-Wcast-align`)...

If you do not want to have these helpers, plain `memcpy()` should also make the compiler happy. Or if you can guarantee (as you know the libiio internals better than me :)) that those addresses will always have some alignment guarantees, I can just use the cast as in the adrv9009 example (though I think that is a bad practice).

pcercuei: The buffers and the samples are guaranteed to be aligned.

If you make `p_dat` a `int16_t *`, then the compiler won't complain and you can just read i/q without casting the pointer.

nunojsa: Ok, that works as long as I also do the update like `p_dat += p_inc / 2`... I guess p_inc is the complete size of the enabled scan elements?

pcercuei: Yes, `p_inc` is the sample size in bytes.
You can update with: `p_dat += p_inc / sizeof(*p_dat)` just so that it's obvious what it does.

nunojsa:
V2:

* Drop unnecessary usage of `{get/put}_unaligned()` helpers.

examples/adrv9002-iiostream.c
```diff
@@ -0,0 +1,304 @@
...
+		/* WRITE: Get pointers to TX buf and write IQ to TX buf port 0 */
+		p_inc = iio_buffer_step(tx);
+		p_end = iio_buffer_end(tx);
+		for (p_dat = iio_buffer_first(tx, tx_i_chan); p_dat < p_end;
+		     p_dat += p_inc / sizeof(*p_dat)) {
+			/*
+			 * Example: fill with zeros
+			 * 14-bit sample needs to be MSB aligned so shift by 2
+			 * https://wiki.analog.com/resources/eval/user-guides/ad-fmcomms2-ebz/software/basic_iq_datafiles#binary_format
+			 */
```
tfcollins, line 221: Remove comment copied from ad9361-iiostream

nunojsa:
V3:

* Removed copied pasted comment.

examples/adrv9002-iiostream.c
```diff
@@ -0,0 +1,299 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include <stdbool.h>
+
+#define error(fmt, ...) \
+	printf("%s, %d: "fmt, __func__, __LINE__, ##__VA_ARGS__)
+
+#define info(fmt, ...) \
+	printf("%s, %d: "fmt, __func__, __LINE__, ##__VA_ARGS__)
```
pcercuei, line 36: OSX doesn't like your `##__VA_ARGS__`...

examples/adrv9002-iiostream.c
```diff
@@ -0,0 +1,299 @@
...
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
```
pcercuei, line 19: You can strip that wall of text and use a SPDX license header, have a look at the C files in the top directory.

pcercuei (changes requested):
Just a couple of changes needed, right now it fails to build on OSX (see the autobuilders)

examples/adrv9002-iiostream.c
nunojsa: Arghhh, I guess I will also have to drop my neat cleanup handling?

https://dev.azure.com/AnalogDevices/OpenSource/_build/results?buildId=4022&view=logs&j=afc9f219-4f50-5293-48d5-d83ea574007e&t=470c08d8-8808-547a-c046-cb3882e7f5bd&l=185

pcercuei: Shouldn't you use `atexit()`?

nunojsa: Hmm the reason why I did not used it is because I cannot pass any argument to the cb... But for this example I can always make the objects global and that solves it! I will update it together with the variadic macro...

nunojsa: Well, it actually makes it very cumbersome with `atexit()`. I will just drop this and add a normal cleanup function...

pcercuei: Have a look at iio-debug.h in the dev branch, there are macros that basically do the same thing and are cross-platform.

nunojsa: Just gave it a look but I do not think I can use it directly without some tweaks. The thing is that on the format I have, I want to prepend `__func__` and `__LINE__` to my format string. So, I cannot guarantee that `__VA_ARGS__` will expand to something as you can [here](https://github.com/analogdevicesinc/libiio/blob/dev/iio-debug.h#L85) with `buf`. First thing that comes to my mind is to just add a dummy empty string at the end of my format so that I can nicely extract my arguments. The other one is just using two printfs like:

```
#define error(...) \
        printf("%s, %d: ERROR: ", __func__, __LINE__); \
        printf(__VA_ARGS__);
```
Anyways, I just don't want to wast to much time on this... As I don't have HW with me today I will check this tomorrow. I want to do a sanity check before pushing this update...

pcercuei:
```
#define __error(line, ...) printf(__func__ ", " #line __VA_ARGS__);
#define _error(line, ...) __error(line, __VA_ARGS__);
#define error(fmt, ...) _error(__LINE__, fmt, __VA_ARGS__);
```

That doesn't work?

nunojsa: Nope, I made a quick test and it does not compile. I'm not sure we can have `printf(__func__"plus something")`... What I have in mind for the dummy string is something like:

```
#define ARGS(fmt, ...)  __VA_ARGS__
#define FMT(fmt, ...)   fmt
#define error(...) \
       printf("%s, %d: ERROR: " FMT(__VA_ARGS__)"%s", __func__, __LINE__, ARGS(__VA_ARGS__, ""))
```

I could also just use `\n` instead of the empty string. I do not like much to enforce that but as this is an example app I don't really mind (if you prefer it...)

pcercuei: @nunojsa 
```
#define __FIRST(a, ...)		a
#define ___OTHERS(a, ...)	a, __VA_ARGS__
#define __OTHERS(a, b, ...)	___OTHERS(a, __VA_ARGS__)
#define error(...) \
	printf("%s, %d: ERROR: " __FIRST(__VA_ARGS__, 0), __func__, __OTHERS(__LINE__,__VA_ARGS__))
```

nunojsa: Does not works for all cases:

```
int n = 10;
error("Hello world: %d\n", n); // works fine
error("Hello world\n"); //compilation error
```
I think the issue with the later is that __VA_ARGS__ will evaluate to nothing in `___OTHERS` which is not nicely handled unless we use the gnu extension. That is why I think the "dummy" string after the format is needed (at least is one way to handle it; maybe not the best one).

Out of curiosity, is there any special reason for you to use `__FIRST(__VA_ARGS__, 0)` (also saw it in iio-debug.h)? If I'm not mistaken, `__FIRST(__VA_ARGS__)` should also work...

pcercuei: C requires that the number of parameters in a macro's "..." is >= 1. iio-debug.h does `__FIRST(__VA_ARGS__, 0)` otherwise the compiler complains.

pcercuei: Following your suggestion:
```
#include <stdio.h>

#define __FIRST(a, ...)		a
#define ___OTHERS(a, ...)	a, __VA_ARGS__
#define __OTHERS(a, b, ...)	___OTHERS(a, __VA_ARGS__)

#define error(...) \
	printf("%s, %d: ERROR: " __FIRST(__VA_ARGS__, 0) "%c", __func__, __OTHERS(__LINE__,__VA_ARGS__, '\0'))

int main(void)
{
	error("Test: %d\n", 5);
	error("Test2\n");
}
```

This seems to compile fine.

nunojsa: > C requires that the number of parameters in a macro's "..." is >= 1. iio-debug.h does __FIRST(__VA_ARGS__, 0) otherwise the compiler complains.

Hmm, I see. I guess with gnu extensions this is not really the case. That is why gcc did not complained when I tried it...

nunojsa:
V4:

* Dropped `on_exit()` and added a normal cleanup function;
* Refactored the log macros to be portable (no use of gnu ##VA_ARGS);
* `sigaction()` is part of POSIX and Windows is not fully compliant with it. Hence, use windows API for handling signals;
* When using `sigaction()`, make sure that the registered signals are unblocked...

nunojsa:
V5:

* Use SPDX license header...

pcercuei (approved):

PR #710, opened 2021-07-08
Update udev rule to only scan USB

pcercuei:
Having "iio_info -s" in the udev rule will scan all available backends,
and some (like the network backend) may take a while to process.

Run "iio_info -S usb" instead to specify that we only want to scan USB
contexts.

@AlexandraTrifan can I add you as Reported-By?


mhennerich (approved):

AlexandraTrifan (approved):

AlexandraTrifan:
Sure!

mhennerich:
can we merge this into 2019_R2 as well?

pcercuei:
I suppose so, yes.

PR #713, opened 2021-07-12
Put library version in XML string

pcercuei:
Hi,

Since the library version on the remote side is not expected to change during a iio_context's life, it is static data and therefore should be passed as part of the XML string.

This PR makes it easier for me to introduce other changes (less code to update to the new APIs and protocols).

It has two notable implications:

- As usual when the XML string is updated, the version of the remote libiio must be equal or newer to the version of the PC (the inverse is not true).

- IIRC, iio-osc (or was it Scopy?) uses a thread that calls iio_context_get_version() in a loop with the sole purpose to detect broken communication. That won't work anymore. I think one safe replacement would be to use iio_device_get_trigger(), since this is guaranteed to be a call to IIOD.

Cheers,
-Paul

### DIFF
--- a/context.c
+++ b/context.c
@@ -272,6 +272,7 @@ void iio_context_destroy(struct iio_context *ctx)
 	free(ctx->devices);
 	free(ctx->xml);
 	free(ctx->description);
+	free(ctx->git_tag);
 	free(ctx->pdata);
 	free(ctx);
 }
--- a/context.c
+++ b/context.c
@@ -366,9 +366,6 @@ int iio_context_get_version(const struct iio_context *ctx,
 		return 0;
 	}
 
-	if (ctx->ops->get_version)
-		return ctx->ops->get_version(ctx, major, minor, git_tag);
-
 	iio_library_get_version(major, minor, git_tag);
 	return 0;
 }
--- a/context.c
+++ b/context.c
@@ -102,14 +102,21 @@ static ssize_t iio_snprintf_context_xml(char *ptr, ssize_t len,
 	ssize_t ret, alen = 0;
 	unsigned int i;
 
-	if (ctx->description)
-		ret = iio_snprintf(ptr, len, "%s<context name=\"%s\" "
-				   "description=\"%s\" >",
-				   xml_header, ctx->name, ctx->description);
-	else
-		ret = iio_snprintf(ptr, len, "%s<context name=\"%s\" >",
-				   xml_header, ctx->name);
+	ret = iio_snprintf(ptr, len, "%s<context name=\"%s\" ",
+			   xml_header, ctx->name);
+	if (ret < 0)
+		return ret;
 
+	iio_update_xml_indexes(ret, &ptr, &len, &alen);
+
+	if (ctx->description) {
+		ret = iio_xml_print_and_sanitized_param(ptr, len,
+							"description=\"",
+							ctx->description,
+							"\" >");
+	} else {
+		ret = iio_snprintf(ptr, len, ">");
+	}
 	if (ret < 0)
 		return ret;
 
--- a/context.c
+++ b/context.c
@@ -24,7 +24,8 @@ static const char xml_header[] = "<?xml version=\"1.0\" encoding=\"utf-8\"?>"
 "<!ELEMENT scan-element EMPTY>"
 "<!ELEMENT debug-attribute EMPTY>"
 "<!ELEMENT buffer-attribute EMPTY>"
-"<!ATTLIST context name CDATA #REQUIRED description CDATA #IMPLIED>"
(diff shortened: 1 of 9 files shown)


context.c
```diff
@@ -102,14 +102,21 @@ static ssize_t iio_snprintf_context_xml(char *ptr, ssize_t len,
...
+	iio_update_xml_indexes(ret, &ptr, &len, &alen);
+
+	if (ctx->description) {
+		ret = iio_xml_print_and_sanitized_param(ptr, len,
+							"description=\"",
+							ctx->description,
+							"\" >");
+	} else {
+		ret = iio_snprintf(ptr, len, ">");
+	}
```
nunojsa, line 127: nit: just to keep consistency (maybe the rest of libiio?) with the change we are doing here, I would not use brackets for single line `if()...else`. I know, linux coding style biased

context.c
```diff
@@ -100,10 +101,17 @@ static ssize_t iio_snprintf_context_xml(char *ptr, ssize_t len,
 					const struct iio_context *ctx)
 {
 	ssize_t ret, alen = 0;
-	unsigned int i;
+	unsigned int i, major, minor;
+	char git_tag[64];
+
+	ret = (ssize_t) iio_context_get_version(ctx, &major, &minor, git_tag);
```
nunojsa, line 107: curious: is the cast needed for other platforms (OSx or Windows)?

xml.c
```diff
@@ -421,17 +423,42 @@ static struct iio_context * iio_create_xml_context_helper(xmlDoc *doc)
...
-			description = (const char *)attr->children->content;
-		else if (strcmp((char *) attr->name, "name"))
+		content = (const char *) attr->children->content;
+
+		if (!strcmp((char *) attr->name, "description")) {
+			description = content;
+		} else if (!strcmp((char *) attr->name, "version-major")) {
+			major = strtol(content, &end, 10);
+			if (*end != '\0')
+				IIO_WARNING("unable to parse major version\n");
```
nunojsa, line 433: nit: I think this message might be a bit misleading. IIRC, this condition pretty much says that we have further characters on our string but the return value could actually be correct. eg; If the major string is something like "2foo", I think val would be 2 (which could be a valid value).

context.c
```diff
@@ -355,6 +355,17 @@ int iio_context_init(struct iio_context *ctx)
 int iio_context_get_version(const struct iio_context *ctx,
 		unsigned int *major, unsigned int *minor, char git_tag[8])
 {
+	if (ctx->git_tag) {
+		if (major)
+			*major = ctx->major;
+		if (minor)
+			*minor = ctx->minor;
+		if (git_tag)
+			iio_strlcpy(git_tag, ctx->git_tag, 8);
```
nunojsa, line 364: `sizeof(git_tag)`?

nunojsa (approved):
LGTM... Just some nits. I think [osc](https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/osc.c#L1636) is the one relying on this. I do have some plans to fully support IIO labels in osc so I can just fix things (while there) for this with your suggestion.

context.c
pcercuei: I'm linux coding style biaised too, if you have multi-line statements in the "if" block, then you need brackets in the "else" even if there is only one line.

xml.c
pcercuei: `invalid format for major version` sounds better?

context.c
pcercuei: That would work, as long as you execute it on a 64-bit OS ;)

Try this:
```
#include <stdio.h>

static void test(char array[16]) {
	char array2[16];

	printf("Size 1: %u size2: %u\n", sizeof(array), sizeof(array2));
}

int main(void)
{
	test(NULL);
	return 0;
}
```

context.c
pcercuei: Casts must be explicit when casting to a smaller type, a different qualifier, or a different signness. I guess in this case it could be implicit.

nunojsa: Yes, in this case the cast is implicit but I do not now if some compilers in other platforms would complain about this. That was why I asked

context.c
nunojsa: Hmm, yes. You can actually have it both ways. As this is still the same line just splitted in multiple lines, you can still remove the brackets. Adding them is also fine... forget this

context.c
nunojsa: Hmm, I see. I was actually mistaken by the argument which I'm not really a fan... I see that we are just doing the size of the pointer :sweat_smile:

pcercuei: Yes, it's a terrible design mistake I made when writing the old code. I'll update that in the 1.x libiio.

context.c
pcercuei: I removed this, and Windows didn't complain, so it's fine.

PR #718, opened 2021-07-22
bindings: python iio.py: find_device() also use label

mhennerich:
When multiple devices with the same name exist, label is used to
differentiate them. Update find_device to also support labels.

Actually not sure, why we're not calling the C API with that...

### DIFF
--- a/bindings/python/iio.py
+++ b/bindings/python/iio.py
@@ -1406,18 +1406,18 @@ class Context(object):
         """
         return Context(_clone(self._context))
 
-    def find_device(self, name_or_id):
+    def find_device(self, name_or_id_or_label):
         """
 
-        Find a IIO device by its name or ID.
+        Find a IIO device by its name, ID or label.
 
-        :param name_or_id: type=str
-            The name or ID of the device to find
+        :param name_or_id_or_label: type=str
+            The name, ID or label of the device to find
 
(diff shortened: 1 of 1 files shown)


bindings/python/iio.py
```diff
@@ -1406,18 +1406,18 @@ def clone(self):
         """
         return Context(_clone(self._context))
 
-    def find_device(self, name_or_id):
+    def find_device(self, name_or_id_or_label):
         """
 
-        Find a IIO device by its name or ID.
+        Find a IIO device by its name or ID or label.
```
pcercuei, line 1412: `name or ID or label` -> `name, ID or label`

bindings/python/iio.py
```diff
@@ -1406,18 +1406,18 @@ def clone(self):
...
 
-        :param name_or_id: type=str
-            The name or ID of the device to find
+        :param name_or_id_or_label: type=str
+            The name or ID or label of the device to find
```
pcercuei, line 1415: Same here.

pcercuei:
Looks good to me.

The C API would give you a handle to a C `struct iio_device`, which you would then have to match with all iio_py's Device objects to find the right one. So it wouldn't be any better than what's present right now.

mhennerich:
pushed v2:
 * fixed strings

pcercuei (approved):

PR #721, opened 2021-07-26
IIOD client work

pcercuei:
This set of commits to the iiod-client code prepares the terrain for the introduction of the async protocol.

First, the "high-speed" network interface is removed, since it was always disactivated by default in the CMake config and in all builds, was still labeled as experimental, and wasn't that much better than the regular path. This will allow a new .get_buffer callback to be implemented on top of the async protocol once times comes.

The rest of the commits clean up the iiod-client code, so that it will later be much easier to introduce the new async code.

Tested with network / usb backends and with iio_info / iio_readdev.


usb.c
```diff
@@ -246,50 +246,56 @@ static int usb_open(const struct iio_device *dev,
...
-	ret = iiod_client_open_unlocked(ctx_pdata->iiod_client, &pdata->io_ctx,
-			dev, samples_count, cyclic);
-
-	if (!ret) {
-		unsigned int remote_timeout =
-			usb_calculate_remote_timeout(ctx_pdata->timeout_ms);
+	ret = iiod_client_open_unlocked(client, &pdata->io_ctx,
+					dev, samples_count, cyclic);
+	if (ret)
+		goto err_close_pipe;
```
nunojsa: `iio_mutex_unlock(pdata->lock);` ? Also in the error handling of `iiod_client_set_timeout()`...

pcercuei: Good catch.

pcercuei:
Updated + a few related extra patches.

nunojsa (approved):
LGTM

PR #722, opened 2021-07-26
lock: Add support for condition variables and threads

pcercuei:
This set of commits add support for condition variables, and creating/joining threads, on Unix and Windows. This will be used to handle the async iiod-client protocol, which requires threads for communication.

### DIFF
--- a/lock.c
+++ b/lock.c
@@ -21,6 +21,13 @@ struct iio_mutex {
 #else
 	pthread_mutex_t lock;
 #endif
+};
+
+struct iio_cond {
+#ifdef NO_THREADS
+	int foo; /* avoid complaints about empty structure */
+#else
+	pthread_cond_t cond;
 #endif
 };
 
@@ -56,5 +63,41 @@ void iio_mutex_unlock(struct iio_mutex *lock)
 {
 #ifndef NO_THREADS
 	pthread_mutex_unlock(&lock->lock);
+#endif
+}
+
+struct iio_cond * iio_cond_create(void)
+{
+	struct iio_cond *cond = malloc(sizeof(*cond));
+
+	if (!cond)
+		return NULL;
+
+#ifndef NO_THREADS
+	pthread_cond_init(&cond->cond, NULL);
+#endif
+
+	return cond;
+}
+
+void iio_cond_destroy(struct iio_cond *cond)
+{
+#ifndef NO_THREADS
+	pthread_cond_destroy(&cond->cond);
+#endif
+	free(cond);
+}
+
+void iio_cond_wait(struct iio_cond *cond, struct iio_mutex *lock)
+{
+#ifndef NO_THREADS
+	pthread_cond_wait(&cond->cond, &lock->lock);
+#endif
+}
+
+void iio_cond_signal(struct iio_cond *cond)
+{
+#ifndef NO_THREADS
+	pthread_cond_signal(&cond->cond);
 #endif
 }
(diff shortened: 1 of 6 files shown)


lock.c
```diff
@@ -101,3 +109,55 @@ void iio_cond_signal(struct iio_cond *cond)
 	pthread_cond_signal(&cond->cond);
 #endif
 }
+
+#ifndef NO_THREADS
+static void * iio_thrd_wrapper(void *d)
+{
+	struct iio_thrd *thrd = d;
+
+	return (void *)(intptr_t) thrd->func(thrd->d);
```
nunojsa, line 118: Why not keeping pthreads signature for the callback? It looks like we are not checking the return code anyways...

lock.c
```diff
@@ -101,3 +109,55 @@ void iio_cond_signal(struct iio_cond *cond)
...
+{
+	struct iio_thrd *iio_thrd;
+	int ret;
+
+	iio_thrd = malloc(sizeof(*iio_thrd));
+	if (!iio_thrd)
+		return NULL;
+
+	iio_thrd->d = d;
+	iio_thrd->func = thrd;
```
nunojsa, line 133: I know it does not make much sense for anyone calling this with `thrd` set to `NULL`  and this would be something (most likely) caught right away but I would still make sure `thrd` in not NULL and return `ERR_PTR(-EINVAL) ` (or something like that) in case it is... Having segfaults is never neat

nunojsa (commented):
Looks good to me... Just two things from my side. I'm also wondering if we should not provide an `iio_thrd_cancel()` function that users could call before  `iio_thrd_join_and_destroy()`? Not super important for now though...

lock.c
nunojsa: Or maybe just use `pthread_exit()` and pass the return value to `pthread_join()`...

pcercuei:
> Looks good to me... Just two things from my side. I'm also wondering if we should not provide an `iio_thrd_cancel()` function that users could call before `iio_thrd_join_and_destroy()`? Not super important for now though...

`iio_thrd_cancel()` could be added the moment it's needed. Right now I don't really need it.

lock.c
pcercuei: This is needed because the prototype of the callbacks is different on Windows and Unix.

lock.c
pcercuei: Alright.

lock.c
nunojsa: Hmm I see. We could get away with some defines but I guess no need for that. Anyways, I would still consider have `pthread_exit()`... But no strong opinion, I'm also ok in leaving that decision to the real implementation of the thread (probably it's even safer)

pcercuei: I wouldn't really use `pthread_exit()`, for instance I don't use in IIOD either. Since exiting the thread function does the exact same thing...

nunojsa: I think the usage of `pthread_exit()` is mainly so you can pass a ret value to `pthread_join()`. But yeah, not that important and honestly I almost never remember to use either

pcercuei: If you "return ptr;" in the thread function, you will retrieve "ptr" in `pthread_join()`, so that's no different.

nunojsa: Hmm yes :facepalm: , it's implicit when you return from the thread function..

nunojsa (approved):
LGTM. Although It seems that the CI error for windows is really an issue? Not sure though...

pcercuei:
That's strange, it builds fine here on Visual Studio for both x86 and amd64.

pcercuei:
I worked around it with a cast to `LPTHREAD_START_ROUTINE`. I don't like it, but at least it builds now.

PR #726, opened 2021-08-03
iiod: Avoid race between buffer destroy and set buffers count

pcercuei:
The client might call iio_device_set_kernel_buffers_count() right after
destroying the device's buffer with iio_buffer_destroy(). Since the
first operation cannot work if the buffer is enabled, we must wait until
the buffer has been effectively disabled before the setting can be
applied.

### DIFF
--- a/iiod/ops.c
+++ b/iiod/ops.c
@@ -1363,6 +1363,8 @@ int set_timeout(struct parser_pdata *pdata, unsigned int timeout)
 int set_buffers_count(struct parser_pdata *pdata,
 		struct iio_device *dev, long value)
 {
+	unsigned int i, nb = (unsigned int) value;
+	struct timespec wait;
 	int ret = -EINVAL;
 
 	if (!dev) {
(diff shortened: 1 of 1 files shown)


iiod/ops.c
```diff
@@ -1363,16 +1363,28 @@ int set_timeout(struct parser_pdata *pdata, unsigned int timeout)
 int set_buffers_count(struct parser_pdata *pdata,
 		struct iio_device *dev, long value)
 {
+	unsigned int i, nb = (unsigned int) value;
```
nunojsa, line 1366: nit: no need for explicit cast...

nunojsa (approved):

iiod/ops.c
pcercuei: Cast from a bigger type to a smaller type require explicit cast, or the compiler will complain.

nunojsa: Hmm, don't think it will. The cast is implicit... Gave it a quick try and no warning issued with gcc (even with --std=c90):

```
long tmp1 = 10000000000;
int tmp0 = tmp1;

printf("%d\n", tmp0)
```

The compiler is more picky when casting pointers (which obviously makes sense)...

Now, it might be that other compilers for other platforms are more picky (I'm basing this comment on gcc + linux)

pcercuei: Try it in Visual Studio now :)

nunojsa: Ok, that goes along with my last comment :).

nunojsa: > warning C4244: conversion from 'int64_t' to 'int', possible loss of data 

FWIW, I do not think that this brings much added value but well, we need to be cross platform :).

pcercuei: Actually, I forgot that this code is in IIOD and not directly in libiio, so we only care about Linux :) But that's something I have to do everywhere in libiio to avoid warnings, so I'm just used to it now.

PR #727, opened 2021-08-03
New context / library get_version API

pcercuei:
This set of commits adds thee new functions `iio_context_get_version_major`, `iio_context_get_version_minor`, `iio_context_get_version_tag`, convert the examples / tests / bindings to this new API, and moves the old `iio_library_get_version` and `iio_context_get_version` functions to the compat library.

These two had a few major problems:
- They returned the values in their parameters; therefore, calling code
  had to define local variables, then call the function, then use these
  local variables in e.g. a printf() call. The new API is simpler, as
  the values are returned directly by each function.
- The "git_tag" parameter was defined in iio.h as an array of 8
  characters (7 ASCII characters + \0 byte). While it may not happen
  anytime soon, it is doomed to fail as soon as the number of commits in
  libiio causes the short hashes to change from being 7 chars to 8 or
  more chars.
- Arrays in function prototypes are really a bad idea in general, since
  it does not offer any guarantee that the buffer is at least the size
  it claims to be. It also misleads developers, since sizeof(foo) won't
  return the same value whether "foo" is an array listed in the
  parameters, or a local or global variable.

Note that the C# binding update was not tested.

### DIFF
--- a/context.c
+++ b/context.c
@@ -392,6 +392,30 @@ int iio_context_get_version(const struct iio_context *ctx,
 	return 0;
 }
 
+unsigned int iio_context_get_version_major(const struct iio_context *ctx)
+{
+	if (ctx && ctx->git_tag)
+		return ctx->major;
+
+	return LIBIIO_VERSION_MAJOR;
+}
+
+unsigned int iio_context_get_version_minor(const struct iio_context *ctx)
+{
+	if (ctx && ctx->git_tag)
+		return ctx->minor;
+
+	return LIBIIO_VERSION_MINOR;
+}
+
+const char * iio_context_get_version_tag(const struct iio_context *ctx)
+{
+	if (ctx && ctx->git_tag)
+		return ctx->git_tag;
+
+	return LIBIIO_VERSION_GIT;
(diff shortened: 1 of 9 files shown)


context.c
```diff
@@ -392,6 +392,30 @@ int iio_context_get_version(const struct iio_context *ctx,
 	return 0;
 }
 
+unsigned int iio_context_get_version_major(const struct iio_context *ctx)
+{
+	if (ctx && ctx->git_tag)
+		return ctx->major;
```
nunojsa, line 378: just curious: Do we only get major/minor if there's a context git_tag? Looks a bit odd to me (just by looking at this code and variable names).

nunojsa (approved):

context.c
pcercuei: Yes, basically if `ctx->git_tag` is initialized, then it was created with the XML backend, and it is safe to assume that ctx->major and ctx->minor have been initialized as well. Otherwise you're either running with the local backend or a very old version of IIOD, and in this case you return the library version.

Issue #648, opened 2021-02-04
bindings: python: TypeError: 'Context' object is not callable


ekigwana:
Just tried running the pluto example from [pyadi-iio](https://github.com/analogdevicesinc/pyadi-iio) and got the following error: ` TypeError: 'Context' object is not callable`

https://github.com/analogdevicesinc/libiio/blob/359305cd482cbcd6a55386adc9a54b68364e7dad/bindings/python/iio.py#L1016

I replaced it with `self._ctx = device.context()` which appears to be what it is intended for and  moved the error to:

https://github.com/analogdevicesinc/libiio/blob/359305cd482cbcd6a55386adc9a54b68364e7dad/bindings/python/iio.py#L1331

Looking at the Context the is not definition to return a reference.

Not sure if the fix to replace all `.ctx()` with `.ctx` or in just some places or to define something in the Context

tfcollins:
Can you back out this change https://github.com/analogdevicesinc/libiio/commit/fe86c6b7c315674e3f7cad0cbac152504757bebe

-Travis

mrhgit:
This fix seems incomplete for two reasons:
 1) there are still calls to self.ctx() in the Device class, instead of self.ctx
 2) weakref does not protect the object from deletion, which is exactly what the Buffer class needs to do, because it uses the context object in the deletion routine (see the C code for buffer_destroy).

pcercuei:
@ekigwana @mrhgit Could you test my PR #735? Thanks!

pcercuei:
This should be fixed now thanks to #735. Closing the issue.

ekigwana:
Late, but works for me now

PR #735, opened 2021-08-17
python: Fix circular dependencies between objects

pcercuei:
Buffer creation was broken since PR #636.

### DIFF
--- a/bindings/python/iio.py
+++ b/bindings/python/iio.py
@@ -339,6 +339,10 @@ _get_device.restype = _DevicePtr
 _get_device.argtypes = (_ContextPtr, c_uint)
 _get_device.errcheck = _check_null
 
+_find_device = _lib.iio_context_find_device
+_find_device.restype = _DevicePtr
+_find_device.argtypes = (_ContextPtr, c_char_p)
+
 _set_timeout = _lib.iio_context_set_timeout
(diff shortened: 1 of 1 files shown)


pcercuei:
I just noticed it was in a bug report (#648), and that my fix is incomplete, so it shouldn't be merged.

Also, I'm pretty sure PR #636 introduced a circular reference between iio.Context and iio.Device objects, which would mean that none of these objects are deleted.

pcercuei:
I think I fixed it properly now.

adisuciu:
Using this script with the Pluto
```
import iio

ctx=iio.Context("ip:192.168.2.1")
dev=ctx.find_device("cf-ad9361-lpc")
dev.channels[0].enabled=True
dev.channels[1].enabled=True
buf=iio.Buffer(dev,1024)
buf.refill()
print(buf.read())

```

master gives out error:
```
  File "iiopybindings.py", line 7, in <module>
    buf=iio.Buffer(dev,1024)
  File "/usr/local/lib/python3.8/dist-packages/iio.py", line 1013, in __init__
    self._ctx = device.ctx()
TypeError: 'Context' object is not callable

```

This PR returns actual data so it fixes the problem.

adisuciu (approved):

PR #750, opened 2021-10-05
C#: Add support for data format

pcercuei:
Provide a "format" field in the Channel object, which contains a
read-only DataFormat structure similar to iio_data_format.


mhennerich (approved):
Looks good to me

pcercuei:
I'll ignore the Codacy issues here because I don't really think they make much sense.

PR #755, opened 2021-10-24
Move the "tests" directory to the properly named "utilities" directory

rgetz:
This makes the command line utilties that we have been growing in the
"tests" directory, into a more properly named directory called
"utilities", which matches what distributions have been calling them.
(and makes sense). This will free up the ./tests directory for actual
self tests.


tfcollins:
Having things in python would make writing and running tests a bit easier. Can leverage pytest-libiio+emulation as well
https://pytest-libiio.readthedocs.io/en/latest
https://pytest-libiio.readthedocs.io/en/latest/emulation/

rgetz:
per discussion - drop this for now.

PR #763, opened 2021-11-15
iio.h: Include iio/types.h when it exists

### LINKED ISSUE #758: Redefinition errors between libiio and kernel iio include files
Writing an application that including libiio iio.h for handling iio buffers and kernel iio include files for handling events, there are redefinition errors between `/usr/include/linux/iio/types.h` and `/usr/include/iio.h`:

```
/usr/include/linux/iio/types.h:14:6: error: redefinition of 'iio_chan_type'
enum iio_chan_type {                                                          
     ^                                                                        
/usr/include/iio.h:94:6: note: previous definition is here
enum iio_chan_type {            
     ^              
```

When compiling on linux architecture, would it make sense for ./iio.h to include `<linux/iio/types.h>`?

gwendalcr:
enum iio_chan_type and iio_modifier are already defined in
/usr/include/linux/iio/types.h on linux systems.
If an application needs to include both (to support iio events for
instance), compiler will complains these enums are defined twice.

This patch is not without problem:
- it works only if compiler supports __has_include (C++17, not 11).
- linux system must have latest kernel-headers installed:
IIO_MOD_O2 modifier has been added to linux 5.9 kernel.

Alter CI/travis/check_kernel.sh accordingly, check it still passes.

Fixes #758.

### DIFF
--- a/iio.h
+++ b/iio.h
@@ -84,6 +84,16 @@ struct iio_context_info;
 struct iio_scan_context;
 struct iio_scan_block;
 
+#define	IIO_CHAN_TYPE_UNKNOWN INT_MAX
+
+#if defined __has_include
+#  if __has_include(<linux/iio/types.h>)
+#    include <linux/iio/types.h>
+#    define __IIO_TYPE_DEFINED__
+#  endif
+#endif
+
+#ifndef __IIO_TYPE_DEFINED__
 /**
  * @enum iio_chan_type
  * @brief IIO channel type
@@ -127,7 +137,6 @@ enum iio_chan_type {
 	IIO_POSITIONRELATIVE,
 	IIO_PHASE,
 	IIO_MASSCONCENTRATION,
-	IIO_CHAN_TYPE_UNKNOWN = INT_MAX
 };
 
 /**
@@ -184,6 +193,8 @@ enum iio_modifier {
 	IIO_MOD_H2,
 	IIO_MOD_O2,
 };
+
+#endif  /* !__IIO_TYPE_DEFINED__ */
 
 /* ---------------------------------------------------------------------------*/
 /* ------------------------- Scan functions ----------------------------------*/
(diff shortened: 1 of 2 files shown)


iio.h
```diff
@@ -84,6 +84,12 @@ struct iio_context_info;
 struct iio_scan_context;
 struct iio_scan_block;
 
+#define	IIO_CHAN_TYPE_UNKNOWN INT_MAX
+
+#if __has_include(<linux/iio/types.h>)
```
nunojsa, line 89: Hmm, according to [this](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html), the right way of doing this is to first test for the existence of the macro... I'm not sure this will actually build when the compiler does not support `__has_include`

nunojsa (commented):
Just one minor comment from side... Alternatively to this we could just think about syncing the current iio uapi within libiio on every stable release . Some other projects (eg: libiouring) apparently do the same... However I'm not sure how practical that is given some potential customs things we have. @pcercuei might have a better idea for this issue

rgetz:
don't we just want to do a:
```
# ensure that kernel headers are included before libiio headers
#ifndef iio_chan_type

enum ....

#endif
```

this would resolve things, but only if you included kernel headers before iio.h

pcercuei:
@rgetz it's not a macro, though.

@gwendalcr please wrap the `__has_include` macro as described in here: https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html

gwendalcr:
I really don't like this patch, it is an ugly band aid:

1. wrapping __has_include will help pass some tests, but an application including both kernel include and iio.h will not be portable anymore between C++ 11 and 17.
2. sync'ing uapi does not work for [embedded] devices running an older kernel. We are not able to use the latest libiio when cros compiling for an older kernel.

In another project, to fill `iio_chan_type_name_spec[]` and `modifier_names[]` type arrays, [X-Macros](
https://en.wikibooks.org/wiki/C_Programming/Preprocessor_directives_and_macros#X-Macros) have been used, but that would require using the macros in the kernel includes as well.

pcercuei:
What about we sync to the latest uapi header now, then protect libiio's copy with the kernel header's `_IIO_TYPES_H_`? That sounds like the simplest and safest solution.

pcercuei:
@gwendalcr could you try 5fc75507?

gwendalcr:
5fc7550 is a better patch.

PR #773, opened 2021-12-13
iio_{readdev,writedev}: Add benchmark feature

pcercuei:
Add a new --benchmark (or -B) option.

With this option enabled, no data is written to the standard output (for
iio_readdev) or read from the standard input (for iio_writedev).
Instead, the buffers are flipped as fast as possible. This functionality
permits to benchmark the speed at which the data flows, and compare the
performance of the various backends and/or with various buffer sizes.

### DIFF
--- a/tests/iio_common.c
+++ b/tests/iio_common.c
@@ -25,6 +25,7 @@
 #include <inttypes.h>
 #include <getopt.h>
 #include <string.h>
+#include <time.h>
 
 #include "iio_common.h"
 #include "gen_code.h"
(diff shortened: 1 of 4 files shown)


pcercuei:
Note that we're only benchmarking the throughput of the medium, which is different than using iio_readdev/iio_writedev with "pv". In this latter case, the measured throughput is halved as time is spent reading/writing the IIO buffers, and reading/writing the standard input/output (which means CPU copies), and "pv" itself does have some overhead.

rgetz (approved):

PR #774, opened 2021-12-16
iiod/init/iiod.service.cmakein: Add support for extra iiod options

mhennerich:
This patch adds support to source iiod options via /etc/default/iiod

### DIFF
--- a/iiod/init/iiod.service.cmakein
+++ b/iiod/init/iiod.service.cmakein
@@ -1,6 +1,8 @@
+# SPDX-License-Identifier: LGPL-2.1+
+#
 # iiod - Systemd init script
 #
-# Copyright (C) 2016 Analog Devices Inc.
+# Copyright (C) 2016-2021 Analog Devices Inc.
 
 [Unit]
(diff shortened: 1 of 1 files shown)


iiod/init/iiod.service.cmakein
```diff
@@ -1,14 +1,17 @@
+# SPDX-License-Identifier: LGPL-2.1+
+#
 # iiod - Systemd init script
 #
-# Copyright (C) 2016 Analog Devices Inc.
+# Copyright (C) 2016-2021 Analog Devices Inc.
```
nunojsa, line 5: The above two options were sneaked in this patch...

nunojsa (approved):
Just a minor comment from side. I would do the SPDX identifier and the Coypright in a different patch...

mhennerich:
Pushed V2:
Split into two patches

iiod/init/iiod.service.cmakein
mhennerich: ACK

PR #775, opened 2021-12-27
added instructions for VS 2019

elektronika-ba:
Steps required to compile this program in Visual Studio 2019


rgetz:
>Steps required to compile this program in Visual Studio 2019

Using the Visual Analog IIO Plugin has nothing to do with using Visual Studio in the generic sense, so I don't think the instructions are correct at all...

If you are having issues with the visual analog plugin - that's something that needs to be addressed on ADI's EngineerZone, not inside libiio.

rgetz:
closing since no action in 30 days.

PR #778, opened 2022-01-05
Add iio_device_is_hwmon() and iio.Device.hwmon property in bindings

pcercuei:
These read-only properties will be True if the iio.Device is a hardware monitoring device, and False if it is a IIO device.

### DIFF
--- a/bindings/csharp/Device.cs
+++ b/bindings/csharp/Device.cs
@@ -210,6 +210,9 @@ namespace iio
         /// <summary>The label of this device.</summary>
         public string label { get; private set; }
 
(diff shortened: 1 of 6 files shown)


pcercuei:
v2: I added a `iio_device_is_hwmon` public function, since I believe it will be useful to C programs as well.

rgetz:
would it be simpler to always make `is_hwmon()` return `false` when `WITH_HWMON` is off? Then you would not need it in every place... (wouldn't you?)

pcercuei:
We can't access `iio-config.h` from `iio.h`, so we can't use WITH_HWMON in there.

rgetz (approved):

PR #781, opened 2022-01-17
local: Add support for setting watermark

### LINKED ISSUE #780: Unable to set buffer watermark correctly
Hi,

I'm reading from a SPI ADC which use software buffer to offload frequent Kernel-user space copy overhead, and I need to set `watermark` attribute of the buffer to enable this.

However, `watermark` needed to be set after `length` and before `enable`.
Here is the kernel soruce: https://github.com/linux4sam/linux-at91/blob/ea5298f615584931c389b14c81af4c1b5da6fbba/drivers/iio/industrialio-buffer.c#L1182

If I use `iio_device_buffer_attr_write(mcp3913, "watermark", "20");` before `iio_device_create_buffer(mcp3913, BUFFER_SIZE, false);` I got `invalid argument` since buffer length is still 1.

If I use `iio_device_buffer_attr_write(mcp3913, "watermark", "20");` after`iio_device_create_buffer(mcp3913, BUFFER_SIZE, false);` I got `ressource busy` since buffer is already enabled.

The workaround is `iio_device_create_buffer` -> `iio_buffer_destroy` -> `iio_device_buffer_attr_write` -> `iio_device_create_buffer` but it's ugly.

pcercuei:
Add support for setting the watermark value, if the driver supports it.
The watermark value represents the number of samples that will be read
in one burst from the hardware. Since we read data at the granularity of
iio_buffer blocks, it makes sense to use the highest watermark value up
to the buffer's size. If the value is too high, the driver will
automatically adjust it to the maximum watermark value possible.

Fixes #780.

### DIFF
--- a/local.c
+++ b/local.c
@@ -104,6 +104,7 @@ static const char * const device_attrs_blacklist[] = {
 static const char * const buffer_attrs_reserved[] = {
 	"length",
 	"enable",
+	"watermark",
 };
 
 static int ioctl_nointr(int fd, unsigned long request, void *data)
@@ -922,6 +923,15 @@ static int local_open(const struct iio_device *dev,
 	ret = local_write_dev_attr(dev, "buffer/length",
 			buf, strlen(buf) + 1, false);
 	if (ret < 0)
+		return ret;
+
+	/*
+	 * Set watermark to the buffer size; the driver will adjust to its
+	 * maximum if it's too high without issueing an error.
+	 */
+	ret = local_write_dev_attr(dev, "buffer/watermark",
+				   buf, strlen(buf) + 1, false);
+	if (ret < 0 && ret != -ENOENT)
 		return ret;
 
 	pdata->cancel_fd = eventfd(0, EFD_CLOEXEC | EFD_NONBLOCK);


HiFiPhile:
Hi,

Thanks for your PR.

I didn't see in kernel's `iio_enable_buffers()` watermark setting error is ignored, so even if the driver returns error it will still works.
```
	if (indio_dev->info->hwfifo_set_watermark)
		indio_dev->info->hwfifo_set_watermark(indio_dev,
			config->watermark);
```

------

Because the actual value of `watermark` and `length` will decide if otherone's value can be successfully written:
- `length` must >= `watermark`
- `watermark` must <= `legnth`
Otherwise a `Invalid argument` will happen when you write the value (you can test simplyly with `echo`)

It's not sufficient to only write watermark once.

For example:
1. length=100, watermark=100 from previous run
2. New run want length=50
3. `local_write_dev_attr(dev, "buffer/length", 50)` will fail will `Invalid argument` since it's smaller than watermark.

I think you need to reset `watermark to 1 before setting length.

pcercuei:
This is incorrect. If the `watermark` is higher than the new buffer length, the kernel will reduce it to the buffer length. See: https://github.com/torvalds/linux/blob/master/drivers/iio/industrialio-buffer.c#L667-L668

HiFiPhile:
> This is incorrect. If the watermark is higher than the new buffer length, the kernel will reduce it to the buffer length. 

Hum... Maybe I've some memory corruption... will try it later.

HiFiPhile:
I've tested creating buffer with size up & down, it's working :)

pcercuei:
@HiFiPhile And it also works if the buffer size isn't a multiple of the watermark?

HiFiPhile:
With a size of 110 and watermark of 20 it seems working. I don't have a high speed device to test if last conversion will be cut off, in low speed device since kernel buffer has a 4*length it should be ok.

pcercuei:
But 110*4 is divisible by 20 :)
Could you try e.g. a buffer size of 22 samples? I just want to make sure it doesn't lock up.

HiFiPhile:
With a length of 123 and watermark of 20 it still works :)

pcercuei:
Great, thanks. I sent patches upstream to fix the two drivers that errored with too high watermarks.

amiclaus (approved):

PR #787, opened 2022-02-09
iio-backend: Remove unused callback .get_description

pcercuei:
This callback was introduced in commit d61f6296 ("iio: backend: add
get_description() backend op") but was never called anywhere, as the
backend's description string is also present in the XML directly.


pcercuei:
Rebased on top of master, which should fix the CI errors. The commit itself was not modified.

rgetz (approved):

PR #790, opened 2022-02-16
iio_info: Remove -s option

pcercuei:
iio_info had two possible options to scan for contexts: -S (aka. --scan)
and -s, which did more or less the same thing, but -s didn't take an
argument and scanned all possible contexts.

Remove the -s option in the help text to reduce the confusion, while
still handling it for backwards-compatibility.

### DIFF
--- a/man/iio_info.1.in
+++ b/man/iio_info.1.in
@@ -78,7 +78,7 @@ normally returned from
 with no address part
 .RE
 .TP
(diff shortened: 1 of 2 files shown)


rgetz (approved):

rgetz:
Will need to go through the wiki, and dox and man looking for -s as well...

PR #791, opened 2022-02-17
Support specifying VID/PID for scanning + steamline scanning in tools

pcercuei:
A rewrite of #764.

This PR does a few things.
* Change the separator in the backends scan string from a colon to a comma (colon still supported for backwards-compatibility)
* Rework scanning to support having backends listed multiple times in the list
* Add support for specifying backend params, currently only for the USB backend: e.g. `usb=0456:b672`
* Update the libiio udev rule to only scan for the detected USB device
* Unify tools on the `--scan` (aka `-S`) option and updated the documentation and man pages
* Print usage examples on iio_readdev / iio_writedev.

The possibility to specify a USB device's VID/PID on the backends scan string makes it possible to only target a specific device, instead of probing all devices on the USB bus. This is useful for applications built around particular pieces of hardware. It also avoids disrupting the functionality of other USB devices on the bus.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,7 @@ include(CheckSymbolExists)
 check_symbol_exists(strdup "string.h" HAS_STRDUP)
 check_symbol_exists(strndup "string.h" HAS_STRNDUP)
 check_symbol_exists(strerror_r "string.h" HAS_STRERROR_R)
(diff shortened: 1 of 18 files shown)


rgetz (approved):

pcercuei:
@rgetz this last commit breaks iio_readdev / iio_writedev.

PR #801, opened 2022-02-23
Windows DNS SD : add port knocking like other implementations

### LINKED ISSUE #559: scan failure on one backend causes failures on all
@mhennerich noticed that an ethernet scan failure causes a total scan failure...

I think this is a problem with bad actors on ethernet with a Windows clients.

But will need to check a little bit more...

rgetz:
Both the Linux and Mac implemations of DNS SD ensures something is there
before passing the host information up the stack. The Windows
implemnation was missing this, so add it.

Fix #559

### DIFF
--- a/dns_sd_windows.c
+++ b/dns_sd_windows.c
@@ -16,6 +16,7 @@
 
 #include "debug.h"
 #include "dns_sd.h"
+#include "iio-lock.h"
 #include "iio-private.h"
 #include "mdns.h"
 
@@ -189,6 +190,7 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 	WSADATA wsaData;
 	const char service[] = "_iio._tcp.local";
 	size_t records, capacity = 2048;
+	struct dns_sd_discovery_data *d;
 	unsigned int i, isock, num_sockets;
 	void *buffer;
 	int sockets[32];
@@ -202,14 +204,19 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 
 	IIO_DEBUG("DNS SD: Start service discovery.\n");
 
-	*ddata = zalloc(sizeof(**ddata));
-	if (!*ddata)
+	d = zalloc(sizeof(*d));
+	if (!d)
 		goto out_wsa_cleanup;
+	/* pass the structure back, so it can be freed if err */
+	*ddata = d;
 
+	d->lock = iio_mutex_create();
+	if (!d->lock)
+		goto out_wsa_cleanup;
 
 	buffer = malloc(capacity);
 	if (!buffer)
-		goto out_wsa_cleanup;
+		goto out_destroy_lock;
 
 	IIO_DEBUG("Sending DNS-SD discovery\n");
 
@@ -250,7 +257,7 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 
 				records += mdns_query_recv(sockets[isock],
 							   buffer, capacity,
-							   query_callback, *ddata,
+							   query_callback, d,
 							   transaction_id[isock]);
 			}
 		} while (records);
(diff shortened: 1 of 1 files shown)


rgetz:
create a draft, so I can test the Windows builds.

dns_sd_windows.c
```diff
@@ -202,14 +204,17 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
...
 
-	*ddata = zalloc(sizeof(**ddata));
-	if (!*ddata)
+	d = zalloc(sizeof(*d));
+	if (!d)
 		goto out_wsa_cleanup;
 
+	d->lock = iio_mutex_create();
+	if (!d->lock)
+		goto out_destroy_lock;
```
pcercuei, line 213: This goto looks fishy.

pcercuei (changes requested):

dns_sd_windows.c
rgetz: yes - that was wrong - should be fixed now. Thanks

pcercuei (approved):

dns_sd_windows.c
```diff
@@ -202,14 +204,17 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 
 	IIO_DEBUG("DNS SD: Start service discovery.\n");
 
-	*ddata = zalloc(sizeof(**ddata));
-	if (!*ddata)
+	d = zalloc(sizeof(*d));
+	if (!d)
 		goto out_wsa_cleanup;
```
pcercuei, line 209: Since you added a `d` variable instead of using `*ddata` you are now leaking memory, because in all error cases the dns_sd_discovery_data won't be freed by the parent function.

pcercuei (changes requested):

dns_sd_windows.c
rgetz: d is passed back as a pointer, like all other implementations (avahi, and bonjour), so when the parent frees the variable it's fine.

rgetz: Memory is allocated inside each implementation (`dns_sd_avahi.c` `dns_sd_bonjour.c` `dns_sd_windows.c`) in the `dnssd_find_hosts()` function. It's cleared when the top level calls `dnssd_free_all_discovery_data()`

You are right - I was missing freeing it in the error paths. That is fixed now.

pcercuei: Not entirely fixed, the `d` is not freed if iio_mutex_create() fails.
Also - why not simply set `*ddata = d;` right after `d` is allocated? Then you wouldn't have to call `dnssd_free_all_discovery_data` everywhere.

rgetz: Yeah, you have sharper eyes/better ideas that I do...

done.

rgetz:
with these changes - just downloading and testing on windows now...

rgetz:
just tested the windows VS-2019 build, and it seems to work correctly, and not manifest the issues reported earlier.

Thanks

rgetz:
> This should probably be called only if `ret == 0`. Because if `ret < 0` that would cause use-after-free errors, when the caller function calls `dnssd_free_all_discovery_data(ddata)`.

I think (but am happy to be corrected), as long as `d` or `ddata` points to zero'd memory (which it does via `zalloc`) it should be fine to call free (null pointer). Correct?

pcercuei:
> I think (but am happy to be corrected), as long as `d` or `ddata` points to zero'd memory (which it does via `zalloc`) it should be fine to call free (null pointer). Correct?

It is fine to call free(NULL), but this is not what would happen here. After `dnssd_free_all_discovery_data(ddata)` is called, the `ddata` pointer is invalid and not NULL, so a second call to dnssd_free_all_discovery_data() would cause use-after-free.

rgetz:
> `dnssd_free_all_discovery_data(ddata)` is called

yeah - sorry - I missed that I had not removed all the `dnssd_free_all_discovery_data(ddata)` calls. I was commenting on what I thought the code was - not what I had actually done... Sorry.

-Robin

rgetz:
> if `open_client_sockets()` returns 0, the `ddata` is never freed.

`dns_sd.c:dnssd_context_scan()` calls `dnssd_find_hosts()`

If `open_client_sockets()` returns 0, it drops down to `out_free_buffer`; which will return with a `0` code.

`dns_sd.c:dnssd_context_scan()` will try to iterate on things:
`for (ndata = ddata; ndata->next != NULL; ndata = ndata->next)`

and since it is all null, (via zalloc), it should drop down to `dnssd_free_all_discovery_data(ddata);`

shouldn't it?

pcercuei:
I guess you are right, yes.

Issue #537, opened 2020-06-03
re-work libiio debug message logging


rgetz:
## Error handling
Typically libiio functions return 0 on success or a negative error code on failure. These negative error codes relate to system error code, which can be multiplied by `-1` and passed directly to `strerror()`.

## Debug message logging

Currently - libiio squirts out random messages on `stdout` or `stderr`.  This is problematic for many applications.

The goal is to move to the model that [libusb](https://www.cs.unm.edu/~hjelmn/libusb_hotplug_api/group__lib.html#ga5f8376b7a863a5a8d5b8824feb8a427a) uses a newly imagined `iio_context_set_debug()` which would be used to enable stdout/stderr logging of certain messages, on a context by context basis.

The goal would be to use the existing levels:
- `NoLog` : no messages ever printed by the library (default). Applications are therefore free to close stdout/stderr and those descriptors may be reused without worry.
- `Error` : error messages are printed to stderr
- `Warning` :  warning and error messages are printed to stderr
- `Info` :  Informational messages are printed to stdout, warning and error messages are printed to stderr
- `Debug` :  Debug and Informational messages are printed to stdout, warning and error messages are printed to stderr

The logged messages are unstructured. There is no one-to-one correspondence between messages being logged and success or failure return codes from libiio functions. There is no format to the messages, and no guarantee about forwards/backwards compatibility, so you should not try to capture or parse them. They are not and will not be localized. These messages are not suitable for being passed to your application user; instead, the application developer should interpret the error codes returned from libiio functions and provide appropriate notification to the user. The messages are simply there to aid you as a programmer/application developer, and if you're confused because you're getting a strange error code from a libiio function, enabling message logging may give you a suitable explanation.

It is planned that libiio can be configured at cmake time to build without any logging functions at all, useful for embedded systems. In this case, `iio_context_set_debug()` will have no effects.

rgetz:
will need to handle some things in iiod differently - and most likely take -v -vv and -vvvv on command line.

rgetz:
Need to handle network and usb errors differently - right now, we always translate things into something that can be passed to strerr, or perror - but this is problematic, since networking errors (in `/usr/include/netdb.h`) 

```
/* Error values for `getaddrinfo' function.  */
# define EAI_BADFLAGS     -1    /* Invalid value for `ai_flags' field.  */
# define EAI_NONAME       -2    /* NAME or SERVICE is unknown.  */
[...]

PR #808, opened 2022-02-28
Improve error handling in public API

pcercuei:
This set of commits introduce new public functions `iio_err`, `iio_ptr` and `iio_err_cast`, and changes the way errors are handled internally and in the API.

Instead of returning NULL and setting the `errno` variable, the policy is now to return an error encoded in a pointer. The error code can then be retrieved with `iio_err` (zero means no error), and a error-encoding pointer can be created from an error code with `iio_ptr`.

This is the first step needed to decouple libiio from using <errno.h> error codes, which are very limiting and don't allow to represent the spectrum of error cases that can happen in libiio and IIOD.

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -26,24 +26,19 @@ static bool device_is_high_speed(const struct iio_device *dev)
 struct iio_buffer * iio_device_create_buffer(const struct iio_device *dev,
 		size_t samples_count, bool cyclic)
 {
-	ssize_t ret = -EINVAL;
 	struct iio_buffer *buf;
-	ssize_t sample_size = iio_device_get_sample_size(dev, NULL);
+	ssize_t ret, sample_size = iio_device_get_sample_size(dev, NULL);
 	size_t mask_size;
 
 	if (!sample_size || !samples_count)
-		goto err_set_errno;
+		return iio_ptr(-EINVAL);
 
-	if (sample_size < 0) {
-		ret = sample_size;
-		goto err_set_errno;
-	}
+	if (sample_size < 0)
+		return iio_ptr((int) sample_size);
 
 	buf = malloc(sizeof(*buf));
-	if (!buf) {
-		ret = -ENOMEM;
-		goto err_set_errno;
-	}
+	if (!buf)
+		return iio_ptr(-ENOMEM);
 
 	buf->dev_sample_size = (unsigned int) sample_size;
 	buf->length = sample_size * samples_count;
@@ -100,9 +95,7 @@ err_free_mask:
 	free(buf->mask);
 err_free_buf:
 	free(buf);
-err_set_errno:
-	errno = -(int)ret;
-	return NULL;
+	return iio_ptr((int) ret);
 }
 
 void iio_buffer_destroy(struct iio_buffer *buffer)
--- a/context.c
+++ b/context.c
@@ -184,17 +184,17 @@ static char * iio_context_create_xml(const struct iio_context *ctx)
 
 	len = iio_snprintf_context_xml(NULL, 0, ctx);
 	if (len < 0)
-		return ERR_PTR((int) len);
+		return iio_ptr((int) len);
 
 	len++; /* room for terminating NULL */
 	str = malloc(len);
 	if (!str)
-		return ERR_PTR(-ENOMEM);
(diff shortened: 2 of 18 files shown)


rgetz:
Paul: just conceptually - does this give is what we want for more robust error codes/more robust error/debug strings that can be controlled in realtime? (vs compile time).

  * [libusb_error_name()](https://libusb.sourceforge.io/api-1.0/group__libusb__misc.html#gacf319b823a1a1a25cc934d80db2f28c7)
  * [libusb_strerror()](https://libusb.sourceforge.io/api-1.0/group__libusb__misc.html#gadee87eb940a5417429fe105de9e3b273)
  * [Debug message logging](https://libusb.sourceforge.io/api-1.0/#msglog)

I think this was discussed in #537 (which was never started/finished)

pcercuei:
@rgetz I don't know what you mean by "error/debug strings controlled in realtime vs. compile-time".

But this patchset is meant to decouple libiio from using `errno`, and is the first step towards an overhaul of error handling. The plan is to later have libiio-specific error codes that would describe with great precision the actual errors.

rgetz:
Paul : Thanks.

yeah, removing errno  is the first step. I just wanted to better understand where the final destination looked like in your mind.

I was trying to figure out how the pointer encoded errors ended up vs having our own libiio specific (int) based error codes.

-Robin

pcercuei:
> sorry - I meant not having to recompile the library to get debug/info messages out. Just set a loglevel or something like that...

That's already implemented and working in the dev branch.

rgetz:
Sorry - just to re-cap.

Today, run-time errors are represented by integer values. These values sometime propagate across long function- call chains before being handled, and sometimes "handled" is just passed back up to the user application. As these error codes propagate, they are often temporarily or permanently encoded into pointer values. Since error-valued pointers are not valid memory addresses, and therefore require special care by programmers, accidental misuse of pointer variables that store error codes can lead to serious problems such as system crashes, data corruption, unexpected results, etc.

Are we expecting applications calling into the library to understand error-valued pointers before de-referencing?

How do we ensure that bad use doesn't occur inside libiio? I know there are some [static tools](https://pages.cs.wisc.edu/~liblit/tr-1687/tr-1687.pdf) for the kernel - but I don't think they exist for generic code. Do they? 

Thanks
-Robin

pcercuei:
> Sorry - just to re-cap.
> 
> Today, run-time errors are represented by integer values. These values sometime propagate across long function- call chains before being handled, and sometimes "handled" is just passed back up to the user application. As these error codes propagate, they are often temporarily or permanently encoded into pointer values. Since error-valued pointers are not valid memory addresses, and therefore require special care by programmers, accidental misuse of pointer variables that store error codes can lead to serious problems such as system crashes, data corruption, unexpected results, etc.

Yes, but the same can be said with NULL pointers - the applications using libiio are still required to do some sort of error handling.
 
> Are we expecting applications calling into the library to understand error-valued pointers before de-referencing?

Yes; that's why the iio_err / iio_ptr are part of the API now.

> How do we ensure that bad use doesn't occur inside libiio? I know there are some [static tools](https://pages.cs.wisc.edu/~liblit/tr-1687/tr-1687.pdf) for the kernel - but I don't think they exist for generic code. Do they?

I don't think generic tools exist for that, no. That's the drawback, the tools won't warn about possible use of invalid pointers.

nunojsa (approved):
LGTM

gvanem:
Still improper network errors on Windows:
```
iio_info.exe -u ip:10.0.0.1
Unable to create IIO context ip:10.0.0.1: Unknown error (10061)
```

Where it ought to be `Unable to create IIO context ip:10.0.0.1: Connection refused (10061)`.

pcercuei:
@gvanem Can you trace where that error is issued?

gvanem:
By `iio_strerror()` and `strerror_s()` I assume. Called from `tests/iio_common.c`:
```c
  iio_strerror(errno, err_str, sizeof(err_str));  // !! errno is not for Winsock2 error-codes
  if (arg)
    fprintf(stderr, "Unable to create IIO context %s: %s\n", arg, err_str);
```

`iio_strerror()` could perhaps test for `err < WSABASEERR` before calling `strerror_s()`.

PR #809, opened 2022-02-28
read|write_dev : fix bug & qualify channel names and err if bad name

rgetz:
Two small commits:

1. Fix a bug that Paul found that I introduced. in iio_read|write_dev - which was an oversight in a previous commit. allow multiple channels to be added on the command line.
2. In the past, you could put gibberish channel names, and read|write_dev would accept them (ignoring them). This sometimes lead to typos causing issues with "wrong" or "incorrect" data. Fix this by checking for a match, and errorring if there is something provided that doesn't match.
    
### DIFF
--- a/tests/iio_common.c
+++ b/tests/iio_common.c
@@ -126,6 +126,21 @@ err_free_ctx:
 	return ctx;
 }
 
+int iio_device_enable_channel(const struct iio_device *dev, const char * channel, bool type)
+{
+	struct iio_channel *ch;
+
+	ch = iio_device_find_channel(dev, channel, type);
+	if (!ch)
+		return -ENXIO;
+
+	if (iio_channel_is_enabled(ch))
+		return -EBUSY;
+
+	iio_channel_enable(ch);
+	return 0;
+}
+
 unsigned long int sanitize_clamp(const char *name, const char *argv,
 	uint64_t min, uint64_t max)
 {
--- a/tests/iio_readdev.c
+++ b/tests/iio_readdev.c
@@ -373,17 +373,16 @@ int main(int argc, char **argv)
 			}
 		}
 	} else {
-		for (i = 0; i < nb_channels; i++) {
-			struct iio_channel *ch = iio_device_get_channel(dev, i);
-			for (j = optind + 1; j < (unsigned int) argc; j++) {
-				const char *n = iio_channel_get_name(ch);
-				if ((!strcmp(argw[j], iio_channel_get_id(ch)) ||
-						(n && !strcmp(n, argw[j]))) &&
-						!iio_channel_is_output(ch)) {
-					iio_channel_enable(ch);
-					nb_active_channels++;
-				}
+		for (j = optind + 1; j < (unsigned int) argc; j++) {
+			ret = iio_device_enable_channel(dev, argw[j], false);
+			if (ret < 0) {
+				char buf[256];
+				iio_strerror(-(int) ret, buf, sizeof(buf));
+				fprintf(stderr, "Bad channel name \"%s\" : %s\n", argw[j], buf);
+				iio_context_destroy(ctx);
+				return EXIT_FAILURE;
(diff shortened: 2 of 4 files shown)


pcercuei (changes requested):

tests/iio_readdev.c
```diff
@@ -373,16 +378,40 @@ int main(int argc, char **argv)
...
+						iio_context_destroy(ctx);
+						return EXIT_FAILURE;
+					} else {
+						iio_channel_enable(ch);
+						nb_active_channels++;
+						ret = 0;
+						break;
+					}
+				}
+			}
```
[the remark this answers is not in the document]
rgetz: >I think you can do much better here, 

Thanks for feedback, appreciate it. I always tell people "aim higher". :)

updated with your ideas (mostly).

tests/iio_common.c
```diff
@@ -126,6 +126,36 @@ struct iio_context * autodetect_context(bool rtn, const char * name, const char
 	return ctx;
 }
 
+/* How many input or output channels are in the device */
```
pcercuei, line 129: That doesn't really seem to describe what this function does.

rgetz: yeah, sorry about that. dropped.

tests/iio_readdev.c
```diff
@@ -373,16 +378,40 @@ int main(int argc, char **argv)
...
+					struct iio_channel *ch = iio_device_get_channel(dev, i);
+					if (iio_channel_is_scan_element(ch) &&
+							!iio_channel_is_output(ch)) {
+						const char *n = iio_channel_get_name(ch);
+						const char *id = iio_channel_get_id(ch);
+						if (n)
+							fprintf(stderr, " %s", n);
+						if (id)
+							fprintf(stderr, " %s", id);
+					}
```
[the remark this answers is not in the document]
rgetz: removed

tests/iio_readdev.c
```diff
@@ -322,6 +322,11 @@ int main(int argc, char **argv)
 		iio_context_destroy(ctx);
 		return EXIT_FAILURE;
 	}
+	if (argc > optind + (int)iio_device_get_scan_channels_count(dev, false) + 1) {
+		fprintf(stderr, MY_NAME " : Too many channels specified\n");
+		iio_context_destroy(ctx);
+		return EXIT_FAILURE;
+	}
```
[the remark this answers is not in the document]
rgetz: removed

pcercuei (approved):

tests/iio_common.c
```diff
@@ -126,6 +126,23 @@ struct iio_context * autodetect_context(bool rtn, const char * name, const char
 	return ctx;
 }
 
+ssize_t iio_device_enable_channel(const struct iio_device *dev, const char * channel, bool type)
+{
+	struct iio_channel *ch;
+	if (!channel)
+		return -EFAULT;
```
pcercuei, line 133: I don't think you need to check that. This will never happen.

tests/iio_common.c
```diff
@@ -126,6 +126,23 @@ struct iio_context * autodetect_context(bool rtn, const char * name, const char
 	return ctx;
 }
 
+ssize_t iio_device_enable_channel(const struct iio_device *dev, const char * channel, bool type)
```
pcercuei, line 129: Function return type could be just `int`.

rgetz: I think that caused an issue/warning on Windows - will double check...

tests/iio_common.c
rgetz: today - but not sure about how others might use us :) 

Removed.

tests/iio_common.c
rgetz: yeah, I didn't see the same warnings this time - changed and squashed.

pcercuei:
Merged, thanks!

PR #812, opened 2022-03-04
iiod: iiod-serial: Fix memory leak

pcercuei:
In the function start_serial_daemon, the "dev" variable was never freed
if the thread couldn't be created.

### DIFF
--- a/iiod/serial.c
+++ b/iiod/serial.c
@@ -262,6 +262,8 @@ int start_serial_daemon(struct iio_context *ctx, const char *uart_params,
 	if (err)
 		goto err_close_fd;
 
+	free(dev);
+
 	return 0;
 
 err_close_fd:
--- a/iiod/serial.c
+++ b/iiod/serial.c
@@ -258,7 +258,11 @@ int start_serial_daemon(struct iio_context *ctx, const char *uart_params,
 	IIO_DEBUG("Serving over UART on %s at %u bps, %u bits\n",
 		  dev, uart_bps, uart_bits);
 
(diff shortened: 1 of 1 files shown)


iiod/serial.c
```diff
@@ -258,7 +258,13 @@ int start_serial_daemon(struct iio_context *ctx, const char *uart_params,
...
 		  dev, uart_bps, uart_bits);
 
-	return thread_pool_add_thread(pool, serial_main, pdata, "iiod_serial_thd");
+	err = thread_pool_add_thread(pool, serial_main, pdata, "iiod_serial_thd");
+	if (err)
+		goto err_close_fd;
+
+	free(dev);
+
+	return 0;
```
nunojsa, line 267: Hmm, maybe there's more to it (or at least to be said in the commit description). This change seems to be more than just `free(dev);`. Now, in case `thread_pool_add_thread()` fails, we are also cleaning pdata and fd. Before we were just directly returning `thread_pool_add_thread()`. This came to my attention because I was thinking that we could do:

```
...
free(dev);
return thread_pool_add_thread();
...
```

Am I missing something?

pcercuei: @nunojsa If `thread_pool_add_thread` fails then we do want to clean up, yes. It should have been done before. You are right that it should be described in the commit message. I will just split it into a separate commit.

pcercuei: @nunojsa fixed now, thanks for the input.

nunojsa (approved):

PR #815, opened 2022-03-18
Remove support for bulk read/write of attributes

pcercuei:
Remove `iio_{device,channel,device_debug,device_buffer}_attr_{read,write}_all` functions for a big code cleanup, and because they are getting in the way of the new binary async. protocol implementation.

These functions may be re-implemented later in a much simpler manner.

### DIFF
--- a/device.c
+++ b/device.c
@@ -308,6 +308,9 @@ ssize_t iio_device_write_raw(const struct iio_device *dev,
 ssize_t iio_device_attr_read(const struct iio_device *dev,
 		const char *attr, char *dst, size_t len)
 {
+	if (!attr)
+		return -EINVAL;
+
 	if (dev->ctx->ops->read_device_attr)
 		return dev->ctx->ops->read_device_attr(dev,
 				attr, dst, len, IIO_ATTR_TYPE_DEVICE);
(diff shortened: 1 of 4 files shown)


device.c
```diff
@@ -650,6 +665,9 @@ ssize_t iio_device_debug_attr_read(const struct iio_device *dev,
 ssize_t iio_device_debug_attr_write_raw(const struct iio_device *dev,
 		const char *attr, const void *src, size_t len)
 {
+	if (!attr)
+		return -EINVAL;
```
nunojsa, line 669: nit: It looks like these little additions have not much to do with this specific patch? Maybe it would make sense to be in a separate patch. They are simple enough though, so fine by me...

nunojsa (approved):

device.c
pcercuei: @nunojsa they actually do have to see with the patch - all the attribute read/write did support bulk read/write of all attributes, which was activated when attr == NULL, and return everything into one giant buffer.

pcercuei:
CI fails to build but it seems to be since the last merge of the master branch and unrelated to this PR, so I'll merge it now.

PR #814, opened 2022-03-17
mdns.h: update to latest release to see if issues are resolved.

rgetz:
Signed-off-by: Robin Getz <robin.getz@analog.com>

PR #816, opened 2022-03-18
Use "git-subrepo" to manage in-tree dependencies

pcercuei:
"git-subrepo" is different than git submodules, in that the files are always present within the main tree.

The idea here, is that external dependencies that are meant to be in-tree are now managed by "git-subrepo", and commits touching these dependencies are strictly forbidden.

The only commits allowed, are the one that are automatically created
when adding a new dependency with:
`git subrepo clone https://... deps/new-dependency`

And the ones created when updating the external repository to the latest
available with:
`git subrepo pull deps/my-dependency`

git-subrepo is an external tool that can be obtained here:
https://github.com/ingydotnet/git-subrepo


pcercuei:
Since I couldn't determine the commit of our original in-tree `mdns.h`, I had to clone the latest version. Therefore I also cherry-picked @rgetz 's second commit in his PR #814.

pcercuei:
Also note that Codacy results should be ignored here. These commits add a lot of code which isn't directly managed by us.

PR #821, opened 2022-03-22
Update attribute read/write functions

pcercuei:
This PR reworks the attribute read/write functions.

It switches libiio to C11, and rename some functions so that `iio_device_attr_read()` and `iio_device_attr_write()` can be redefined as generic macros, which will accept multiple types.

It also adds a buffer ID argument to the buffer attribute read/write functions, in preparation of the multi-buffer support.

### DIFF
--- a/iio.h
+++ b/iio.h
@@ -750,22 +750,6 @@ __api __check_ret ssize_t iio_device_attr_read(const struct iio_device *dev,
 		const char *attr, char *dst, size_t len);
 
 
-/** @brief Read the content of all device-specific attributes
- * @param dev A pointer to an iio_device structure
- * @param cb A pointer to a callback function
- * @param data A pointer that will be passed to the callback function
- * @return On success, 0 is returned
- * @return On error, a negative errno code is returned
- *
- * <b>NOTE:</b> This function is especially useful when used with the network
- * backend, as all the device-specific attributes are read in one single
- * command. */
-__api __check_ret int iio_device_attr_read_all(struct iio_device *dev,
-		int (*cb)(struct iio_device *dev, const char *attr,
-			const char *value, size_t len, void *d),
-		void *data);
-
-
 /** @brief Read the content of the given device-specific attribute
  * @param dev A pointer to an iio_device structure
  * @param attr A NULL-terminated string corresponding to the name of the
@@ -1669,21 +1653,6 @@ __api __check_ret ssize_t iio_device_debug_attr_write(const struct iio_device *d
  * @return On error, a negative errno code is returned */
 __api __check_ret ssize_t iio_device_debug_attr_write_raw(const struct iio_device *dev,
 		const char *attr, const void *src, size_t len);
-
-
-/** @brief Set the values of all debug attributes
- * @param dev A pointer to an iio_device structure
- * @param cb A pointer to a callback function
- * @param data A pointer that will be passed to the callback function
- * @return On success, 0 is returned
- * @return On error, a negative errno code is returned
- *
- * <b>NOTE:</b> This function is especially useful when used with the network
- * backend, as all the debug attributes are written in one single command. */
-__api __check_ret int iio_device_debug_attr_write_all(struct iio_device *dev,
-		ssize_t (*cb)(struct iio_device *dev,
-			const char *attr, void *buf, size_t len, void *d),
-		void *data);
 
 
 /** @brief Read the content of the given debug attribute
--- a/iio.h
+++ b/iio.h
@@ -746,8 +746,9 @@ __api __check_ret __pure const char * iio_device_find_buffer_attr(
  * @param len The available length of the memory area, in bytes
  * @return On success, the number of bytes written to the buffer
  * @return On error, a negative errno code is returned */
-__api __check_ret ssize_t iio_device_attr_read(const struct iio_device *dev,
-		const char *attr, char *dst, size_t len);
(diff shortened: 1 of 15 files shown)


pcercuei:
Looks like it fails only on VS2019.
Would it be OK to drop support for VS2019 in the dev branch then?

iio.h
```diff
@@ -928,19 +842,13 @@ __api __check_ret int iio_device_buffer_attr_write_bool(const struct iio_device
...
- * @return On error, a negative errno code is returned */
-__api __check_ret int iio_device_buffer_attr_write_double(const struct iio_device *dev,
-		const char *attr, double val);
+#define iio_device_buffer_attr_write(buf, attr, val)			\
+	_Generic((val),							\
+		 const char *: iio_device_buffer_attr_write_string,	\
+		 char *: iio_device_buffer_attr_write_string,		\
+		 bool: iio_device_buffer_attr_write_bool,		\
+		 long long: iio_device_buffer_attr_write_longlong,	\
+		 double: iio_device_buffer_attr_write_double)(buf, attr, val)
```
nunojsa: Hmm not sure if we are being too restrictive here... While allowing for different pointer sign is arguable (personally, I would probably allow it; unless we compile with -Wpointer-sign), I would allow `unsigned long long` to be used. Not sure if it applies in some other macro

local.c
```diff
@@ -493,12 +495,17 @@ static ssize_t local_get_buffer(const struct iio_device *dev,
...
-		const char *attr, char *dst, size_t len, enum iio_attr_type type)
+				   unsigned int buf_id, const char *attr,
+				   char *dst, size_t len,
+				   enum iio_attr_type type)
 {
 	FILE *f;
 	char buf[1024];
 	ssize_t ret;
 
+	if (buf_id > 0)
```
nunojsa, line 506: nit: IMO, `if (!buf_id)` would be more readable. Like this, you are implicitly "playing" with the fact that buf_id is unsigned

nunojsa (approved):
Just some minor comments from my side that aren't blockers (for me) for this to be merged...

iio.h
pcercuei: The pointer types are different enough that it shouldn't cause a problem - I mean I wouldn't use a `void *` pointer in there.
About `long long`, here it's just a value, so the unsigned version would be implicitely cast. In the read function however it's a pointer, so the unsigned version needs to be explicitely cast.

The way I see it, is that before this PR you'd have specific API functions to read attributes, with the same types that are available here, so this PR doesn't change it. If we wanted to add support for unsigned integers here, I'd have to add an extra function (which I can still do, in a follow-up).

nunojsa: > About long long, here it's just a value, so the unsigned version would be implicitely cast. In the read function however it's a pointer, so the unsigned version needs to be explicitely cast.

Did you try to run the macro with an `unsigned long long` argument? AFAIK, `_Generic` does not see signed vs unsigned as type compatible so it should fail to compile as there's mo selectable type. That is why I brought this up...

pcercuei: For `unsigned long long` I guess it's possible to add an entry with the same `iio_device_buffer_attr_write_longlong`, and the argument will be cast implicitely.
(Some compilers like MSVC might warn about that though...)

nunojsa: > For unsigned long long I guess it's possible to add an entry with the same iio_device_buffer_attr_write_longlong, and the argument will be cast implicitely.

Yes, that is what I had in mind... No idea about the MSVC compilers though

local.c
pcercuei: Writing `if (!buf_id)` (which is the same as `buf_id == 0`) would change the behaviour though.

nunojsa: Ups :man_facepalming: . Yeah, different condition...

tfcollins:
> Looks like it fails only on VS2019. Would it be OK to drop support for VS2019 in the dev branch then?

I think that might be a bit aggressive. I think we just dropped 2016. 2019 is EoL 2029

pcercuei:
V2: We no longer compile as C11, but still use `_Generic()` in the public API. This makes it compile on VS2019.

trupples:
This change broke the [libiio ABI compatibility "promise"](https://github.com/analogdevicesinc/libiio/blob/v0.25/mainpage.dox#L269). While the changes were transparent to C code, projects that resolve `iio_device_attr_write` (etc.) at link-time are broken because these macros don't become symbols in the compiled library. (In my case, this breaks `libad9361-iio`, but I don't think this is particularly relevant)

I understand I'm more than 1yr late to the party and this probably won't change functionally. Rather, I'd like to ask for some input on how such compatibility-breaking changes could be better documented, because debugging this was quite a bother.

Also, [the v0.25 docs don't reflect these changes](http://analogdevicesinc.github.io/libiio/v0.25/libiio/group__Device.html#gaaa2d1867c15ef8f8424164d0ccea4dd8), though that's likely another issue in and of itself 🙃

pcercuei:
@trupples this change has never been merged in the v0.25 release, nor in the latest v0.x development branch, `libiio-v0`.

It was merged in the `dev` branch which has then be merged back to `master` (which is now called `main`), which is what will eventually become Libiio v1.0.

The ABI compatibility promise is only valid between versions with the same major version number. We updated the major version number in the `main` branch because we did want to break the ABI.

trupples:
Ah, I mistakenly thought it was merged into `main` before the v0.25 tag.

PR #823, opened 2022-03-25
CMake: Install public headers into iio/ folder

pcercuei:
Instead of installing iio.h into {prefix}/include/ directly, install it
inside {prefix}/include/iio/.

Alongside, install iio-debug.h, iio-backend.h, iio-lock.h and
iiod-client.h, since those export API functions that can be used by
modular backends, or simply by applications.

Since IIOD and all the tests/examples should also build out-of-tree,
they now include <iio/iio.h> instead of "iio.h". This means that we need
to move the headers in the tree to a include/iio/ folder, and update all
source files accordingly.


pcercuei:
This is ready for review.

mhennerich (approved):

PR #822, opened 2022-03-23
bug fix: recent merge of undating mdns library broke things

rgetz:
So try to fix it.

Signed-off-by: Robin Getz <robin.getz@analog.com>

PR #828, opened 2022-04-04
CI: Enable IPv6 on Windows

pcercuei:
Build libiio with support for IPv6 on Windows.


rgetz (approved):

rgetz:
With this - did you want to support ipv6 scanning in the new code (patch #822) ? (right now, it does not, but would be a quick update).

PR #838, opened 2022-04-07
mdns: early private networks on Windows are flakey

rgetz:
It takes some time for private networks (cross over cables, with
unconfigured networks on them) to propagate through Window's normal
routing tables (from what I can tell), so this just increases the
ability to work around that and get the answers when we need.

### DIFF
--- a/network.c
+++ b/network.c
@@ -301,16 +301,33 @@ static char * network_get_description(struct addrinfo *res)
 	if (res->ai_family == AF_INET6) {
 		struct sockaddr_in6 *in = (struct sockaddr_in6 *) res->ai_addr;
 		char *ptr;
-		inet_ntop(AF_INET6, &in->sin6_addr,
+		const char *ptr2;
+		ptr2 = inet_ntop(AF_INET6, &in->sin6_addr,
 				description, INET6_ADDRSTRLEN);
+		if (!ptr2) {
+			char buf[256];
+			iio_strerror(errno, buf, sizeof(buf));
+			IIO_ERROR("Unable to look up IPv6 address: %s\n", buf);
+			free(description);
+			return NULL;
+		}
 
 		if (IN6_IS_ADDR_LINKLOCAL(&in->sin6_addr)) {
 			ptr = if_indextoname(in->sin6_scope_id, description +
-					strlen(description) + 1);
+					strnlen(description, len) + 1);
 			if (!ptr) {
-				IIO_ERROR("Unable to lookup interface of IPv6 address\n");
-				free(description);
-				return NULL;
+#ifdef _WIN32
+				if (errno == 0) {
+					/* Windows uses numerical interface identifiers */
+					ptr = description + strnlen(description, len) + 1;
+					iio_snprintf(ptr, IF_NAMESIZE, "%u", in->sin6_scope_id);
+				} else
+#endif
+				{
+					IIO_ERROR("Unable to lookup interface of IPv6 address\n");
+					free(description);
+					return NULL;
+				}
 			}
 
 			*(ptr - 1) = '%';
--- a/network.c
+++ b/network.c
@@ -1145,6 +1145,17 @@ struct iio_context * network_create_context(const char *host)
 		ret = getaddrinfo(addr_str, port, &hints, &res);
 	} else {
 		ret = getaddrinfo(host, port, &hints, &res);
+#ifdef _WIN32
+		/* Yes, This is lame, but Windows is flakey with local addresses
+		 * WSANO_DATA = Valid name, no data record of requested type.
+		 * Normally when the host does not have the correct associated data
+		 * being resolved for. Just ask again. Try a max of 2.5 seconds.
+		 */
+		for (i = 0; HAVE_DNS_SD && ret == WSANO_DATA && i < 10; i++) {
+			Sleep(250);
+			ret = getaddrinfo(host, port, &hints, &res);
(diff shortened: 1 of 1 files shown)


dns_sd.c
```diff
@@ -98,6 +98,14 @@ static int dnssd_fill_context_info(struct iio_context_info *info,
 	}
 
 	ctx = network_create_context(uri);
+
+	/* Sometimes, on Windows local/private networks, numeric IP address
+	 * fails but hostname works (don't ask me why), so try both.
+	 */
```
pcercuei, line 104: Wouldn't it be better to try the hostname first then? Or does that cause problems on other places?

network.c
```diff
@@ -1139,6 +1140,21 @@ struct iio_context * network_create_context(const char *host)
...
+#ifdef _WIN32
+		/* Yes, This is lame, but Windows is flakey with local addresses
+		 * WSANO_DATA = Valid name, no data record of requested type.
+		 * Normally when the host does not have the correct associated data
+		 * being resolved for. Just ask again.
+		 */
+		if (HAVE_DNS_SD && ret == WSANO_DATA) {
+			i = 0;
+			while (ret == WSANO_DATA && i < 5) {
+				sleep(1);
```
pcercuei, line 1152: You can inline Sleep(1000) here.

pcercuei (changes requested):

dns_sd.c
rgetz: yeah, this causes problems in cases where the host name can't be looked up. On pre-Windows 10  - for the OS to do lookups for *.local addresses, you need to have bonjour installed - otherwise ip addresses are the only way it works.

network.c
rgetz: Ok - will update.

dns_sd.c
rgetz: this seems to only be a problem on private/local networks (unconfigured cross over cables) when you plug it in, and then check within the first 30 seconds or so, (at least that is when I noticed it); so I think IP addresses failing is the minor case...

pcercuei: Shouldn't that be protected with a `#ifdef _WIN32` guard then?

network.c
```diff
@@ -1139,6 +1139,21 @@ struct iio_context * network_create_context(const char *host)
...
+#ifdef _WIN32
+		/* Yes, This is lame, but Windows is flakey with local addresses
+		 * WSANO_DATA = Valid name, no data record of requested type.
+		 * Normally when the host does not have the correct associated data
+		 * being resolved for. Just ask again.
+		 */
+		if (HAVE_DNS_SD && ret == WSANO_DATA) {
+			i = 0;
+			while (ret == WSANO_DATA && i < 5) {
+				Sleep(500);
```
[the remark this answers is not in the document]
rgetz: yeah, I tested with both, and it seemed to only need a single 500ms delay most of the time (I saw once it looped twice) - I did it to one second before so it would be the same on linux and windows if we ever wanted to do that, but this does make it more responsive...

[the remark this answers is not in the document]
pcercuei: What about 25 times 100ms? Would it be faster?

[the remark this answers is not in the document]
rgetz: yes - updated.

rgetz:
fixed an issue that Codacy pointed out.

network.c
```diff
@@ -1145,6 +1162,21 @@ struct iio_context * network_create_context(const char *host)
...
 		ret = getaddrinfo(host, port, &hints, &res);
+#ifdef _WIN32
+		/* Yes, This is lame, but Windows is flakey with local addresses
+		 * WSANO_DATA = Valid name, no data record of requested type.
+		 * Normally when the host does not have the correct associated data
+		 * being resolved for. Just ask again. Try a max of 2.5 seconds.
+		 */
+		if (HAVE_DNS_SD && ret == WSANO_DATA) {
+			i = 0;
+			while (ret == WSANO_DATA && i < 25) {
```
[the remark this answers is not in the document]
rgetz: ok - and on more testing - poking at networking stack every 100ms - doesn't let it settle, so increase delay to 250ms. which seems to help.

network.c
```diff
@@ -1145,6 +1145,17 @@ struct iio_context * network_create_context(const char *host)
 		ret = getaddrinfo(addr_str, port, &hints, &res);
 	} else {
 		ret = getaddrinfo(host, port, &hints, &res);
+#ifdef _WIN32
+		/* Yes, This is lame, but Windows is flakey with local addresses
+		 * WSANO_DATA = Valid name, no data record of requested type.
+		 * Normally when the host does not have the correct associated data
+		 * being resolved for. Just ask again. Try a max of 2.5 seconds.
+		 */
+		for (i = 0; HAVE_DNS_SD && ret == WSANO_DATA && i < 10; i++)
```
pcercuei, line 1171: Missing opening brace {.

rgetz: yeah - sorry - copy paste from windows to linux to git...

rgetz:
so, I figured out what was going on (and why names sometimes worked, when addresses did not), and boils down to for local link addresses - Windows sets up a ipv6 routing table, but not an ipv4 one. If you have an use a name, you will get a ipv6 result.

How we are supporting ipv6 on windows - just handle this normally - and if ipv4 port knocking fails, remove it from the list, so you don't actually need to try by ipv4 and then by name....

PR #839, opened 2022-04-07
scan: ignore localhosts in scans

rgetz:
sometimes we get:

	2: 127.0.0.1 on Raspberry Pi 4 Model B Rev 1.5 [ip:phaser.local]
	3: 192.168.1.105 on Raspberry Pi 4 Model B Rev 1.5 [ip:phaser.local]
	4: 192.168.2.10 on Raspberry Pi 4 Model B Rev 1.5 [ip:phaser.local]
	5: 2601:190:400:da:a448:8c38:150a:7d4e on Raspberry Pi 4 Model B Rev 1.5 [ip:phaser.local]

obviously, 127.0.0.1 is not useful at all, so throw it away when we are looking for duplicates.

### DIFF
--- a/dns_sd.c
+++ b/dns_sd.c
@@ -251,6 +251,26 @@ void remove_dup_discovery_data(struct dns_sd_discovery_data **ddata)
 		}
 		i++;
 	}
+
+	prev = NULL;
+	ndata = d;
+	i = 0;
+	while (ndata->next) {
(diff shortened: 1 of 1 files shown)


rgetz:
yes, this needs something like this, but there is also a bug in here I found when doing some robustness testing - which I will commit in the morning after an overnight test...

rgetz:
After some robustness testing - this is Ok for review/merge now.

PR #840, opened 2022-04-09
Cmake: shift some iiod specific things into iiod Cmake

rgetz:
This moves all the iiod specific things out of the main Cmake file
and into the iiod specific cmake file, as well as prints out where
the startup scripts are being installed.


pcercuei:
Unless there is a good reason, I'm not merging this into master - it is already done in the "dev" branch and will only cause breakages and headaches when merging.

PR #843, opened 2022-04-21
debug messages: Add file, func and line number to message

rgetz:
I don't think this is exactly what we want - but I have found this helpful, and didn't want to loose it in re-basing...

-Robin

### DIFF
--- a/include/iio/iio-debug.h
+++ b/include/iio/iio-debug.h
@@ -14,9 +14,9 @@
 #define __api __iio_api
 
 #if defined(__MINGW32__)
-#   define __iio_printf __attribute__((__format__(gnu_printf, 3, 4)))
+#   define __iio_printf __attribute__((__format__(gnu_printf, 6, 7)))
 #elif defined(__GNUC__)
-#   define __iio_printf __attribute__((__format__(printf, 3, 4)))
+#   define __iio_printf __attribute__((__format__(printf, 6, 7)))
 #else
 #   define __iio_printf
 #endif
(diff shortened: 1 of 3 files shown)


include/iio/iio-debug.h
```diff
@@ -34,16 +34,17 @@
...
 #define __chn_dev_or_null(chn)	((chn) ? iio_channel_get_device(chn) : NULL)
 
-#define prm_err(prm, ...)	iio_prm_printf((prm), LEVEL_ERROR, "ERROR: " __VA_ARGS__)
-#define prm_warn(prm, ...)	iio_prm_printf((prm), LEVEL_WARNING, "WARNING: " __VA_ARGS__)
-#define prm_info(prm, ...)	iio_prm_printf((prm), LEVEL_INFO, __VA_ARGS__)
-#define prm_dbg(prm, ...)	iio_prm_printf((prm), LEVEL_DEBUG, "DEBUG: " __VA_ARGS__)
+#define prm_err(prm, ...)	iio_prm_printf((prm), LEVEL_ERROR, __FILE__, __func__, __LINE__, "ERROR: " __VA_ARGS__)
+#define prm_warn(prm, ...)	iio_prm_printf((prm), LEVEL_WARNING, __FILE__, __func__, __LINE__, "WARNING: " __VA_ARGS__)
+#define prm_info(prm, ...)	iio_prm_printf((prm), LEVEL_INFO, __FILE__, __func__, __LINE__, __VA_ARGS__)
+#define prm_dbg(prm, ...)	iio_prm_printf((prm), LEVEL_DEBUG, __FILE__, __func__, __LINE__, "DEBUG: " __VA_ARGS__)
```
nunojsa, line 47: I'm not sure adding both `__func__` and `__FILE__` adds much. For instance, if you have `__FILE__` and `__LINE__` you can already know exactly in which line of code your line is...

rgetz: yes - it's just easier/convenience.... __func__ is easier for me (to debug at distance and point people in the right direction) - since it's likely not to have identical debug/error text in the same function (hopefully)...

pcercuei:
I don't agree to have these unconditionally used - but there could be a `prm_trace` macro for that.

rgetz:
> I don't agree to have these unconditionally used - but there could be a `prm_trace` macro for that.

I understand - it make things more noisy for sure -- what about if it was just things are the ERROR level? or an additional WITH_LOCATION setting in the enum?

pcercuei:
I guess it could be useful for `prm_debug` (or a new prm_trace). I am less inclined to add it for `prm_err`, because to me this will make users think that there is an error with the library itself, while the vast majority of errors will have external causes...

Also note that you can use `__FILE__` and `__LINE__` as string literals instead of variables. Something like:
```c
#define __PRM_STRING(str) #str
#define _PRM_STRING(str) __PRM_STRING(str)

#define prm_dbg(prm, ...) \
iio_prm_printf((prm), LEVEL_DEBUG, "DEBUG: " __FILE__ ":" _PRM_STRING(__LINE__) ": " __VA_ARGS__)
```

Note that this does not work with `__func__` as it is a variable and not a string literal. The benefit is that you are not passing any extra variable, and you also don't need __iio_printf2.

PR #850, opened 2022-05-20
Fix data convert function in python bindings

tfcollins:
The python binding to convert did not work at all. This updates the convert function and was tested
with devices with various buffer datatypes.

### DIFF
--- a/bindings/python/iio.py
+++ b/bindings/python/iio.py
@@ -966,18 +966,41 @@ class Channel(object):
         """
         return ChannelType(_channel_get_type(self._channel))
 
-    def convert(self, dst, src):
+    def convert(self, src, buffer):
         """
         Convert src and saves the result in dst, using current channel's data format.
 
-        :param dst: type=list
-            The variable where the result is stored.
         :param src: type=list
             Data to be converted.
+        :param buffer: type=iio.Buffer
+            Associated iio.Buffer with channel data.
+        :return: type=list
+            The converted data.
         """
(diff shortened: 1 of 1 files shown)


tfcollins:
I can add my tests just need a couple days

pcercuei:
@tfcollins The convert function works as intended... it just doesn't perform what you expect it to do. Have a look at Channel.read() instead.

tfcollins:
@pcercuei The c function of convert operates on single samples. Are you expecting the python convert function to take a list of length 1? How is the python binding supposed to work?

pcercuei:
Yes, basically buffer.read() gives you a byte array, and you can convert individual values using channel.convert().

tfcollins:
That doesn't really make any sense base on the current method. If I pass a list as documented `(c_char * (len(dst) * self.data_format.length))(*dst)` will create a new object which I do not have access to. Then _convert will update that object and not a reference to the thing I passed.

pcercuei:
I was trying to remember how that function is supposed to be used, but it turns out it's not my code. Maybe @cristi-iacob can help?

tfcollins:
So if you want it to map more to the C API, I would argue we just return the pointer and not rely on pass-by object. I can put the looping logic and pointer math into a pyadi-iio method. Kinda like gr-iio.

-Travis

pcercuei:
@tfcollins what annoys me with your solution is that it's basically a duplicate of Channel.read().

I don't know why Channel.convert() was added, or if it's used anywhere, but we can't just rewrite it to make it functionally different, without breaking the API...

PR #852, opened 2022-06-08
Remove python bindings to convert method

tfcollins:
The convert method from the Channel class had no functionality and the
read method performs the desired operation without requiring external
loops. Removing the method and dependent ctype definitions.


pcercuei:
@tfcollins I believe you can remove the convert_inverse() too.

pcercuei (approved):

PR #855, opened 2022-06-20
CMake: When using pkg-config, link libusb using its full path

pcercuei:
When libusb-1.0 is detected using pkg-config, we should include the
library's directory to the link path.

Since target_link_directories() is not available in the minimum version
of CMake that we support, override LIBUSB_LIBRARIES with the full path
to the library, to ensure that it will build fine even if the library is
installed at a non-standard path.


pcercuei:
@ffontaine, just checking with you that I'm not breaking your fix.

pcercuei:
Understood, thanks.

ccraluca (approved):

PR #858, opened 2022-06-30
iiod-client: Use iiod_client_write_all() in iiod_client_exec_command()

pcercuei:
It is not guaranteed that the command will be sent in a single call to
.write(). Therefore we need to send the command with
iiod_client_write_all().


pcercuei:
The Codacy issue can be ignored, since I'm not changing the behaviour here.

pcercuei:
@mhennerich I'd like to merge this before tagging v0.34…

mhennerich (approved):

PR #872, opened 2022-08-11
Rework XML context creation

pcercuei:
- Add support for creating a XML from a XML string embedded into a URI
- This means that every possible IIO context can now be created with the single function `iio_create_context`.
- The public function `iio_create_xml_context_mem` is now removed.
- The IIOD client code will now build a URI using the received XML string, and create a IIO context the standard way.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -271,6 +271,7 @@ stages:
     steps:
     - script: |
         set -e
+        sudo apt-get update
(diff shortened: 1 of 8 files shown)


pcercuei:
Added a fix to the Azure script to `apt-get update` before installing packages, to get the CI to work.

mhennerich (approved):

PR #877, opened 2022-08-26
libiio / iiod async interface fixes

pcercuei:
Some fixes for the new libiio / IIOD async. interface.

- Fix IIOD sometimes answering a -ETIMEDOUT (-110) error, as the iiod_responder code did not handle infinite timeouts properly;
- Fix deadlock on client side occurring when the context creation failed (e.g. when IIOD returns a timeout...);
- Fix IIOD not understanding "write attribute" commands;
- Fix iiod-client sending invalid commands when trying to access an attribute that does not exist;
- Fix the network backend not destroying the iiod-client properly.
- Fix the dynamic loading code which was completely broken (wrong include paths on Linux, undefined variable on Windows)
- Fix prototype of iio_buffer_get_channels_mask()
- Remove carriage returns at the end of strings used in prm_perror
- Fix iio_block_destroy() blocking until timeout happens

### DIFF
--- a/iiod-client.c
+++ b/iiod-client.c
@@ -237,6 +237,12 @@ static ssize_t iiod_client_read_all(struct iiod_client *client,
 	return (ssize_t) (ptr - (uintptr_t) dst);
 }
 
+static void iiod_client_cancel(struct iiod_client *client)
+{
+	if (client->ops->cancel)
+		client->ops->cancel(client->desc);
+}
+
 struct iiod_client * iiod_client_new(const struct iio_context_params *params,
 				     struct iiod_client_pdata *desc,
 				     const struct iiod_client_ops *ops)
@@ -269,8 +275,10 @@ struct iiod_client * iiod_client_new(const struct iio_context_params *params,
 	return client;
 
 err_free_responder:
-	if (client->responder)
+	if (client->responder) {
+		iiod_client_cancel(client);
 		iiod_responder_destroy(client->responder);
+	}
 err_free_lock:
 	iio_mutex_destroy(client->lock);
 err_free_client:
@@ -280,8 +288,10 @@ err_free_client:
 
 void iiod_client_destroy(struct iiod_client *client)
 {
-	if (client->responder)
+	if (client->responder) {
+		iiod_client_cancel(client);
 		iiod_responder_destroy(client->responder);
+	}
 
 	iio_mutex_destroy(client->lock);
 	free(client);
--- a/iiod-client.c
+++ b/iiod-client.c
@@ -535,6 +535,9 @@ static ssize_t iiod_client_read_attr_new(struct iiod_client *client,
 			if (!strcmp(iio_channel_get_attr(chn, i), attr))
 				break;
 
+		if (i == iio_channel_get_attrs_count(chn))
+			return -ENOENT;
+
 		arg1 = (uint16_t) i;
 	} else {
 		switch (type) {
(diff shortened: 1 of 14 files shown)


iiod-responder.c
```diff
@@ -376,13 +376,18 @@ static int iiod_enqueue_command(struct iiod_io *writer, uint8_t op,
...
 	iio_mutex_lock(io->lock);
 
-	done = (io->write_token && iio_task_is_done(io->write_token)) ||
-		(read_counter_us() - io->w_io.start_time > timeout_us);
+	done = io->write_token && iio_task_is_done(io->write_token);
+
+	if (io->timeout_ms) {
+		timeout_us = io->timeout_ms * 1000;
+
+		done |= read_counter_us() - io->w_io.start_time > timeout_us;
```
nunojsa, line 389: nit: I guess we do not need `|=`?

iiod-responder.c
```diff
@@ -414,33 +421,47 @@ bool iiod_io_has_response(struct iiod_io *io)
...
 
 	while (!io->r_done) {
-		diff_ms = (read_counter_us() - io->r_io.start_time) / 1000;
-
-		if (diff_ms < timeout_ms)
-			ret = iio_cond_wait(io->cond, io->lock, timeout_ms - diff_ms);
-		else
-			ret = -ETIMEDOUT;
-
+		ret = iiod_io_cond_wait(io);
```
nunojsa, line 458: Hmmm, not a big deal as the change is not that big but maybe we should introduce the `iiod_io_cond_wait()` helper in it's own patch?

iiod-responder.c
```diff
@@ -414,33 +421,47 @@ bool iiod_io_has_response(struct iiod_io *io)
...
+
+	if (!io->timeout_ms)
+		return false;
+
+	timeout_us = io->timeout_ms * 1000;
+
+	return read_counter_us() - io->w_io.start_time > timeout_us;
+}
+
+static int iiod_io_cond_wait(struct iiod_io *io)
```
nunojsa, line 435: Can we make `struct iiod_io *io() `const`? Or the calls to `iio_cond_wait()` will WARN?

iiod-client.c
```diff
@@ -280,8 +284,12 @@ struct iiod_client * iiod_client_new(const struct iio_context_params *params,
 
 void iiod_client_destroy(struct iiod_client *client)
 {
-	if (client->responder)
+	if (client->responder) {
+		if (client->ops->cancel)
+			client->ops->cancel(client->desc);
```
nunojsa, line 289: nit: maybe we could introduce an helper to avoid the nested checks... Not that terrible though

nunojsa (approved):

iiod-client.c
pcercuei: As long as it doesn't require gymnastics to fit under the 80-characters rule, and the pattern is not duplicated elsewhere, I think it can stay inline.

iiod-responder.c
pcercuei: We actually do - check the original code that it replaces and you'll see why.

iiod-responder.c
pcercuei: Yes it can be const, I'll change it. It will be fine with `io->cond` and `io->lock` because those are pointers, so making `io` const will make these immutable pointers and not pointers to const data.

iiod-responder.c
pcercuei: Ok.

pcercuei:
@nunojsa I updated the code with your suggestions. It should be *perfect* now :)

iiod-responder.c
nunojsa: You're right... I was failing to see the first assignment to `done`. But this now raised another (not that important) question to me... I think we could extend the check to be `if (io->timeout_ms && !done)`. IIUC, there's no point in doing anything else if we are already done. That said, we could safely drop the OR...

nunojsa (approved):

pcercuei:
@nunojsa fair enough, the `|=` is gone now.

PR #886, opened 2022-08-29
update man pages

maitbot:
Debian wants each user command documented.
This patch has some help2man generated output, might
be better to follow up and further improve things before merging.

Signed-off-by: A. Maitland Bottoms <bottoms@debian.org>


pcercuei:
@maitbot if you don't mind, I think the .TH should be lowercase, and the iio_adi_xflow_check command line should not include any paths (it says `./tests/iio_adi_xflow_check` right now). Otherwise looks good.

pcercuei:
Closing this one, as #959 was merged.

PR #897, opened 2022-09-02
fix op codes in attr_write

catkira:
There was a copy paste error, all the opcodes where xxx_READ_xxx, but they have to be xxx_WRITE_xxx.

### DIFF
--- a/iiod/responder.c
+++ b/iiod/responder.c
@@ -148,14 +148,14 @@ static ssize_t attr_write(struct parser_pdata *pdata,
 	dev = iio_context_get_device(pdata->ctx, cmd->dev);
 
 	switch (cmd->op) {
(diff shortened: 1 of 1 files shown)


pcercuei:
A few notes:

- Please make the merge target the `dev` branch, since the bug is there. We only accept pull requests on the main branches, which are "master" for 0.x and "dev" for 1.x.
- The first line of the commit should be a title that describes briefly the changes, e.g. "iiod: Fix writing of attributes", followed by a small paragraph that describes a bit more in-depth the change (your "there was a copy-paste error" explanation can go there)
- You need to sign your commit, adding a `Signed-off-by:` line followed by your full name + usual email address. This is for copyright attribution.

catkira:
ok I will do that. I was afraid the ADI weekend had already started :P

Issue #894, opened 2022-09-02
ad9361-iiostream does not work for Pluto because of "rf_port_select"


catkira:
ad9361-iiostream does not work for Pluto because "rf_port_select" is disabled in the device tree for Pluto.
It would be nice to just catch this error and let the application continue. That way it is also usable on Pluto.

PR #900, opened 2022-09-03
ad9361-iiostream: Fix rf_port_select for PlutoSDR devices

catkira:
PlutoSDR devices don't support changing the rf_port_select attribute,
because its fixed in the device tree. Therefore this error should not lead
to a shutdown for these type of devices.

Fixes: https://github.com/analogdevicesinc/libiio/issues/894

Signed-off-by: Benjamin Menküc <benjamin@menkuec.de>

### DIFF
--- a/examples/ad9361-iiostream.c
+++ b/examples/ad9361-iiostream.c
@@ -15,6 +15,7 @@
 #include <string.h>
 #include <signal.h>
 #include <stdio.h>
+#include <errno.h>
 
 /* helper macros */
 #define MHZ(x) ((long long)(x*1000000.0 + .5))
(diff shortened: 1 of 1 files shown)


catkira:
I also thought about detecting the device by the hwinfo string and then just skip the rf_port_select for pluto. But I decided against it, because there are pluto derivatives out there that have the same problem.

pcercuei:
I think the error will be -ENOENT (aka. "no such file or directory"), you could ignore this error in particular and still handle the other ones.

catkira:
ok, I will adopt that this evening. I m sitting in a plane atm and dont have my device connected.

catkira:
@pcercuei I updated it. It gives a `EINVAL` error.

examples/ad9361-iiostream.c
```diff
@@ -162,7 +163,11 @@ bool cfg_ad9361_streaming_ch(struct stream_cfg *cfg, enum iodev type, int chid)
 	// Configure phy and lo channels
 	printf("* Acquiring AD9361 phy channel %d\n", chid);
 	if (!get_phy_chan(type, chid, &chn)) {	return false; }
-	wr_ch_str(chn, "rf_port_select",     cfg->rfport);
+	int ret = iio_channel_attr_write_string(chn, "rf_port_select", cfg->rfport);
+	if (ret == -EINVAL)
+		printf("Error writing %s to channel \"%s\"\nThis error can be ignored for PlutoSDR and derivatives.\n", cfg->rfport, "rf_port_select");
```
pcercuei, line 168: You don't really need to use a "%s" for "rf_port_select", just write it in-line.
Also please break line before the parameters to avoid the line to be super-long.

pcercuei:
@mhennerich thoughts about this commit?

examples/ad9361-iiostream.c
catkira: True. I force pushed it again.

catkira:
> > @pcercuei I updated it. It gives a `EINVAL` error.
> 
> Does that mean that a "rf_port_select" attribute is present, and it just doesn't like the value set to it?

The attribute is present, but it is fixed in the device tree, because the pluto hardware has only one output port of the ad9361 connected to a hardware connector. There is not a suitable error code for this circumstance yet. I just tried it on hardware and saw that the returned error code in this case is EINVAL.

pcercuei:
@mhennerich gentle ping, does that look good to you?

pcercuei (approved):

pcercuei:
No feedback, so I'll just merge it.

Issue #901, opened 2022-09-03
ad9361-iiostream crashes when running on host


catkira:
ad9361 crashes when running on host (tested with the latest dev branch):
```
benja@LAPTOP:/mnt/d/git/libiio/build/examples$ ./ad9361-iiostream ip:192.168.137.2
* Acquiring IIO context
* Acquiring AD9361 streaming devices
* Configuring AD9361 for streaming
* Acquiring AD9361 phy channel 0
Error A_BALANCED writing to channel "rf_port_select"
value may not be supported. This error can be ignored for PlutoSDR and derivatives.
* Acquiring AD9361 RX lo channel
* Acquiring AD9361 phy channel 0
Error A writing to channel "rf_port_select"
value may not be supported. This error can be ignored for PlutoSDR and derivatives.
* Acquiring AD9361 TX lo channel
* Initializing AD9361 IIO streaming channels
* Enabling IIO streaming channels
* Creating non-cyclic IIO buffers with 1 MiS
* Starting IO streaming (press CTRL+C to cancel)
        RX     0.26 MSmp, TX     0.26 MSmp
ERROR: Unable to dequeue block: Bad file descriptor (9)
ERROR: Unable to send block: Bad file descriptor (9)
* Destroying streams
* Destroying buffers
* Destroying channel masks
* Destroying context
```
it is working when I run in directly on device and it is also working on host when I outcomment all tx stuff.

pcercuei:
Works fine here. With firmware v0.34 on the Pluto.

catkira:
I experienced the crash when using libiio 1.0 on host _and_ device.

PR #903, opened 2022-09-06
iiod-client: Dequeueing an already dequeued block is not an error

pcercuei:
Dequeueing an already dequeued block is not an error, but a no-op.
Return 0 instead of -EBADF.

Fixes: #901.

### DIFF
--- a/iiod-client.c
+++ b/iiod-client.c
@@ -1659,7 +1659,7 @@ int iiod_client_enqueue_block(struct iio_block_pdata *block,
 	iio_mutex_lock(block->lock);
 
 	if (block->enqueued) {
-		ret = -EBUSY;
+		ret = -EPERM;
 		goto out_unlock;
 	}
 
@@ -1686,7 +1686,7 @@ int iiod_client_dequeue_block(struct iio_block_pdata *block, bool nonblock)
 	iio_mutex_lock(block->lock);
 
 	if (!block->enqueued) {
-		ret = -EBADF;
+		ret = -EPERM;
 		goto out_unlock;
 	}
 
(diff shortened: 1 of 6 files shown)


iiod-client.c
```diff
@@ -1685,10 +1685,8 @@ int iiod_client_dequeue_block(struct iio_block_pdata *block, bool nonblock)
 
 	iio_mutex_lock(block->lock);
 
-	if (!block->enqueued) {
-		ret = -EBADF;
+	if (!block->enqueued)
```
nunojsa, line 1688: I agree is not an error but maybe something that apps should not do and try to fix... Just wondering if a warning or even debug message would make sense?

nunojsa (approved):

iiod-client.c
pcercuei: So I'm just trying to make the behaviour consistent across the library. If we consider it is an error and not a no-op to dequeue an already dequeued block, then I'll have to change other parts of the library, because right now it's inconsistent.

I am fine with both approaches. But I'd prefer to return an error if it's something that the apps should not do.

pcercuei:
@catkira ad9361-iiostream uses iio_stream_get_next_block(), so the stream.c fix makes it work.

The function is not *that* complicated, your solution is less readable IMHO. Besides with my latest force-push the code is even simpler.

catkira:
@pcercuei yes, the function looks much much easier to read after the force push.
Now only the !is_tx inside the for loop prevents me from fast reading the code. I think its because it is an uncommon pattern and one expects a for loop to just repeat the body x times.
Instead of 
```c
    if ()        
        for (i = 1; !is_tx && i < stream->nb_blocks; i++) {
        // ...
        }
}
```
I would donate one more line and do
```c
    if (!is_tx)        
        for (i = 1; i < stream->nb_blocks; i++) {
        // ...
        }
}
```

PR #906, opened 2022-09-12
Compatibility layer overhaul

pcercuei:
Switch to using dynamic loading of symbols.

Instead of linking the compatibility library to the Libiio v1.x library,
it will now dlopen() the v1.x library. This makes it possible to wrap
functions that changed their prototype in the new library, for instance
iio_device_create_buffer().

The compatibility library also now covers the whole Libiio v0.x API.


compat.c
```diff
@@ -176,59 +657,1456 @@ ssize_t iio_scan_context_get_info_list(struct iio_scan_context *ctx,
...
+	blk->info = NULL;
+	blk->ctx_cnt = iio_scan_context_get_info_list(blk->ctx, &blk->info);
 
-		free(info->description);
-		free(info->uri);
+	return blk->ctx_cnt;
+}
+
+struct iio_context_info *iio_scan_block_get_info(
+		struct iio_scan_block *blk, unsigned int index)
```
nunojsa, line 687: nit:

```
struct iio_context_info *
iio_scan_block_get_info(struct iio_scan_block *blk, unsigned int index);
```

compat.c
```diff
@@ -176,59 +657,1456 @@ ssize_t iio_scan_context_get_info_list(struct iio_scan_context *ctx,
...
 	}
 
-	free(list);
+	return blk->info[index];
 }
 
-const char *
-iio_context_info_get_description(const struct iio_context_info *info)
+struct iio_scan_block *iio_create_scan_block(
+		const char *backend, unsigned int flags)
```
nunojsa, line 698: ditto...

compat.c
```diff
@@ -176,59 +657,1456 @@ ssize_t iio_scan_context_get_info_list(struct iio_scan_context *ctx,
...
+  static void f(void)
+#ifdef _WIN64
+#define _CONSTRUCTOR(f) __CONSTRUCTOR(f, "")
+#else
+#define _CONSTRUCTOR(f) __CONSTRUCTOR(f, "_")
+#endif
+#elif defined(__GNUC__)
+#define _CONSTRUCTOR(f) static void __attribute__((constructor)) f(void)
+#else
+#define _CONSTRUCTOR(f) static void f(void)
```
nunojsa, line 2097: I might be missing something but how is compat_lib supposed to be initialized in this case?

compat.c
```diff
@@ -176,59 +657,1456 @@ ssize_t iio_scan_context_get_info_list(struct iio_scan_context *ctx,
...
+	}
+
+#define FIND_SYMBOL(lib, symbol) \
+	do { \
+		ctx->symbol = iio_dlsym(lib, #symbol); \
+		if (!ctx->symbol) { \
+			fprintf(stderr, "Unable to find symbol: %s\n", #symbol); \
+			err = -ENOENT; \
+			goto err_dlclose; \
+		} \
```
nunojsa, line 1916: I honestly do not know if we can ever get any strange issue but according to the `dlsym()` man page, the proper way to handle errors is:

```
dlerror(); 
ctx->symbol = iio_dlsym(lib, #symbol);
if (dlerror()) {
     /* error handling */
}

```

Apparently in some weird cases it's actually expected that `dlysm()` returns NULL

compat.c
```diff
@@ -4,37 +4,508 @@
...
+
+		dev_compat = (void *)((uintptr_t)compat
+				      + sizeof(*compat)
+				      + i * sizeof(*dev_compat));
+		dev_compat->nb_channels = nb_channels;
+		dev_compat->is_tx = iio_device_is_tx(dev);
+
+		IIO_CALL(iio_device_set_data)(dev, dev_compat);
+
+		dev_compat->nb_kernel_buffers = 4;
```
nunojsa, line 403: nit: I guess we could do this before calling `iio_device_set_data()`...

nunojsa (commented):
Just some minor stuff...

compat.c
```diff
@@ -4,37 +4,508 @@
...
+	char *(*iio_strdup)(const char *);
+	void (*iio_strerror)(int, char *, size_t);
+	void (*iio_prm_printf)(const struct iio_context_params *params,
+			       unsigned int msg_level,
+			       const char *fmt, ...);
+};
+
+static struct compat *compat_lib_ptr;
+
+#define IIO_CALL(fn) (*compat_lib_ptr->fn)
```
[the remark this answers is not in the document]
pcercuei: There is a reason why it's done this way: `__VA_ARGS__` must have at least one argument.

So this:
`#define IIO_CALL(fn, ...) (*compat_lib_ptr->fn)(__VA_ARGS__)`

would make it impossible to call like this:
`IIO_CALL(iio_get_backends_count);`

iio-debug.h has the same problem and gets around it with some black magic. But in this case it seemed smarter to define `IIO_CALL` like that.

compat.c
pcercuei: My man page says `On failure, they return NULL`.

compat.c
```diff
@@ -4,37 +4,508 @@
...
+	int err;
+
+	nb_devices = IIO_CALL(iio_context_get_devices_count)(ctx);
+
+	for (i = 0; i < nb_devices; i++) {
+		dev = IIO_CALL(iio_context_get_device)(ctx, i);
+		nb_channels = IIO_CALL(iio_device_get_channels_count)(dev);
+
+		size += sizeof(struct iio_device_compat)
+			+ nb_channels * sizeof(struct iio_channel_compat);
```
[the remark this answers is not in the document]
pcercuei: I am the first one to use this pattern, when the variable is the target of the allocation.
In this case, `dev_compat` and `chn_compat` are used much later in the code, it felt strange for me to use them there...

But I don't really mind, if it gets this PR reviewed, I'm happy to change it.

compat.c
pcercuei: Right. I guess it would make more sense to use an `#error` and only enable the compat library with supported compilers.

compat.c
pcercuei: Nevermind, I overlooked it.

Anyway... I honestly think we can ignore this case. I can assure it will never happen. Programs are loaded at address 0x0 on Linux, unless requested otherwise during the linking phase with e.g. `-Wl,-Ttext-segment=0x10000000`. In both cases, the run-time linker will load the shared libraries at a higher address; so I do not see any possible case where we get a symbol at address 0x0.

compat.c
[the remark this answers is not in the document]
nunojsa: Ahh I remember I also had to do some magic in the adrv9002 stream example... I guess this is because we cannot tak gnu extensions as granted? IIRC, with the extension we don't have this constrain.

compat.c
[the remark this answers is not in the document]
nunojsa: no need to change... I don't really have any strong feeling either

compat.c
nunojsa: Agree...

compat.c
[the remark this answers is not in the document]
pcercuei: Yes, for instance Visual Studio does not have the GNU extensions ;)

compat.c
nunojsa: I'll just paste the man page

``` text
There are several scenarios when the address of a global symbol is NULL.  For example, a symbol can be placed at
zero address by the linker, via a linker script or with --defsym command-line option.  Undefined weak symbols also have NULL value.  Finally, the symbol value may be the result of a GNU  indirect  function  (IFUNC) resolver function that
returns NULL as the resolved value.  In the latter case, dlsym() also returns NULL without error.  However, in the former
two cases, the behavior of GNU dynamic linker is inconsistent: relocation processing succeeds and the symbol can be
observed to have NULL value, but dlsym() fails and dlerror() indicates a lookup error.
```
I guess the above is likely not to happen with us but OTOH, the code for handling this properly is not *so* horrible so I would probably play safe in here.

That said, is really up to you... I won't be pushing for this

dynamic-unix.c
```diff
@@ -26,7 +26,8 @@ void * iio_dlsym(void *lib, const char *symbol)
 {
 	void *ptr;
 
+	dlerror();
 	ptr = dlsym(lib, symbol);
 
-	return ptr ? ptr : iio_ptr(-EINVAL);
+	return (ptr || !dlerror()) ? ptr : iio_ptr(-EINVAL);
```
nunojsa, line 32: nit: maybe this would make it more simple?

`return dlerror() ? iio_ptr(-EINVAL) : ptr;`

nunojsa (approved):
LGTM

dynamic-unix.c
pcercuei: Well, is it in the spec that dlerror() returns NULL when dlsym() suceeds?

nunojsa: `dlerror() returns NULL if no errors have occurred since initialization or since it was last called.`

I guess this is why we first call it before calling `dlysm()` so it's internal state is cleaned...

pcercuei: Hmmmk. Fixed.

nunojsa (approved):

PR #913, opened 2022-10-12
fixing bindings

misko:
Errors are being thrown from pyadi library because integer arguments for some libiio calls are not being passed through. The reason for this is that the argtypes are not correctly specific in iio.py


nunojsa:
Hi @misko, thanks for the patches...

Please add your S-o-b tag to the patches. Run `git commit` with -s. The first patch also needs some more description other than "fixing bindings". It's important to explain what are you actually fixing in the commit message.

misko:
Hey @nunojsa , I squashed the commits and added a better description. I don't know what an S-o-b tag is however.

nunojsa:
> I don't know what an S-o-b tag is however.

S-o-b stands for Signed-off-by. Note libiio `git log` for example in (here)(https://github.com/analogdevicesinc/libiio/commit/cee2d1364b12221b37ede3591bb37a623c7b313e). From the linux kernel docs:

"The Signed-off-by: tag indicates that the signer was involved in the development of the patch, or that he/she was in the patch's delivery path." and we also require it for our projects (or most of them AFAIK)... As I said, you can do it by running `git commit -s` or in your case `git commit -s --amend`

misko:
@nunojsa awesome, done! thank you for explaining :)

rgetz:
> > I don't know what an S-o-b tag is however.
> 
> S-o-b stands for Signed-off-by. Note libiio `git log` for example in [here](https://github.com/analogdevicesinc/libiio/commit/cee2d1364b12221b37ede3591bb37a623c7b313e). From the linux kernel docs:
> 
> "The Signed-off-by: tag indicates that the signer was involved in the development of the patch, or that he/she was in the patch's delivery path." and we also require it for our projects (or most of them AFAIK)... As I said, you can do it by running `git commit -s` or in your case `git commit -s --amend`

The "Signed-off-by" tag is more than "involved", it means you can agree to the [Developer Certificate of Origin](https://developercertificate.org/). It's partly legally - do you have the rights/permissions to contribute the code, and a GPDR disclaimer (once you commit something, git log never forgets).

-Robin

pcercuei:
The ubuntu 22.04 build failure can be ignored, it chokes on something unrelated.

@misko, could you improve slightly the commit message? Your first comment in this thread has more information than the commit itself which is never good. You could just add the text of that comment below the commit's title ;)

rgetz:
yeah, the build failures pointed out here are fixed in #915

pcercuei (approved):

PR #917, opened 2022-11-11
CMake: Fix detection of libxml2

pcercuei:
Without the NO_SYSTEM_ENVIRONMENT_PATH option, find_package() would result on my system in a random toolchain's libxml2 library/include paths to be picked, just because it happens to be in $PATH (with a low priority).


tfcollins:
I guess this assumes a LibXml2Config.cmake exists somewhere?

pcercuei:
@tfcollins I didn't debug it much further, but I *think* what happens is that my system's libxml2-config.cmake tries to run the `libxml2-config` program, and for some reason it picks the one from my toolchain.

tfcollins:
Kk I've never really looked hard at what find_package searches but I guess this makes sense

tfcollins (approved):

PR #924, opened 2023-01-09
Clarify licenses

pcercuei:
Libiio was always meant to be LGPL, but the tools, examples and IIOD were supposed to be GPL (as it does not really make sense for them to use LGPL).

However, the SPDX identifiers (and the comment blocks before that) of the source files all stated that the license was LGPL, since the beginning.

Since re-licensing would be quite a headache, because of the huge number of copyright holders, simply embrace the mistake and use the LGPL license for everything.


pcercuei:
@rgetz is that OK for you?

nunojsa (approved):

rgetz:
No. It's not Ok.

The examples were GPL; and you moved them to LPGL when you moved to the short license.
https://github.com/analogdevicesinc/libiio/commit/3a85aa9fc88b37964b581b5949881bcc6c37a85a

The examples/test cases had been GPL for the last 4 years...
https://github.com/analogdevicesinc/libiio/issues/350

and was done to handle upstream comments.
https://bugzilla.redhat.com/show_bug.cgi?id=1426465

pcercuei:
@rgetz you are totally right. I will fix it ASAP, sorry about that.

@nunojsa you created `examples/adrv9002-iiostream.c` with a LGPLv2+ license, I will re-license it to GPLv2+, if you are okay with it.

pcercuei:
@rgetz Actually some files in tests/ were already LGPL, see e65a9786.

The `tests/gen_code.{c,h}` were only touched by you and me, so we could re-license the code to GPL, but `iio_stresstest.c` was also touched by @matejk. We would need his agreement to re-license the code as GPL.

rgetz:
Ok - take this as my approval to relicense iio_stresstest.c as GPL 2 or greater.

@matejk - can you provide the same here? Your change is here:
https://github.com/analogdevicesinc/libiio/commit/c954463edd1f39eca5ba18f06f11992c14d6afca

(which isn't really any "code", it's just the inclusion of a header file, but just to be certain, it would be great to get your approval).

Thanks
-Robin

matejk:
@rgetz, I agree with the change if the license.

larsclausen:
I'm ok with the license change

nunojsa:
> @nunojsa you created examples/adrv9002-iiostream.c with a LGPLv2+ license, I will re-license it to GPLv2+, if you are okay with it.

I'm fine with the change...

PR #928, opened 2023-01-20
DMABUF API v2

pcercuei:
These commits will modify how the local-DMABUF interface works, because the interface itself in the kernel changed (as it is not yet upstream).

The local backend will now make use of the "udmabuf" kernel driver, which was added in 5.19, to create DMABUF objects from memfd buffers. These DMABUFs can then be attached to the IIO device and/or other devices from other subsystems, and memory-mapped to access the data.

IIOD has also been modified to support enqueueing DMABUFs to the USB (functionfs) interface, and will support passing the buffers between IIO and USB in a zero-copy fashion.

From a Libiio API standpoint, some changes were introduced:
- `iio_block_enqueue` will support partial transfers on RX buffers as well;
- Setting bytes_used=0 in `iio_block_enqueue` will default to transferring the full buffer;
- New API function `iio_block_disable_cpu_access`, to disable CPU synchronization with the buffers when doing zero-copy;
- New API function `iio_block_get_dmabuf_fd` which can be used when doing zero-copy as well.

### DIFF
--- a/iiod/responder.c
+++ b/iiod/responder.c
@@ -751,7 +751,7 @@ static void handle_transfer_block(struct parser_pdata *pdata,
 	int ret;
 
 	buf = get_iio_buffer(pdata, cmd, &entry);
-	ret = iio_err(block);
+	ret = iio_err(buf);
 	if (ret)
 		goto out_send_response;
 
--- a/iiod/responder.c
+++ b/iiod/responder.c
@@ -15,6 +15,7 @@
 #include <fcntl.h>
 #include <iio/iio-lock.h>
 #include <pthread.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
@@ -341,7 +342,17 @@ static int buffer_dequeue_block(void *priv, void *d)
 	if (ret < 0)
 		goto out_send_response;
 
-	if (!iio_buffer_is_tx(buffer->buf)) {
+	if (!buffer->is_tx) {
+		if (WITH_IIOD_USB_DMABUF && entry->dmabuf_fd > 0) {
+			/* We need to send the error code before the data.
+			 * If usb_transfer_dmabuf() fails, we're screwed... */
+			iiod_io_send_response_code(entry->io, entry->bytes_used);
+
+			return usb_transfer_dmabuf(buffer->pdata->fd_out,
+						   entry->dmabuf_fd,
+						   entry->bytes_used);
+		}
+
 		data.ptr = iio_block_start(entry->block);
 		data.size = iio_block_end(entry->block) - data.ptr;
 		nb_data++;
@@ -379,6 +390,8 @@ static void handle_create_buffer(struct parser_pdata *pdata,
 		ret = -ENOMEM;
 		goto err_send_response;
 	}
+
+	entry->pdata = pdata;
 
 	nb_channels = iio_device_get_channels_count(dev);
 	nb_words = (nb_channels + 31) / 32;
@@ -444,6 +457,8 @@ static void handle_create_buffer(struct parser_pdata *pdata,
 		else
 			entry->words[BIT_WORD(i)] &= ~BIT_MASK(i);
 	}
+
+	entry->is_tx = iio_buffer_is_tx(buf);
 
 	/* Success, destroy the temporary mask object */
 	iio_channels_mask_destroy(mask);
@@ -629,13 +644,14 @@ static void handle_create_block(struct parser_pdata *pdata,
 				const struct iiod_command *cmd,
 				struct iiod_command_data *cmd_data)
 {
+	struct buffer_entry *buf_entry;
 	struct block_entry *entry;
 	struct iio_block *block;
 	struct iio_buffer *buf;
 	struct iiod_buf data;
 	uint64_t block_size;
 	struct iiod_io *io;
-	int ret;
+	int ret, ep_fd;
 
 	io = iiod_command_create_io(cmd, cmd_data);
 	ret = iio_err(io);
@@ -651,7 +667,7 @@ static void handle_create_block(struct parser_pdata *pdata,
 	if (ret < 0)
 		goto out_send_response;
 
-	buf = get_iio_buffer(pdata, cmd, NULL);
+	buf = get_iio_buffer(pdata, cmd, &buf_entry);
 	ret = iio_err(buf);
 	if (ret)
 		goto out_send_response;
@@ -681,6 +697,24 @@ static void handle_create_block(struct parser_pdata *pdata,
 	entry->io = io;
 	entry->client_id = cmd->client_id;
 
+	if (WITH_IIOD_USB_DMABUF && pdata->is_usb) {
+		entry->dmabuf_fd = iio_block_get_dmabuf_fd(block);
+		if (entry->dmabuf_fd > 0) {
+			ep_fd = buf_entry->is_tx ? pdata->fd_in : pdata->fd_out;
+
+			ret = usb_attach_dmabuf(ep_fd, entry->dmabuf_fd);
+			if (!ret) {
+				/* We could attach to functionfs. Disable CPU
+				 * access to the block as we won't need it. */
+				iio_block_disable_cpu_access(block, true);
+			} else {
+				/* If we can't attach - no problem. The
+				 * data will be transferred the regular way. */
+				entry->dmabuf_fd = -ENOSYS;
+			}
+		}
+	}
+
 	/* Keep a reference to the iiod_io until the block is freed. */
 	iiod_io_ref(io);
 
@@ -697,10 +731,17 @@ static void handle_free_block(struct parser_pdata *pdata,
 			      const struct iiod_command *cmd,
 			      struct iiod_command_data *cmd_data)
 {
+	struct buffer_entry *buf_entry;
 	struct block_entry *entry;
 	struct iio_block *block;
+	struct iio_buffer *buf;
 	struct iiod_io *io;
-	int ret;
+	int ret, ep_fd;
+
+	buf = get_iio_buffer(pdata, cmd, &buf_entry);
+	ret = iio_err(buf);
+	if (ret)
+		goto out_send_response;
 
 	block = get_iio_block(pdata, cmd, NULL);
 	ret = iio_err(block);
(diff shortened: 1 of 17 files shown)


pcercuei:
To be tested with a ZedBoard or ZCU102 running Linux 6.1 from this branch:
https://github.com/analogdevicesinc/linux/tree/pcercuei/dmabuf-import-v6.1

With the following caveats:
- For now, killing iio_rwdev (or in general calling `iio_buffer_cancel`) when running with the USB interface might cause IIOD to hang. But this bug is also present in the dev branch.
- Cyclic buffers are not supported for now. The AXI DMA driver need to be updated to support the cyclic mode on scatterlists.
- Partial transfers are not supported either, for now. The DMA API function `dmaengine_prep_slave_sg` will always transfer the full scatterlist. So we'll need either a new API function which also takes a size as argument, or a mechanism to "cut" the scatterlist to the right size.
- The AXI DMA driver and IP currently don't support scatter-gather, so the amount of IRQs generated might be excessive - as it may issue an IRQ after each memory page. However it is not uncommon to have the DMABUFs created on physically contiguous memory, and in this case can be transferred entirely in one go.

iiod/responder.c
```diff
@@ -681,6 +697,24 @@ static void handle_create_block(struct parser_pdata *pdata,
...
+				/* We could attach to functionfs. Disable CPU
+				 * access to the block as we won't need it. */
+				iio_block_disable_cpu_access(block, true);
+			} else {
+				/* If we can't attach - no problem. The
+				 * data will be transferred the regular way. */
+				entry->dmabuf_fd = -EINVAL;
+			}
+		}
+	}
```
nunojsa, line 716: nit: maybe we should already have a mechanism in here to properly handle this in the proper backend? For now, yes, only USB supports this but ideally we can have the same in net.

Maybe a simple helper function or a new callback? 

Not really a blocker from my point of but maybe something easy enough that could be done already instead of doing it when a new backend supports this mechanism

iiod/responder.c
```diff
@@ -681,6 +697,24 @@ static void handle_create_block(struct parser_pdata *pdata,
...
+
+			ret = usb_attach_dmabuf(ep_fd, entry->dmabuf_fd);
+			if (!ret) {
+				/* We could attach to functionfs. Disable CPU
+				 * access to the block as we won't need it. */
+				iio_block_disable_cpu_access(block, true);
+			} else {
+				/* If we can't attach - no problem. The
+				 * data will be transferred the regular way. */
+				entry->dmabuf_fd = -EINVAL;
```
nunojsa, line 713: hmm, can't we use a better error code in here? "Invalid argument" is not really the same as "Functionality not supported.". Is there any subtleties that enforces other error code than "ENOSYS" (seems to be the proper fit)?

iiod/responder.c
```diff
@@ -715,6 +756,10 @@ static void handle_free_block(struct parser_pdata *pdata,
 		if (entry->block != block)
 			continue;
 
+		ep_fd = buf_entry->is_tx ? pdata->fd_in : pdata->fd_out;
+		if (WITH_IIOD_USB_DMABUF && entry->dmabuf_fd > 0)
+			usb_detach_dmabuf(ep_fd, entry->dmabuf_fd);
```
nunojsa, line 761: nit: another place where we could already have some "backend agnostic" mechanism in place...

iiod/usb-dmabuf.c
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * libiio - Library for interfacing industrial I/O (IIO) devices
+ *
+ * Copyright (C) 2022 Analog Devices, Inc.
```
nunojsa, line 5: 2023? not sure if what matters is the time when the file was created or when it's merged into the project...

iiod/usb-dmabuf.c
```diff
@@ -0,0 +1,60 @@
...
+#include <stdint.h>
+#include <sys/ioctl.h>
+
+#define IIO_FFS_DMABUF_ATTACH	_IOW('g', 131, int)
+#define IIO_FFS_DMABUF_DETACH	_IOW('g', 132, int)
+#define IIO_FFS_DMABUF_TRANSFER	_IOW('g', 133, struct iio_ffs_dmabuf_transfer)
+
+struct iio_ffs_dmabuf_transfer {
+	int fd;
+	uint32_t flags;
```
nunojsa, line 20: maybe let's make this u64 already? I know the kernel side has u32 (which would have to be also updated) and I was actually wondering why not `unsigned long` but now I see this is used in ioctl().

BTW, the only reason I'm stating to move this into u64 is just to be on the safe side and have more space for flags (we never know how this will grow even though now, it seems that 32 flags is more than enough)

nunojsa (commented):
Cool work... Mostly nitpicks from my side!

local-dmabuf.c
```diff
@@ -45,30 +55,74 @@ struct dma_buf_sync {
...
 
-	/* If we get -ENODEV or -EINVAL errors here, the ioctl is wrong and the
-	 * high-speed DMABUF interface is not supported. */
-	if (ret == -ENODEV || ret == -EINVAL || ret == -ENOTTY)
-		ret = -ENOSYS;
-	if (ret < 0)
 		goto err_free_priv;
+	}
+
+	memfd = syscall(__NR_memfd_create, "/libiio-udmabuf", MFD_ALLOW_SEALING);
```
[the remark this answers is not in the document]
pcercuei: It is present in the libc I am using, but it is not everywhere; doing it like that is safer.

iiod/responder.c
```diff
@@ -341,7 +342,17 @@ static int buffer_dequeue_block(void *priv, void *d)
 	if (ret < 0)
 		goto out_send_response;
 
-	if (!iio_buffer_is_tx(buffer->buf)) {
+	if (!buffer->is_tx) {
+		if (WITH_IIOD_USB_DMABUF && entry->dmabuf_fd > 0) {
+			/* We need to send the error code before the data.
+			 * If usb_transfer_dmabuf() fails, we're screwed... */
+			iiod_io_send_response_code(entry->io, entry->bytes_used);
```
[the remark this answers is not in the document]
pcercuei: The client needs to know how many bytes to read (the response code is either a negative error or a number of bytes); so we need to send it before the data.

iiod/responder.c
pcercuei: USB does not support DMABUFs, the gadget/functionfs API does. So this is very much a IIOD-side things that shouldn't be in Libiio.

If we ever get DMABUF support to the network stack, the related code could be shared (as a static lib akin to iiod-responder) between the network backend and IIOD.

iiod/responder.c
pcercuei: The error code isn't really used in itself, any negative value will be treated as "zero-copy not possible". So I can change it to ENOSYS, it doesn't really matter.

iiod/responder.c
pcercuei: But this is IIOD and not Libiio, and interfaces gadget/functionfs and not client-side USB, so it wouldn't fit in one of Libiio's backends.

iiod/usb-dmabuf.c
pcercuei: Well I currently have a total of 0 flags, I would think that 32 available flags is more than enough :)

I could switch to u64 flags, at the cost of doubling the size of the structure (24 bytes vs. 12 bytes), but I'm not really sure that will ever be needed.

I will send a patchset for the Linux bits very soon, let's see if they say anything about that.

local-dmabuf.c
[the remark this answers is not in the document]
nunojsa: Hmm, then I would wrap this is in a macro and call the libc version when available and fallback to this only when necessary...

iiod/responder.c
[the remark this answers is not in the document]
nunojsa: I see...

iiod/responder.c
nunojsa: > So this is very much a IIOD-side things that shouldn't be in Libiio.

But don't we have a similar situation if network is being used? IIOD still needs to "respond" the data right?

iiod/responder.c
nunojsa: I guess at the very least matters in terms for reading the code... But if that's the case, stating that in the comment would be enough

iiod/responder.c
nunojsa: I'm aware it's IIOD, but for example if we ever get support for DMABUF in network (I know this is a big 'if') would we have something like `net_detach_dmabuf()`? So wouldn't we need to have another another:

```
if (WITH_IIOD_NET_DMABUF && ...)
   ...
```

But maybe there's no `ops` kind of thing in IIOD and adding new helpers for this, yeah, can be done when really needed... 

Or maybe I'm just completely off here and should just check iiod code :sweat_smile:

pcercuei: Yes, there are no "frontends" in IIOD yet, and adding this separation would be a bit outside the scope of this PR. I would refactor this the moment where we have DMABUF support on the network stack.

local-dmabuf.c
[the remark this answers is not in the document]
pcercuei: I could do that, but `memfd_create` is just a wrapper around syscall( ) anyway, I don't really see what would be the benefit.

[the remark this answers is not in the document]
nunojsa: > I could do that, but memfd_create is just a wrapper around syscall( ) anyway, I don't really see what would be the benefit.

Well, for now... Like this, you can potentially bypass any new code that lands in libc. 

Anyways, up to you :)

iiod/responder.c
pcercuei: Right. We could have a "frontend" system, but I'd prefer refactoring the code once we do have DMABUF support in the network stack.

iiod/responder.c
```diff
@@ -445,6 +458,8 @@ static void handle_create_buffer(struct parser_pdata *pdata,
 			entry->words[BIT_WORD(i)] &= ~BIT_MASK(i);
 	}
 
+	entry->is_tx = iio_buffer_is_tx(buf);
```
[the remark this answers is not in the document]
pcercuei: It is used in the same commit, though. The value is just cached to avoid calling `iio_buffer_is_tx` every time.

I guess this could be cached inside the `iio_buffer` structure instead, yeah.

PR #931, opened 2023-01-30
serial: Properly flush the input buffer

pcercuei:
Calling sp_flush() does not seem to be enough to flush the input buffer with some serial-to-USB adapters.


buha (approved):
tested this with max78000-fthr running tinyiiod, works well

PR #932, opened 2023-01-30
iiod-client: Restore support for tinyiiod based programs

pcercuei:
Tinyiiod may not handle the TIMEOUT command. This shouldn't prevent us from creating a IIO context, so in the case where we get a -22 error (aka. command does not exist), continue the context creation as usual.


buha:
yep, this is the good approach

PR #934, opened 2023-02-15
channel: Support attribute names that start with the channel's label

pcercuei:
Support attribute names that start with the channel's label to avoid breaking compatibility with old kernels, which did not offer a 'label' attribute, and caused Libiio to sometimes misdetect the channel's extended name as being part of the attribute name.

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -344,15 +344,43 @@ const char * iio_channel_get_attr(const struct iio_channel *chn,
 		return chn->attrs[index].name;
 }
 
-const char * iio_channel_find_attr(const struct iio_channel *chn,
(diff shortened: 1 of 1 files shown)


nunojsa (approved):

nunojsa:
Ahh, one thing.... Shouldn't we also have this in master?

pcercuei:
At this point it doesn't really matter, the dev branch will be merged back to master as soon as #928 is merged.

pcercuei:
Ignoring the codacy error, as we know for a fact that `channel->name` when set will always be NULL-terminated.

PR #935, opened 2023-02-15
cmake: Install: add PUBLIC_HEADER DESTINATION

nunojsa:
Without adding a DESTINATION in the iio target, cmake complains with the following warning:

"
CMake Warning (dev) at cmake/Install.cmake:38 (install):
  Target iio has PUBLIC_HEADER files but no PUBLIC_HEADER DESTINATION.
  CMakeLists.txt:593 (include)
This warning is for project developers.  Use -Wno-dev to suppress it. "


nunojsa:
v2:

* Removed unneeded `install()` calls...

pcercuei:
This changes how things are done on OSX, which is fine, but the commit message should say something about that.

nunojsa:
v3:

* better commit message.

pcercuei (approved):

Issue #925, opened 2023-01-11
iio_attr -a mode fails on PlutoSDR


tfcollins:
```
Welcome to:
______ _       _        _________________
| ___ \ |     | |      /  ___|  _  \ ___ \
| |_/ / |_   _| |_ ___ \ `--.| | | | |_/ /
|  __/| | | | | __/ _ \ `--. \ | | |    /
| |   | | |_| | || (_) /\__/ / |/ /| |\ \
\_|   |_|\__,_|\__\___/\____/|___/ \_| \_|

v0.34-3-g401e-dirty
https://wiki.analog.com/university/tools/pluto
# iio_attr -a -d
Scanning for IIO contexts failed: Input/output error (5)
```
I guess its missing some discovery/scanning packages?

pcercuei:
It shouldn't need any. Could you try to see where the -EIO error is returned from?

tfcollins:
Ok. Let me get my cross-compiling setup going

rgetz:
Pluto doesn't ship with avahi client on it. This failure is expected.

I don't think -a (which does a scan, and attaches to the first device) tries local. That's a feature enhancement.

-Robin

pcercuei:
@rgetz In that case I would expect it to fail, but not with a -EIO code.

rgetz:
I was wrong - it should try local...

`LIBIIO_SCAN_BACKENDS` should have `local` defined when compiled with `WITH_LOCAL_BACKEND`

a scan returns the same.
```
# iio_info -s
Library version: 0.23 (git tag: v0.23)
Compiled with backends: local xml ip usb serial
Scanning for IIO contexts failed: Input/output error (5)
```

so - scans fail on local for some reason.

from a host:
```
C:\Users\rgetz>iio_info -s
Library version: 0.24 (git tag: c4498c2)
Compiled with backends: xml ip usb serial
Unable to create Local IIO context : Function not implemented (40)
Available contexts:
        0: 192.168.2.1 (Analog Devices PlutoSDR Rev.B (Z7010-AD9364)), serial=1044735411960002e7ff210031f11e3d48 [ip:pluto.local]
        1: 0456:b673 (Analog Devices Inc. PlutoSDR (ADALM-PLUTO)), serial=1044735411960002e7ff210031f11e3d48 [usb:3.1.5]
C:\Users\rgetz>
```

I can try on a ARM target later (using Kuiper).

-Robin

rgetz:
[...]

PR #939, opened 2023-02-20
usb: don't error during a scan on linux when there are no usb hosts.

rgetz:
libiio scans via libusb when the usb backend is built in. For many embedded linux targets that support On The Go (OTG) the USB port can be configured as device or host. Sometimes, end users may want to scan for local devices when the OTG is configured as a device (and there aren't any USB hosts).

The issue with this is libusb returns an error when there are not any hosts. So - on linux - catch that, and override the error handling.

New functionality is when Linux's OTG USB is in device mode, usb scans can still take place.

This fixes issue #925 (iio_attr -a fails on plutoSDR).

### DIFF
--- a/usb.c
+++ b/usb.c
@@ -11,6 +11,9 @@
 #include "iiod-client.h"
 
 #include <ctype.h>
+#ifdef __linux__
+#include <dirent.h>
+#endif
 #include <errno.h>
 #include <libusb.h>
 #include <stdbool.h>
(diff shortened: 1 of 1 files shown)


usb.c
```diff
@@ -1253,8 +1258,20 @@ int usb_context_scan(struct iio_scan_result *scan_result, const char *args)
...
+#ifdef __linux__
+		/* When Linux's OTG USB is in device mode, and there are no hosts,
+		 * libusb_init() is expected to fail, but we shouldn't treat that
+		 * as a hard failure - only that there are no devices.
+		 */
+		DIR* dir = opendir("/dev/bus/usb/");
+		if (dir)
+			closedir(dir);
+		else if (errno == ENOENT)
+			ret = 0;
```
pcercuei, line 1271: You could `return 0` here, then you don't need to add a `LIBUSB_SUCCESS` (which shouldn't really be an "error" code IMHO).

pcercuei (changes requested):

usb.c
rgetz: Updated per your request...

pcercuei (approved):

rgetz:
Thanks

PR #942, opened 2023-02-25
Rgetz move iio monitor

rgetz:
This moves iio-monitor to the test directory, (since it's built as a test file today) and gives it some help and common command line arguments.


pcercuei:
Honestly, I'd just prevent it from being installed. The iio-monitor is a toy program which just acts as an example of how to use the API, so it does belong in "examples" IMHO.

rgetz:
In the existing cmake - it gets installed in the debs - I can just remove it if that is the conclusion.

I use to show it to lots of people... (I always thought it was cool).

pcercuei:
@rgetz it is cool but doesn't really provide any functionality unlike iio_attr / iio_info / iio_rwdev etc., it is more an experiment.

rgetz:
Sure - I will nack this one then, and do something different than just unwinds the examples and tests.

PR #944, opened 2023-02-26
Rgetz update man pages

rgetz:
This uses a slightly different solution than #886, (which checks in auto generated files), but adds some cmake to actually just generate the files on the build machine, and the time of building.

assumption is that all utils generate help (with #942 ; they do), and that the host has access to 'help2man'.

-Robin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,8 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND OSX_FRAMEWORK)
 	set(SKIP_INSTALL_ALL ${OSX_PACKAGE})
 endif()
 
+option(WITH_MAN "Generate on-line reference manuals (man pages)" OFF)
+include(cmake/CheckManPages.cmake)
 option(WITH_NETWORK_BACKEND "Enable the network backend" ON)
 option(WITH_TESTS "Build the test programs" ON)
 option(WITH_EXAMPLES "Build examples" OFF)
@@ -551,8 +553,9 @@ option(CSHARP_BINDINGS "Install C# bindings" OFF)
 option(PYTHON_BINDINGS "Install Python bindings" OFF)
 add_subdirectory(bindings)
 
-option(WITH_MAN "Generate on-line reference manuals (man pages)" OFF)
-add_subdirectory(man)
+if(WITH_MAN)
+	add_subdirectory(man)
+endif()
 
 # Create an installer if compiling for OSX
 if(OSX_PACKAGE AND OSX_FRAMEWORK)
--- a/cmake/CheckManPages.cmake
+++ b/cmake/CheckManPages.cmake
@@ -1,0 +1,31 @@
+# check to make sure man pages are there, otherwise build them
+
+macro( CheckManPages FILE_LIST )
+	if(WITH_MAN AND NOT CMAKE_CROSSCOMPILING)
+		foreach(UTIL IN ITEMS ${FILE_LIST})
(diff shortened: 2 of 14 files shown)


rgetz:
update to fix some quoting issues in shell script.

pcercuei:
I think the first commit is OK, my main gripe with it is that it touches CMake code... which means it won't merge cleanly to the "dev" branch. I'd prefer if you could make it work in the "dev" branch instead.

For the second commit... I don't really see the point of having CMake auto-generate documentation, why not provide these auto-generated (or hand-written) docs instead?

rgetz:
I can rebase on dev.

Checking in generated files has always been poor form for me - if someone updates the operation and then updates the help - they need to update the generated code.

It's less work to just let Cmake do it for you long term. 

And yes - they should all have comprehensive hand written man pages. Right now - they don't.

pcercuei:
@rgetz I would agree if it generated files that can be used verbatim without touch-ups, but I think some files need to be fixed up a bit after generation, no?

rgetz:
I thought help2man did a pretty well, and was suggesting to use it as is.

I thought that in absence of a well written (manual) man page - this would suffice since many distributions now require man pages for every executable  (which is why the script only calls help to man if a hand maintained one doesn't exist).

pcercuei:
@rgetz Fine with me, as long as the generated man pages are good enough to be used in production.

rgetz:
yeah, I thought they were OK (a little minimal), but still OK.

pcercuei:
I'm still not convinced by this one. Calling help2man means that the build machine must be able to execute the programs that were built. I'm not sure that works e.g. on MinGW, and definitely won't work when cross-compiling.

I still think it's better to have pre-generated man files. We could add a CI pass that would build libiio, call `help2man` then compare with the pre-generated ones, and fail if they don't match.

rgetz:
Yes - you are correct - using help2man assumes that you have a "working" binary. (either native or emulated via QEMU or something - which is pretty common now, and easy to set up in CI). but - you are correct - it should be protected from doing that by `CMAKE_CROSSCOMPILING` or something. (but cross compiling, and wanting man pages? - I'm not sure I see the use case).

You can make the same argument for lex/yacc/bison - why don't we check in parser.c (and other) files? and not have people install those tools, and only compare in a CI job... (To me - it's because they aren't the "truth", they are a derived work, and we want to make sure that people don't start modifying the derived work, and the "truth" is always the "truth").

-Robin

pcercuei:
Not sure what you mean with the "truth".

help2man itself is not the problem; it's that you have to execute the iio_info, iio_rwdev etc. executable in order to get its --help print. You don't need to do that with flex/bison.

rgetz:
I mean the "truth" is the "source".

installing qemu on most modern distributions to run an ARM binary on x86 is a pretty trivial task for most. (it's large, but it's easy).

I can reformat this into two patches, one for CI, and a different one for the manually generated files.

pcercuei:
@rgetz can you add a commit that adds the pre-generated missing man pages, plus a commit that adds a command to generate man pages like you have before, but that is not called automatically in the build process (you'd have to run `make gen-man` or something like that).

Then we can bother about the CI later.

rgetz:
based on the phone call we had earlier today - did you still want me to do this?

From the call - I got the impression that the first part of this was OK, but to drop the 2nd completely. Correct? (you volunteered to just hand write the files?)

pcercuei:
I will close this PR and send a new one with your first commits included, as well as the new man pages.

PR #945, opened 2023-03-12
cmake: ensure when someone turns something off, it prints out

rgetz:
normally when you turn something off during the cmake, you invoke it like:
cmake ../ -DWITH_USB_BACKEND=0

it shows up internally as prefixed with a _0, not a _OFF. Today we only print _OFF, not _0, so we silently eat things without letting people know (which is bad form).

This patch fixes that, by making sure _1 and _ON; and _0 and _OFF are printed out. This should make sure people know what they are building.


pcercuei:
I think you could use CMake generators to sanitize the options.
For instance:

Before:
```list(APPEND IIOB_FEATURES_${CSHARP_BINDINGS} "c#-bindings")```

After:
```list(APPEND IIOB_FEATURES_$<BOOL:${CSHARP_BINDINGS}> "c#-bindings")```

This will append the tring to `IIOB_FEATURES_0` if `CSHARP_BINDINGS` is 0, OFF, NO, NOTFOUND.

This is usable starting from CMake 3.0, I think.

**Comment by rgetz** [comment]
It seems to be in the 2.8.10 doc: https://cmake.org/cmake/help/v2.8.10/cmake.html

```
  $<BOOL:...>               = '1' if the '...' is true, else '0'
```

so, I agree that's better - and will bump up the min cmake requirements to match that (from the existing 2.8.7)

rgetz:
Ok - that won't work - BOOL is part of the generator expressions, and cmake doesn't mix it's configuration and generation phases (for all the right reasons). Generator expressions are only evaluated at generation time, not at configure time. The message() command is not able to evaluate them because it tries to do that at configure time. 

https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

To do that - would require something like:
https://jeremimucha.com/2021/03/cmake-generator-expressions/

which I'm afraid would get nacked...

Doing something with some macros looking for the same thing bool does , check if:
 * string is empty,
 * string is a case-insensitive equal of 0, FALSE, OFF, N, NO, IGNORE, or NOTFOUND, or
 * string ends in the suffix -NOTFOUND (case-sensitive).

if yes, "0", if not "1".

-Robin

PR #946, opened 2023-03-13
CI: remove builds with deprecated version

ccraluca:
Remove builds for Ubuntu 18.04 which is deprecated from Azure.


rgetz:
Ok - to be clear - https://ubuntu.com/about/release-cycle says that 18.04 LTS (Bionic Beaver) is in the "Extended Security Maintenance (ESM)" phase of it's lifetime - it's not officially dead until 2028. (but you should be able to get it from older libiio releases)...

However - what ESM really means - https://ubuntu.com/security/esm - is if you have a CVE - you should fix it for the old release that worked on 18.04 LTS. (which I think was 0.10 https://packages.ubuntu.com/bionic/libiio-dev ; which I don't think anyone really maintains)...

Github actions seems to be tossing it...
https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

The reason it still runs is because we are using the docker that @tfcollins maintains, not the github one (so it has all the prerequisites installed), and you don't need to build/install libusb (for example). It will continue to run until some incompatibility comes up.

tfcollins:
Handling 2 LTS versions seems enough so we just decided to move away from 18.04 across the board. Docker images will be dropped from our Docker builds in April

pcercuei:
So the CI will start to fail in April then. I guess that means we can merge this now.

pcercuei (approved):

PR #951, opened 2023-04-21
Distribution.xml:Make libiio.pkg installable without using Rosetta on M1/M2

AlexandraTrifan:
On Apple Silicon, the installer will require Rosetta 2 when installing libiio unless the arch specifiers are added in the installer xml file.


pcercuei:
Can you shorten the title to `Distribution.xml: Make libiio.pkg installable without using Rosetta` (also with a space after the colon).
Commit titles should generally have <= 72 characters.

rgetz:
Is this only needed for universal binaries, or only ARM binaries?

AlexandraTrifan:
I also changed x86_64 to i386 based on the docs that state: 
`hostArchitectures | String | Optional. A comma-separated list of supported architecture codes. Valid codes are i386, x86_64, and ppc. Note that i386 matches both 32-bit and 64-bit systems, but x86_64 matches only 64-bit systems.`

> Is this only needed for universal binaries, or only ARM binaries?

This is a needed update in order to be able to build the pkg on Apple Silicon and then install it without requiring Rosetta 2. This does not affect the Intel build.
From all the docs I found, this is something that gets added automatically when building a package for macOS 11.0 or newer. 
`NOTE: Starting on macOS 11.0 (Big Sur), productbuild will automatically specify support for both arm64 and x86_64 unless a custom value for arch is provided.`
`When you use productbuild to create a distribution pkg on Big Sur (Intel and Apple silicon) both arm64 and x86_64 will be added to the configuration by default.`

As for universal packages..from what I understand, those should contain the binaries for both versions and run the appropriate one on the host machine? I was able to try this by specifying "CMAKE_OSX_ARCHITECTURES=arm64;x86_64" but it failed when libusb had a different arch than the one currently built.
So there should be 2 separate binaries, one for each supported architecture, right?

rgetz:
Yes - I think two unique packages will be easier to manage (one for x86_64 and one for ARM64).

pcercuei (approved):

PR #953, opened 2023-04-24
tests: don't dereference a var before checking it.

rgetz:
The strchr() return a NULL if the character is not found. Make sure to handle that.

### DIFF
--- a/tests/gen_code.c
+++ b/tests/gen_code.c
@@ -51,7 +51,7 @@ bool gen_test_path(const char *gen_file)
 		return false;
 
 	last = strrchr(gen_file, '.');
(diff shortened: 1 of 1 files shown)


nunojsa (approved):

pcercuei:
Actually it should probably just be `if (!last)`, no? I don't think strchr / strrchr can return any thing else than a pointer to the character, or NULL.

nunojsa:
> Actually it should probably just be if (!last), no? I don't think strchr / strrchr can return any thing else than a pointer to the character, or NULL.

Agreed

rgetz:
updated.

PR #954, opened 2023-04-25
set errno to 0, and check after calling library functions

rgetz:
A few times we are using strtol and strtoul, where we were not setting errno to zero, and then checking it afterwards - this could lead to undefined behaviour based on end user inputs, so do like the man page suggests.

### DIFF
--- a/network.c
+++ b/network.c
@@ -1126,8 +1126,10 @@ struct iio_context * network_create_context(const char *host)
 	}
 	if (port) {
 		unsigned long int tmp;
+
+		errno = 0;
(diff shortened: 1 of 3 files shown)


pcercuei:
My assumption was that on range error, we would get the "endptr" variable point to the beginning of the string. A quick local test showed that it's not the case, so I guess the check for ERANGE makes sense.

The check for EINVAL doesn't though, we always use either base 10 or base 16, which are always supported.

rgetz:
Can someone restart the checks - none of those failures are build issues.

PR #958, opened 2023-04-29
cmake: tell people if we are/are not install udev rules

rgetz:
we currently don't tell people if we are or aren't installing udev rules (which I think is an oversight). So do that.


pcercuei:
@rgetz and does it work? Because the CMake variable is called `INSTALL_UDEV_RULE` :)

rgetz:
well, it worked that it said it was off, but you are right - it never turned on... should be functional now.

pcercuei (approved):

PR #959, opened 2023-05-04
Add missing man pages

pcercuei:
Rework man pages generation, and add missing man pages.

@rgetz


nunojsa (approved):

rgetz:
minor nitpics - fix the ones you want to. Otherwise looks good to me.


`iio_reg is a utility for debugging local IIO devices. `
is no longer true - it accepts uri, and can be used with remote devices.

for the usage in `iio_readdev`, & `iio_writedev` you no longer need to use iio_attr to find buffers, iio_readdev / writedev will do that for you. Just call it with a uri, but no device.

for `iiod` :
 * `-p 0` will pick a Ephemeral port. (dynamic / unused in the range from 32768–60999).
 * it includes `-u` twice, and scan (which I don't think it supports)
 * might be good to say that default context is local, but others are possible to allow forwarding (forward a USB device over network or serial).

There is no description in the iio_stresstest page

Do you really want to ship iio_axi_xflow_check as part of iio utils? or should that be moved to "examples"? It's pretty custom to a specific vendor...???

pcercuei:
PR updated.

rgetz:
Just noticed that doc/index.html.in should be updated with the newly created man pages...

pcercuei:
Updated again.

rgetz:
Thanks - looks good to me.

Issue #694, opened 2021-06-02
cmake: add uninstall target


willcode:
An "uninstall" target is useful both for development and for source-based package management (PyBOMBS in our case).

While cmake does not have a convenient uninstall, it is typically implemented using code like:
https://gist.github.com/royvandam/3033428

rgetz:
Thanks - we will have a look - and try to add

rgetz:
There should be file "install_manifest.txt" after executing `make install`. Either of the following commands should work:

`cat install_manifest.txt | sudo xargs rm`

or 

``xargs sudo rm < install_manifest.txt``

Does that solve your issue?

willcode:
Hi @rgetz, sure that's fine for a manual uninstall. I put this in originally because PyBOMBS runs `make uninstall` explicitly. This issue has been up for quite a while, and I don't hear any complaints. Not that surpising since the PyBOMBS recipe calls for an ancient tag. Feel free to close it if adding `make uninstall` is not a priority.

The recipe has specified v0.21 for a while now. I'd guess that people using PyBOMBS and PLUTO are installing via a package manager.

rgetz:
When #961 gets merged - this can be closed.

rgetz:
this can be closed - it was included in 0.25

PR #961, opened 2023-05-06
add uninstall target

rgetz:
Per the request #694, this adds an "uninstall" target, which is (a) good form, and (b) requested by Jeff Long, and will make PyBOMBs work more natively.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -652,6 +652,18 @@ if (WITH_USB_BACKEND AND CMAKE_SYSTEM_NAME MATCHES "^Linux")
 	endif()
 endif()
 
+# Add uninstall target
+if(NOT TARGET uninstall)
+	configure_file(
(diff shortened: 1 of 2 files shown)


willcode:
Thanks!

pcercuei:
The license of the code you referenced in the gist is uncertain...

The code would also need to respect the `DESTDIR` environment variable.

willcode:
From SoapySDR (Boost Software License):
https://github.com/pothosware/SoapySDR/blob/master/cmake_uninstall.cmake.in

GNU Radio (GPLv3)
https://github.com/gnuradio/gnuradio/blob/main/cmake/cmake_uninstall.cmake.in

From the CMake wiki (presumable meant to be used under any license)
https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake

pcercuei (approved):

PR #965, opened 2023-05-07
Update Doxyfile.in files

rgetz:
Currently - the Doxfiles that are checked in, are from (old? nine years ago), and currently produces warnings when you build things:

warning: Tag 'TCL_SUBST' at line 168 of file './Doxyfile.in' has become obsolete.
         This tag has been removed.
warning: Tag 'COLS_IN_ALPHA_INDEX' at line 871 of file './Doxyfile.in' has become obsolete.
         This tag has been removed.
warning: Tag 'PERL_PATH' at line 1558 of file './Doxyfile.in' has become obsolete.
         This tag has been removed.
warning: Tag 'MSCGEN_PATH' at line 1580 of file './Doxyfile.in' has become obsolete.
         This tag has been removed.

Update to 1.9.1 via:
doxygen -u ./bindings/csharp/Doxyfile.in
doxygen -u ./Doxyfile.in

and check in the difference.
Unfortunately - this has quite a few comment differences :(
 2 files changed, 1885 insertions(+), 374 deletions(-)

but if you only look at the actual options that have been moved/updated, it's pretty minimal.

git diff | grep -v "\+\#" | grep -v "\-\#" | diffstat
 b/Doxyfile.in               |   10 +++++++++-
 bindings/csharp/Doxyfile.in |    8 ++++++++
 2 files changed, 17 insertions(+), 1 deletion(-)


pcercuei:
@rgetz does that still build on a very old system, e.g. Ubuntu 16.04?

rgetz:
I did not check something that far back - as I thought we didn't care when we dropped things from CI? 

If we are concerned about 16.04 why did we remove it?

pcercuei:
Well, I'm not sure. We do stick to a really old CMake to stay compatible with old systems, so it's probably better to stay compatible with old compilers and build tools like doxygen as well, at least while they work.

I would love your doxygen fix in the "dev" branch though.

rgetz:
The only reason we were sticking with ancient cmake - was that was what ADI was distributing on their SD Cards as part of the Linaro release - that was updated when ADI's Kuiper was released to cmake 3.13.4

Both Ubuntu 16.04 LTS (Xenial) was end of lifed on April 30, 2021; Ubuntu 18.04 LTS (Bionic): end of life on May 31, 2023. Both are now in their "Extended Security Maintenance (ESM)" release - people using that shouldn't expect new packages to build/install on it (IMHO).

-Robin

pcercuei:
Fair enough.

PR #969, opened 2023-05-11
CI: add required builds

ccraluca:
Following OSs was added to the CI:
- Linux-Ubuntu-18.04 (since it was required and we still have a docker image for it)
- Linux-Fedora-28
- Linux-Debian-11
- Linux-openSUSE-15.4
- Linux-CentOS-7
- macOS-13


pcercuei:
@ccraluca, can you just squash the two commits together, and keep both signed-off-by.

pcercuei:
Github also says that there is a conflict on `azure-pipelines.yml`, could you rebase it on top of latest master?

tfcollins:
Can we get the Windows zips in here as well?

ccraluca:
@tfcollins, by "here" do you mean publish them to Azure pipeline artifacts for the master and staging/* branches?

tfcollins:
I mean as part of this PR

rgetz:
I don't think all the tar files include the required libraries- do they?

It seems only the windows zip file includes libusb, libxml, etc...

rgetz:
yes - and what I'm saying is the old tar balls are not what we want to provide; it doesn't help as much as when the prerequisite libraries are included.

I have something as part of #971 that builds up a .so manifest (the *.so's that are expected to be installed for the tarballs to work); that could be added to things as a custom target - which could be run before the final "make package" for the tar balls.

rgetz:
I saying - we need them, we just need them packaged with the required libraries, like the windows zip files.

pcercuei:
@rgetz what do we need those for?

rgetz:
That is what MathWorks uses to install as part of their Pluto (and other IIO) support packages. 

We don't ask for root privileges to install anything on the system path (other than the udev rule), so all the libraries just goes in the normal user path via the tarball.

rgetz:
I added something to #971 which adds the required libraries to the tarball (for debian). I just need to do the same for fedora.

PR #970, opened 2023-05-12
C++ API

sternmull:
This adds a C++17 API to the library. Implemented as a single header without dependencies. It simplifies usage of the library when a modern C++ compiler is available.


pcercuei:
Hi @sternmull,

This looks great and we definitely appreciate the contribution!

I have a few remarks on the commit itself.

First of all - please sign your commit (with `git commit --amend -s`). It would be also great to have a better commit message.
Nothing fancy, your description in the PR's message is good enough:
```
Add C++ API

This adds a C++17 API to the library. Implemented as a single header without dependencies. It simplifies usage of the library when a modern C++ compiler is available.
```
(Note that the commit's description should be limited to 72 chars per line)

As for the new C++ API: it arrives at an awkward time where we are rushing to tag v0.25 as the very last v0.x release before switching to the next-gen libiio (currently in the "dev" branch). As soon as we merge the "dev" branch back to master, the C++ bindings will be outdated unfortunately, and would have to be updated.

I wanted to tag v0.25 today but I think it'd be better to merge your C++ API first - so that the last v0.x release has it, at least.

About the commit's content: could you move `iiopp.h` to `bindings/cpp/`, and add a `CPP_BINDINGS` CMake option (defaulting to OFF)? Since it is just a header, the option would just enable installing the .hpp.

I would also move the example to `bindings/cpp/examples` (to mirror what's done with Python). Also, please error-check the context creation.

Overall looks fantastic :)

rgetz:
@sternmull : Thanks

I don't know if you can get to: https://app.codacy.com/gh/analogdevicesinc/libiio/pullRequest?prid=11755628

I think most are false positives - but you should check them.

pcercuei:
@sternmull I meant to catch the context creation error and print something accordingly in the C++ example, yes. While examples don't have to be bullet-proof, I'd prefer to have this one error case covered - because it is likely to happen (e.g. when compiling host tools on PC, without IIO devices) and the exception messages can be quite cryptic.

sternmull:
@pcercuei All your concerns should be fixed.

@rgetz I am sure it was false alarm. I found a way to disable the issues by adding comments to ignore the respective lines.

rgetz:
I'm not sure I completely understand the changes to the Doxygen.in file?

sternmull:
@rgetz The Doxygen file had to be adjusted because the documentation of the C++ API extends the documentation of the C API (which makes sense because it needs to refer to the C datatypes), but only if `CPP_BINDINGS` is activated. The conditional stuff is handled in the toplevel `CMakeLists.txt` and a bunch of variables is set that are used the `Doxygen.in` to produce the final `Doxygen` configuration file. The changes have the following effects:

1. `STRIP_FROM_PATH` is set to the `bindings/cpp`  directory so it does not appear in the documentation as prefix. The reader is not interested where the library was built, only where the installed headers are located.
2. `ENABLED_SECTIONS` allows Doxygen-comments to use `@if ...  @endif` and `@cond ... @endcond` for conditional stuff. This is used to activate the few places where the C++ API is mentioned in the documentation (if it is activated).
3. `INPUT` no longer only uses just the source directory but uses a variable that conditionally also add `bindings/cpp`.
4. The directory with C++ examples is added to `EXAMPLE_PATH` (if it is activated).
5. `PREDEFINED` is always set to `DOXYGEN`. This define allows to use `#if ... #endif` to handle the Doxygen pass in a special way. I use this in `iiopp.h` to simplify the parameter lists of some class templates because the real code contains details that are irrelevant to a user of the library and would only make the documentation unnecessarily painful to read.

All this is done to make sure the documentation makes sense with and without the optional C++ bindings.

rgetz:
Thanks for explanation. That makes sense.

rgetz:
@sternmull ; like @pcercuei indicated:
>As for the new C++ API: it arrives at an awkward time where we are rushing to tag v0.25 as the very last v0.x 
>release before switching to the next-gen libiio (currently in the "dev" branch). As soon as we merge the "dev"
>branch back to master, the C++ bindings will be outdated unfortunately, and would have to be updated.

I just want to understand what you are willing to do (I don't know if this was a contract thing (drop things off and run away), or you did for fun, or other project - and you want to help more - BTW - either is OK, as long as we know) - since the C++17 expertise on the libiio team is low - can you comment on adding things to the dev branch? 

To get an idea of what libiio 1.x will look like, check out:
- slides : https://events.gnuradio.org/event/18/contributions/242/attachments/101/207/grcon2022.pdf
- video : https://youtu.be/hb3Au_cYfuQ?t=19670

The API does change somewhat, but does have a compatibility layer. I want to make sure that things are designed properly for dev (future looking) as well as for today's 0.25 branch (master) and understand who can help with the work. If it is all on the existing team - we just need to understand that.

-Robin

sternmull:
@rgetz @pcercuei About my motivation: At DIFITEC GmbH we are using libiio for some projects. I found its C-API to be a little inconvenient and started some C++ helpers. With that in place it just seemed natural to wrap the whole API and to make it available to the public. I can help to maintain the C++ API for the current and new API. I took a look at the dev branch. There are a lot of changes but the general picture did not change drastically and I see no problem in wrapping the new API in a very similar way as I did for the old one.

Is the new API still open for suggestions? At a first glance I would recommend:

Introduce a `iio_attrs` type that provides a unified way to access all the context/device/device_buffer/device_debug/channel-attributes. This would significantly reduce the number of functions in the API and also make it explicit that all the attributes use the same concept. This would look like this:

```
struct iio_attrs;

// one attrs-accessor for every atrribute-list
struct iio_attrs * iio_context_get_attrs(struct iio_context * ctx);
struct iio_attrs * iio_device_get_attrs(struct iio_device * dev);
struct iio_attrs * iio_device_get_buffer_attrs(struct iio_device * dev);
struct iio_attrs * iio_device_get_debug_attrs(struct iio_device * dev);
struct iio_attrs * iio_channel_get_attrs(struct iio_channel * chan);

// unified API to access an attribute-list (instead of having this set of functions for every specific attribute-list of the objects)
size_t iio_attrs_count(const struct iio_attrs * attrs);
ssize_t iio_attrs_find(const struct iio_attrs * attrs, char const * name);
char const * iio_attrs_name(const struct iio_attrs * attrs, size_t idx);
int iio_attrs_read_bool(const struct iio_attrs * attr, size_t idx, bool * out);
int iio_attrs_write_bool(struct iio_attrs * attr, size_t idx, bool val);
int iio_attrs_read_longlong(const struct iio_attrs * attr, size_t idx, long long * out);
int iio_attrs_write_longlong(struct iio_attrs * attr, size_t idx, long long val);
int iio_attrs_read_double(const struct iio_attrs * attr, size_t idx, double * out);
int iio_attrs_write_double(struct iio_attrs * attr, size_t idx, double val);

int iio_attrs_get_raw_size(const struct iio_attrs * attr, size_t idx, size_t * size); // I think this one would be useful
int iio_attrs_read_raw(const struct iio_attrs * attr, size_t idx, void * out, size_t size);
int iio_attrs_write_raw(struct iio_attrs * attr, size_t idx, void const * val, size_t size);
```
Another thing is the error handling. To me the "pointer-encoded error" concept feels very unusual (Is it safe to assume that the lower addresses are unused on all platforms?). I think it should be replaced by returning `NULL` and setting an (possibly thread local) error state (at least an error code) in the library that can be queried afterwards. It would also be nice to use the same error-checking/handling concept for all functions.

I don't know if it is too late for such changes or if the current design is motivated by good reasons that I am not aware of. Wrapping the new API for C++ would be possible any way. But if you are open for suggestions then I could offer help. There might also be some other things that come up when I take a closer look at the new API. Let me know if you are interested.

pcercuei:
Hi @sternmull,

About changing how attributes are handled:

That's something that I could do, I just didn't want to change too much of the API to make it easier for people to update and adapt to the upcoming version. But I'll do some tests and see if I like it better.

However I would use a slightly different approach:
```
size_t iio_context_get_attrs_count(const struct iio_context *ctx);
size_t iio_device_get_attrs_count(const struct iio_device *dev);
const struct iio_attr * iio_context_get_attr(const struct iio_context *ctx, unsigned int idx);
const struct iio_attr * iio_device_get_attr(const struct iio_device *dev, unsigned int idx);
...

const char * iio_attr_name(const struct iio_attr *attr);
int iio_attr_read_bool(const struct iio_attr *attr, bool *out);
int iio_attr_write_double(const struct iio_attr *attr, double val);
...
```

While both your solution and mine reduce the number of API functions, I'm not convinced that it simplifies the API - because it adds one level of indirection, as you need to retrieve the iio_attr(s) pointer first.

About pointer-encoded errors: this is just the C variant of your typical `std::optional<>` C++ construct, or Rust's `std::Option<>`. Their usage is not very wide-spread, but the Linux kernel uses them everywhere for instance, it is not something I came up with. I understand that they don't feel "natural" for C developers as you don't NULL-check the pointers, but the complexity is not higher - it's just different.

The big disadvantage of a central error code approach is that you only store the code of the *last* function call, which is prone to bugs caused by bad function ordering. For instance:
```
ctx = iio_create_context(...);
if (!ctx)
    goto handle_err;

dev = iio_context_find_device(...);
if (!dev)
    goto free_ctx;

...

free_ctx:
    iio_context_destroy(ctx);
handle_err:
    return iio_get_error();
```

In this case, if for some reason you get a NULL-pointer "dev", and `iio_context_destroy` fails and sets the error code, then you are not getting the *actual* error code that describes the reason why "dev" was NULL. Even worse, is that those bugs are typically found in error handling code, which is generally rarely executed - so rarely found.

Pointer-encoded errors are much more bug-proof because of that.

sternmull:
@pcercuei
About the attributes: Sounds good! Your solution was actually the first that I came up with. But then I changed my mind to get rid of the `get_attrs_count` of the objects. But that came at the price of the index argument for all the attrs-functions. So your way may be the better one. I am happy you consider it.
Actually both are a compromise. The super-strictly object oriented consistent approach would be to return an `iio_attrs` with `iio_attrs_get_count` and `iio_attrs_get` functions. And then use your functions for accessing the attributes. But while this looks "right" it also feels a bit too much indirection, especially when coming from the old API.

About the error handling: I understand the motivation and its implementation. But I did some research and came to the conclusion that this approach is only valid inside the kernel. Linux user space applications can allocate memory at low addresses, even 0x0. I created a little test that allocates memory at 0x0 and writes/reads it: https://gist.github.com/sternmull/85ff3ef20b410ef90e03a7aff4470184

I guess this is very exotic. But the current design of the library could make applications fail with strange error codes, depending on system configuration and address space layout.

I share your concerns about `errno`-style error handling. One upside is that the concept (including its pitfalls) should be familiar to most C programmers. I personally would prefer to return the error codes explicitly in some way, at the cost of less ergonomic function signatures (but of course I tend to wrap C APIs with easy to use C++ helpers if they feel too inconvenient/error prone to use). Some approaches that come to mind:

A) Always return an error code, move actual return values to pointer-to-output arguments:

```
if (error = iio_whatever(foo, bar, &retval)
  return handle_error(error);

use_return_value(retval);
```
B) Return real result but add pointer-to-error argument to every function that might fail:

```
int retval = iio_whatever(foo, bar, &error);

if (error)
  return handle_error(error);

use_return_value(retval);
```

C) Return a struct that pairs the return value with an error code:

```
struct iio_int_ret ret = iio_whatever(foo, bar);
if (ret.error)
  return handle_error(error);

use_return_value(ret.val);
```
A) Gets strange for functions that do not fail. Because it would be confusing to return the actual return-value instead of the error in that special cases. But it also looks stupid to make them void-return and use a pointer-to-output for the return value. So it would be either inconsistent (and dangerous to use) or look stupid in some places.

C) I don't see a real problem with that. But it is very uncommon.

I guess B) is the best I can come with up with right now.

In any case I would avoid to return non-NULL-pointers for errors. It is just too common for C code expect NULL to mean "error instead of valid object" and everything else to be the address to a valid object. At least in user space.

P.S. If we continue to discuss the new API then I think we shouldn't do it here where it should actually be about the pull request for the old API.

pcercuei:
@sternmull the error codes are *negative* values between [-4095, -1]. So a pointer encoding for error -EINVAL (aka. -22) would point to address 0xffff'ffea. You will never have a case where this is a valid pointer value.

rgetz:
while debating pro/cons of different approaches for libiio 1.x is excellent, and to be encouraged - I don't want to loose sight of this PR :)

@sternmull - based on anything moving forward for libiio 1.x - are there any changes that you might want to make to the existing PR, so things don't change as much when you move to libiio 1.x (the dev branch)?

@pcercuei : does that make sense to split discussion of libiio 1.x and review of the C++ headers for 0.25 separately? or you think it's potentially linked?

Thanks
-Robin

CMakeLists.txt
```diff
@@ -485,11 +496,13 @@ set_target_properties(iio PROPERTIES
 	VERSION ${VERSION}
 	SOVERSION ${LIBIIO_VERSION_MAJOR}
 	FRAMEWORK ${OSX_FRAMEWORK}
-	PUBLIC_HEADER ${LIBIIO_HEADERS}
 	C_STANDARD 99
 	C_STANDARD_REQUIRED ON
 	C_EXTENSIONS OFF
 )
+
+set_property(TARGET iio PROPERTY PUBLIC_HEADER ${LIBIIO_HEADERS})
```
pcercuei, line 504: Why that change? It does the same as the set_target_properties() above.

CMakeLists.txt
```diff
@@ -514,7 +527,13 @@ if(WITH_DOC)
 	# just switching off the Doxygen's HAVE_DOT option
 	find_package_handle_standard_args(Dot REQUIRED_VARS DOXYGEN_DOT_EXECUTABLE)
 
-	include(cmake/CheckCaseSensitiveFileSystem.cmake)
+
+        string(REPLACE ";" " " LIBIIO_DOXYGEN_DEFINES "${LIBIIO_DOXYGEN_DEFINES}")
+        string(REPLACE ";" " " DOXYGEN_ENABLED_SECTIONS "${DOXYGEN_ENABLED_SECTIONS}")
+        string(REPLACE ";" " " DOXYGEN_INPUT "${DOXYGEN_INPUT}")
+        string(REPLACE ";" " " DOXYGEN_STRIP_FROM_PATH "${DOXYGEN_STRIP_FROM_PATH}")
```
pcercuei, line 534: Looking at the CMake script, those lists will only ever have one entry each, do you really need to use lists there? I would think that using regular CMake strings would be enough. Then you wouldn't have to do character substitution like that.

bindings/cpp/examples/iiopp-enum.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+    try
+    {
+        enumerateIioEntities();
+    }
+    catch (std::exception & e)
+    {
+        cerr << "ERROR: " << e.what() << endl;
+        return EXIT_FAILURE;
+    }
```
pcercuei, line 100: This works, but not exactly what I had in mind :)

I meant something like:
```cpp
try {
    context = create_default_context();
} catch (std::exception& e) {
    cerr << "Unable to create Libiio context: " << e.what() << endl;
    return EXIT_FAILURE;
}
```

Just so that it's clear for users that the error occurs when creating the IIO context.

pcercuei (changes requested):
Just a few comments below.

CMakeLists.txt
sternmull: Because `LIBIIO_HEADERS` became a list when i append `iopp.h`. And CMake complained about multiple values when it was used as `PUBLIC_HEADER` in `set_target_properties`. So this was my workaround. Maybe there is a better way, I am no CMake expert.

CMakeLists.txt
sternmull: `DOXYGEN_INPUT` can have multiple entries. As the others are conceptually lists as well they got the same treatment. Just for consistency.

bindings/cpp/examples/iiopp-enum.cpp
sternmull: The error messages of the exceptions always mention the C-function that reported the error. So the error message is quite explicit and mentions the context creation function (if it was the reason). Of course it also works well if some other function fails.

sternmull:
@pcercuei 

> @sternmull the error codes are _negative_ values between [-4095, -1]. So a pointer encoding for error -EINVAL (aka. -22) would point to address 0xffff'ffea (on a 32-bit system). You will never have a case where this is a valid pointer value.

Oh, yeah... I should have paid more attention to the details. I guess at that end of the address space there is no problem.

But even if it works, I personally would still prefer a `NULL` return as error indicator. If I see a NULL-pointer in the debugger then I immediately think "aha, this points nowhere". For other addresses I usually assume that they point to a valid object. So does the debugger when it tries to inspect the pointee.

CMakeLists.txt
pcercuei: If you escape the variable with commas it works: `PUBLIC_HEADER "${LIBIIO_HEADERS}"`

CMakeLists.txt
pcercuei: It would probably be better to replace your:
```cmake
list(APPEND DOXYGEN_INPUT "${CMAKE_SOURCE_DIR}/bindings/cpp/")
```
with:
```cmake
set(DOXYGEN_INPUT "${DOXYGEN_INPUT} ${CMAKE_SOURCE_DIR}/bindings/cpp/")
```

Then none of them would need character substitution.

Note that with CMake 3.x you could use `string(APPEND ...)` or `list(JOIN ...)` but we can't use them yet as we keep compatibility with CMake 2.8.

bindings/cpp/examples/iiopp-enum.cpp
pcercuei: Fair enough.

CMakeLists.txt
sternmull: Done.

CMakeLists.txt
sternmull: Done.

sternmull:
@rgetz 
> @sternmull - based on anything moving forward for libiio 1.x - are there any changes that you might want to make to the existing PR, so things don't change as much when you move to libiio 1.x (the dev branch)?

No. I think the approach I took for the old API is sufficiently generic and should transfer well enough to the new one. Of course with some adjustments where the new API differs.

CMakeLists.txt
```diff
@@ -485,11 +496,12 @@ set_target_properties(iio PROPERTIES
...
 	SOVERSION ${LIBIIO_VERSION_MAJOR}
 	FRAMEWORK ${OSX_FRAMEWORK}
-	PUBLIC_HEADER ${LIBIIO_HEADERS}
-	C_STANDARD 99
+        PUBLIC_HEADER "${LIBIIO_HEADERS}"
+        C_STANDARD 99
 	C_STANDARD_REQUIRED ON
 	C_EXTENSIONS OFF
 )
+
```
pcercuei, line 504: And please unstage this. Although it looks better, we prefer to have the commits as minimal as possible and not carry unrelated changes.

sternmull:
@pcercuei Done. I didn't pay attention to the whitespace that the autoformatter introduced when i authored the previous draft... damn.

pcercuei (approved):
Looks good!
Thanks a lot for your contribution!

PR #973, opened 2023-06-01
iiod: Re-create IIO context when receiving USR1 signal

pcercuei:
Handle USR1 as a signal that IIOD should re-create the IIO context.
This allows support for hot-plugging IIO devices at runtime; the USR1
signal can then be sent from a udev script, for instance.


pcercuei:
Closing this one early. I will rebase it on top of master instead, so that it is included in the v0.25 release.

Issue #911, opened 2022-10-05
make iiod dynamic on kernel (sysfs) changes


rgetz:
Today, when the kernel changes sysfs (due to a hardware, or device tree overlay change), iiod needs to be restarted manually. This sort of stinks, and should be fixed for libiio 1.0

It would be better if:
- iiod was using poll() on sysfs (greg says not to use inotify - https://www.spinics.net/lists/linux-fsdevel/msg73924.html but that might be outdated now). Maybe @mhennerich might have some thoughts.
- when sysfs changes - figure out what to do. If a device removes/adds attributes, it should likely return errors for all future calls, until the context is closed/re-opened. If the device doesn't change - it's likely OK to continue to work on it. I don't know if you chance device "A" inside a context - if all devices should be closed, or not.

This is based on a discussion that @pcercuei and I were having at GNU Radio Conference.

-Robin

mhennerich:
Well that's what udev is for. 
Please see uevent: https://documentation.suse.com/sles/12-SP4/html/SLES-all/cha-udev.html

You can write a simple udev rule which will restart IIOD automatically, when a device get's removed or added.
The only thing is that we need to inform the remote side when this has happened.
Exchange a new context and handle error conditions.

rgetz:
I think making udev killing iiod at random times when devices can be attached to them - and then just letting things time out / would be problematic from a user standpoint. I get that is what we have today - but we should hopefully be getting better...

iiod needs to notice things, and then turn off the existing connections off gracefully (or as graceful as they can be). There is a difference between getting an error saying "device changed, re-open context" and all the devices just randomly disappearing...

-Robin

nunojsa:
> I think making udev killing iiod at random times when devices can be attached to them 

Not sure about what this but I would suspect that you can control how udev can notify iiod on device unplug/plug (I guess we just care about any change on sysfs) without having to kill it (using udev could save us from polling on sysfs). I'm thinking on a specific signal for instance or something with pipes/sockets...

Anyways, my 5 cents on this one is that we should do this as simple as possible. Ultimately, it should be applications to handle this (reloading contexts). IIOD/libiio should just have a mechanism to let them know that things have changed. Handling this things dynamically is extremely hard (not even the kernel does it right...), so things can become extremely complicated very quickly...

rgetz:
> If IIOD gets killed then the applications would most likely get -EPIPE on all operations,

[...]

Issue #914, opened 2022-10-16
iio_context_get_devices_count


panxishan:
There is only one device found by the Libiio computer client through API (iio_context_get_devices_count) through the network connection, there are two devices found through API (iio_context_get_devices_count) through local connection on the ARM board, and there are two IIO devices under the actual connection, why is there one less device found on the computer client?

tfcollins:
Have you loaded/reloaded a driver after boot?

tfcollins:
@panxishan this still an issue?

panxishan:
It has been solved. I must load the driver before starting the iiod

Issue #640, opened 2020-12-22
Support for rescanning/reopening devices in iiod


yaperez-anssi:
I'm currently playing with a PlutoSDR where I want to later expose some MMIO registers (from the PL/FPGA part). Since it seems that the PlutoSDR heavily relies on the IIO subsystem I'm trying to craft an IIO kernel module for that, so it can get exposed by libiio/iiod accross USB or network.

Right now I'm testing the driver using a kernel module so I don't have to rebuild and reflash the image (even to RAM) everytime.

I can load the kernel module and create a new IIO_SW_DEVICE device node which locally appears on the PlutoSDR, but is not exposed by iiod. I guess that's because iiod would need to rescan the devices, or be restarted. Unfortunately it seems that on the PlutoSDR it's not possible to restart iiod (it'll disappear from the network), so having a way to ask iiod for a rescan would be really nice.

rgetz:
This really isn't easy with the current architecture. I did have a branch that did something similar (looking for changes in sysfs), but it's complicated - since you can have multiple clients, and the backing iiod needs to be the same as the the client that is attached to it.

Your concept works in the simple use case, but not in the complicated ones (multiple connected clients)...

Let me think about it some more.

yaperez-anssi:
Good point, the clients might need to be notified so they can rescan themselves or something. I have to admit my issue might be specific to the PlutoSDR, because usually you might just want to restart the daemon (and maybe the clients), but unfortunately I can't do that.

rgetz:
There is no "notification" in the existing libiio implementation. All the client can do is close the context, and re-open things if it different. That's the way that the current implementation is architected. I think the tests I was doing would function, but would require a lot more testing... Let me think about it a bit more...

rgetz:
This is sort of a dup of #911 now.

yaperez-anssi:
Yes indeed, thanks for the pointer, I'll subscribe to that one.

rgetz:
This can be closed now (I can't anymore).

yaperez-anssi:
Closing because duplicate of #911

PR #975, opened 2023-06-05
iiod: Re-create IIO context when receiving USR1 signal

pcercuei:
Handle USR1 as a signal that IIOD should re-create the IIO context.
This allows support for hot-plugging IIO devices at runtime; the USR1
signal can then be sent from a udev script, for instance.

### DIFF
--- a/iiod/iiod.c
+++ b/iiod/iiod.c
@@ -37,6 +37,11 @@
 #define _STRINGIFY(x) #x
 #define STRINGIFY(x) _STRINGIFY(x)
 
+static int start_iiod(const char *uri, const char *ffs_mountpoint,
+		      const char *uart_params, bool debug, bool interactive,
+		      bool use_aio, uint16_t port, unsigned int nb_pipes,
+		      int ep0_fd);
+
 struct client_data {
 	int fd;
 	bool debug;
@@ -128,6 +133,14 @@ static void set_handler(int signal, void (*handler)(int))
 
 static void sig_handler(int sig)
 {
+	thread_pool_stop(main_thread_pool);
+}
+
+static bool restart_usr1;
+
+static void sig_handler_usr1(int sig)
+{
+	restart_usr1 = true;
 	thread_pool_stop(main_thread_pool);
 }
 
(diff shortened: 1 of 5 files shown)


iiod/iiod.c
```diff
@@ -459,40 +469,77 @@ int main(int argc, char **argv)
...
+	do {
+		thread_pool_restart(main_thread_pool);
+		restart_usr1 = false;
+
+		ret = start_iiod(arg, ffs_mountpoint, uart_params, debug,
+				 interactive, use_aio, port, nb_pipes, ep0_fd);
+	} while (!ret && restart_usr1);
+
+	thread_pool_destroy(main_thread_pool);
+	close(ep0_fd);
```
nunojsa, line 507: is this on purpose? In case we don't get into `if (WITH_IIOD_USBD && ffs_mountpoint) `, won't we close stdin in here (being this a daemon it's likely already closed anyways)?

I'm fairly sure it's not really a problem but for correctness shouldn't we do something like:

```
if (WITH_IIOD_USBD && ffs_mountpoint)
    close(ep0_fd);
```

nunojsa (approved):
LGTM... You sneaked in some helpers in the patches but they are still easy enough to read so bahh, fine by me.

iiod/iiod.c
pcercuei: It's closed just before the main() exits, so it shouldn't really have any side effects. But yes, you are correct, that's an oversight.

pcercuei:
@nunojsa Updated with your feedback.

nunojsa (approved):

rgetz:
can we add an example of a udev script that does something like this then?

Thanks
-Robin

pcercuei:
@rgetz I don't actually have any ready. We have one script that loads DTS overlays and then send a USR1 signal to IIOD, we don't handle it with a udev rule.

rgetz:
This is a interesting feature that many (including me) have asked for in the past, and would close/resolve #911 ; #640 ; #914  with a little more doc, it would be great.

you should be able to emulate things by unbinding and re-binding a driver. 
https://lwn.net/Articles/143397/

PR #976, opened 2023-06-05
spelling: Add a github action to point out spelling issues

rgetz:
so they don't need to be fixed just before a release.


rgetz:
don't merge this yet - github actions don't run on draft PRs,

rgetz:
Thanks - this is Ok to merge now.

tfcollins:
I updated the action rules a bit. Will need approval on all external PRs

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,27 @@
+# GitHub Action to automate the identification of common misspellings in text files.
+# https://github.com/codespell-project/actions-codespell
+# https://github.com/codespell-project/codespell
+name: codespell
+on:
+  push:
+    branches:
+      - main
+      - '[0-9]+.x'
```
tfcollins, line 9: Not sure this is correct. Should at least contain master but [0-9]+.x I don't think applies to our namings

pcercuei:
Does Codespell only parse the changes of a PR, or the whole project every time?

Underlying question - if somebody sends a PR that triggers a false-positive, and I merge it as-is, will the autobuilders be broken in the master branch?

.github/workflows/codespell.yml
rgetz: yeah, this is fixed now.

rgetz:
> I updated the action rules a bit. Will need approval on all external PRs

seems excessive for spell check, but I understand (sort of). 

based on the doc, "By default, all first-time contributors require approval to run workflows." Isn't that good enough?

pcercuei:
I believe if we have it only on pull requests, and one PR gets merged with a typo, then the autobuilder job will fail for every subsequent PR (as it parses the whole project and not just the changes). So I guess the current setup is better.

rgetz:
I guess the other thing is with @tfcollins changes to the way that action rules are enabled - is that an ADI person needs to run the spell check manually for 3rd party PRs. You can merge to master with or without doing a spell check.

pcercuei:
Yes it's fine, I was just wondering if we could do without the `.codespell-ignore.txt` by just ignoring the false positives when merging PRs. But if it affects every subsequent PR then the current system is better.

tfcollins (approved):

PR #971, opened 2023-05-13
packaging: update

rgetz:
when things moved from travis-ci to azure, some packaging options were lost. Specifically DEB_DETECT_DEPENDENCIES for debian based systems. This ensures that the version numbers from the host systems are put in the deb files so that things are actually compaitible with the downstream systems.

Turn that back on, and fix the issues that weren't seen over the last 3 years while this was turned off.

PR #977, opened 2023-06-06
cmake: turn HWMON on by default, and SERIAL ON for CI

rgetz:
This turns HWMON on by default, and turns on serial backends in the CI systems so the default binaries include it.


rgetz:
This is pulled out of #971, to make it easier to review and merge that one. (and it was sort of separate anyway).

pcercuei (approved):

Issue #986, opened 2023-06-15
iio_device_set_kernel_buffers_count is not exported in libiio 1.0


catkira:
`iio_device_set_kernel_buffers_count` is declared in iio.h but not exported in libiio.1.0.
This causes a link error when the function is used:
```bash
benja@benjaopti7000:~/git/open5G_tools/cell_monitor/build$ make -j10
[ 20%] Automatic MOC and UIC for target cell_monitor
[ 20%] Built target cell_monitor_autogen
[ 60%] Building CXX object CMakeFiles/cell_monitor.dir/main.cpp.o
[ 60%] Building CXX object CMakeFiles/cell_monitor.dir/iio_buf_reader.cpp.o
[ 80%] Building CXX object CMakeFiles/cell_monitor.dir/cell_monitor_autogen/mocs_compilation.cpp.o
[100%] Linking CXX executable cell_monitor
/usr/bin/ld: CMakeFiles/cell_monitor.dir/iio_buf_reader.cpp.o: in function `Iio_buf_reader::open_iio_buffer()':
iio_buf_reader.cpp:(.text+0x436): undefined reference to `iio_device_set_kernel_buffers_count'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/cell_monitor.dir/build.make:134: cell_monitor] Error 1
make[1]: *** [CMakeFiles/Makefile2:84: CMakeFiles/cell_monitor.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
```
I see that `iio_device_set_kernel_buffers_count` is defined in compat.c. Is this function obsolete in libiio 1.0? if yes the declaration should also probably move from iio.h to iio-compat.h.

catkira:
ah I see in the code that the number of kernel buffers of libiio 0.x is equivalent to the number of created blocks in libiio 1.0. `iio_device_set_kernel_buffers_count` declaration should be moved to iio-compat.h then.

catkira:
solved with this PR https://github.com/analogdevicesinc/libiio/pull/987

PR #987, opened 2023-06-15
fix: move declaration of iio_device_set_kernel_buffers_count from iio/iio.h into iio-compat.h

catkira:
this addresses issue https://github.com/analogdevicesinc/libiio/issues/986

### DIFF
--- a/iio-compat.h
+++ b/iio-compat.h
@@ -216,6 +216,18 @@ __api __check_ret int iio_context_get_version(const struct iio_context *ctx,
 					      char git_tag[8]);
 
 
+/** @brief Configure the number of kernel buffers for a device
+ *
+ * This function allows to change the number of buffers on kernel side.
+ * @param dev A pointer to an iio_device structure
+ * @param nb_buffers The number of buffers
+ * @return On success, 0 is returned
+ * @return On error, a negative errno code is returned */
+__api __check_ret int iio_device_set_kernel_buffers_count(const struct iio_device *dev,
+		unsigned int nb_buffers);
+
+
+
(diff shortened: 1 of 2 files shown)


nunojsa:
I'm not sure about this... It feels to me that with #906 iio-compat.h is actually not used/needed anymore. But @pcercuei should know better.

nunojsa:
> hm yes, in that case the problem would be somewhere else.

IIRC (when I spoke with Paul), the idea of the compatibility layer is just to allow apps linked against  libiio0.x to still run if libiiio is upgraded to libiio1.x and not to allow you to still compile those same apps with the new lib.

I kind of asked if this was expected since I also saw all kind of errors when trying to use libiio1.x to compile osc.

That said, it does looks like `iio_device_set_kernel_buffers_count()` should be removed from iio/iio.h

pcercuei:
I think it is still possible to build a v0.x application with the compat layer (although I did not try that in ages), since the compatibility layer will have all the v0.x symbols. The `iio-compat.h` file can then be used as the v0.x `iio.h`, which includes the v1.x `iio/iio.h`.

However - this file could very well go away, since you could very well just use the regular v0.x `iio.h` from the master branch.

I didn't decide yet whether we should trash this file; in the meantime I guess it's OK to move old functions there.

@catkira just a comment on the commit description, try to keep the title below 72 characters long, and add a commit description about why this change is needed (basically, just say that the function has been removed from the new API).

pcercuei (approved):

pcercuei:
@catkira Thanks!

PR #989, opened 2023-06-20
codespell: ignore 'MiS'

pcercuei:
Otherwise Codespell fails when processing the examples.

### DIFF
--- a/.codespell-ignore.txt
+++ b/.codespell-ignore.txt
@@ -1,6 +1,7 @@
 SOM
 som
 mitre
+mis
 tread
 elease
 tREAD


tfcollins:
codespell has a bug: https://github.com/codespell-project/codespell/issues/2375

rgetz:
@pcercuei : can you check if "mis" works as the ignore?

Thanks

pcercuei:
For some reason `mis` works while `MiS` doesn't, even if it's spelled `MiS` in the code…

tfcollins (approved):

PR #994, opened 2023-06-21
debug: add timestamps to debug messages

rgetz:
To track down some issues on Windows - let's print out the checkpoints, and add timestamps to things.


rgetz:
this is a draft - do not push to the main branch.

rgetz:
I think this is replaced with #1086 and this should be closed without merging.

PR #995, opened 2023-06-22
README_BUILD.md: Fix Github not displaying array properly

pcercuei:
Github was getting confused as it thought some rows had two cells and others had three.


rgetz:
Looks good to me

README_BUILD.md
```diff
@@ -89,11 +89,11 @@ this is invoked as "-DLOG_LEVEL=Debug".
 
 Cmake Options     | Default | Description                                    |
```
tfcollins, line 90: Do you need a "|" at the start of this line as well?

pcercuei: Github only chokes when the first cell is empty - which is not the case here.

Issue #981, opened 2023-06-08
iio_buffer_refill() returns "bad file descriptor" error after timeout


catkira:
I am using a separat thread to read a iio buffer with `iio_buffer_refill`. When there is not data in the buffer for a duration larger than the timeout, I get a "timeout" return value which is ok. However when I try calling `iio_buffer_refill` after that I get a "bad file descriptor" error and I need to close to buffer to get out of this state. 
Is this the intended behaviour or a bug?

My current workaround is to make the timeout super long, so that the timeout is practically never reached. But in order to join the buffer-read thread, I need to call `iio_buffer_cancel` from the outside thread now to make the buffer-read thread join. I don't know if `iio_buffer_cancel` is thread safe or not. (I did not experience a crash so far).

pcercuei:
This is with the local backend, or the network backend?

We didn't really design the timeout feature as the way you are trying to use it, more as a way to detect that something is seriously wrong - and in that case it is expected that the application would just destroy the context.

I'll have a look to see if it's possible to tweak it.

pcercuei:
Could you try with the `pcercuei/iiod-handle-kernel-timeout` branch?

catkira:
that seems to be fixed now

catkira:
@pcercuei  This problem is now solved when I use libiio 0.x on front (ad936x device) and backend (network PC) (switching the frontend to this new PR solved it). But it still persists when I use libiio 1.0 on fronend and libiio 0.24 on backend, the fix does not help here. Should I make a separate issue for that or can You port this fix to libiio 1.0 ?

pcercuei:
It will (hopefully) be fixed there as soon as the master branch is merged back into the dev branch.

PR #996, opened 2023-06-22
IIOD: Better handle errors of iio_buffer_{refill,push}

pcercuei:
Pass the error code to the connected clients, instead of disconnecting them.

This means that for a very slow device, a connected client will simply get a -ETIMEDOUT error when a configured timeout expires, and will still be able to communicate with the server.

Fixes: #981.

### DIFF
--- a/iiod/ops.c
+++ b/iiod/ops.c
@@ -678,9 +678,22 @@ static void rw_thd(struct thread_pool *pool, void *d)
 			}
 
 			if (ret < 0) {
-				IIO_ERROR("Reading from device failed: %i\n",
-						(int) ret);
-				break;
+				/* Reading from the device failed - signal the
+				 * error to all connected clients. */
+
+				/* Don't use SLIST_FOREACH - see comment below */
+				for (thd = SLIST_FIRST(&entry->thdlist_head);
+				     thd; thd = next_thd) {
+					next_thd = SLIST_NEXT(thd, dev_list_entry);
+
+					if (!thd->active || thd->is_writer)
+						continue;
+
+					signal_thread(thd, ret);
+				}
+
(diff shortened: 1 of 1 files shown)


iiod/ops.c
```diff
@@ -760,8 +773,13 @@ static void rw_thd(struct thread_pool *pool, void *d)
 			for (thd = SLIST_FIRST(&entry->thdlist_head);
 					thd; thd = next_thd) {
 				next_thd = SLIST_NEXT(thd, dev_list_entry);
-				if (thd->active && thd->is_writer &&
-						thd->nb < sample_size)
+
+				if (!thd->active || !thd->is_writer)
+					continue;
+
+				if (ret < 0)
```
rgetz, line 780: I think I agree with codacy, this check will always be the same, since there is a check above that breaks out of the while loop...

nunojsa: Yeah, I guess you need to remove the outer check...

pcercuei: Yes that was supposed to be removed, good catch.

pcercuei:
Updated.

rgetz:
Is the removal of all the IIO_ERROR() on purpose? We no longer print anything out when we use to....

pcercuei:
> Is the removal of all the IIO_ERROR() on purpose? We no longer print anything out when we use to....

Yes. Before, we considered that if `iio_buffer_refill` or `iio_buffer_push_partial` failed, it was an error of IIOD itself (something that should not happen), and IIOD would cancel and close the buffer.

Now, we consider that these errors might happen, but they are not errors of IIOD itself. The error codes are therefore sent back to the connected client(s).

nunojsa (approved):

PR #1000, opened 2023-07-04
Merge master back into dev branch

pcercuei:
The merge did not go smoothly and I got breakages in a dozen files that had to be fixed manually.

As a consequence some bugs might have slipped in (or bug fixes slipped out) so I would advise some caution.


nunojsa:
> As a consequence some bugs might have slipped in (or bug fixes slipped out) so I would advise some caution.

Hmm but do you have some specific patches where you would like an extra eyes? As you can imagine, going through the whole merge is not doable :)

If there's no special patches to look at, I can just ack this and then we (you :wink:) deal with possible bugs...

rgetz (approved):

mhennerich (approved):

PR #1003, opened 2023-07-05
mdns: update windows to properly timeout

rgetz:
The current mdns code returns when there has not been any mdns traffic for 2 seconds - however - on a noisy network,  there is sometimes mdns traffic we don't care about, and we never timeout.

Refactor things, so it keeps track of when we last success(that matches "_iio._tcp"...

### DIFF
--- a/dns_sd_windows.c
+++ b/dns_sd_windows.c
@@ -376,8 +376,8 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 	int nfds, res, ret = -ENOMEM;
 	struct timeval timeout;
 	FILETIME nowtime;
-	ULARGE_INTEGER now; then;
-	__int64 now64 then64, diff;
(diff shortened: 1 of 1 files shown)


pcercuei:
I'll wait for Sean to confirm that it works.

Please squash the two commits together, and close that parenthesis ;)

pcercuei:
@rgetz note that I am using a cross-platform implementation of a counter in my `pcercuei/dev-new-dmabuf-api` branch. See the `iio_read_counter_us()` function in 95b02004acd236e4dfe3a769bc937c222e53e387.

If that implementation works for you I can add that function to the master branch so that you can use it directly.

pcercuei:
@rgetz I added a `pcercuei/master-fix-win-dns` branch which (should?) fix the issue differently.

rgetz:
replaced by #1005

PR #1005, opened 2023-07-06
Libiio DNS/SD scan fixes

pcercuei:
My take at #1003. Just creating a draft PR to get the CI builds for now.

### DIFF
--- a/dns_sd.h
+++ b/dns_sd.h
@@ -15,11 +15,15 @@
 
 #ifdef _WIN32
 #include <winsock2.h>
+#include <ntddndis.h>
+#include <netioapi.h>
 #else
+#include <net/if.h>
 #include <sys/param.h>
 #endif
 
-#define DNS_SD_ADDRESS_STR_MAX (40) /* IPv6 Max = 4*8 + 7 + 1 for NUL */
+/* IPv6 Max = 4*8 + 7 + 1 for '%' + interface length */
+#define DNS_SD_ADDRESS_STR_MAX (40 + IF_NAMESIZE)
 #define FQDN_LEN (255)              /* RFC 1035 */
 
 /* MacOS doesn't include ENOMEDIUM (No medium found) like Linux does */
--- a/dns_sd_avahi.c
+++ b/dns_sd_avahi.c
@@ -51,10 +51,13 @@ static struct dns_sd_discovery_data *new_discovery_data(void)
 }
 
 static void avahi_process_resolved(struct dns_sd_discovery_data *ddata,
+				   AvahiIfIndex iface,
 				   const AvahiAddress *addr,
 				   const char *host_name,
 				   const uint16_t port)
 {
+	char *ptr;
+
 	/* Avahi is multi-threaded, so lock the list */
 	iio_mutex_lock(ddata->lock);
 	ddata->resolved++;
@@ -82,6 +85,16 @@ static void avahi_process_resolved(struct dns_sd_discovery_data *ddata,
 		IIO_ERROR("Avahi Resolver : memory failure\n");
 	}
 	iio_mutex_unlock(ddata->lock);
+
+	ptr = ddata->addr_str + strnlen(ddata->addr_str, DNS_SD_ADDRESS_STR_MAX);
+
+	if (addr->proto == AVAHI_PROTO_INET6
+	    && addr->data.ipv6.address[0] == 0xfe
+	    && addr->data.ipv6.address[1] == 0x80
+	    && iface != AVAHI_IF_UNSPEC
+	    && if_indextoname((unsigned int)iface, ptr + 1)) {
+		*ptr = '%';
+	}
 
 	IIO_DEBUG("\t\t%s:%u (%s)\n", host_name, port, ddata->addr_str);
 }
@@ -126,7 +139,7 @@ static void __avahi_resolver_cb(AvahiServiceResolver *resolver,
 			  avahi_strerror(err));
 		break;
 	case AVAHI_RESOLVER_FOUND: {
-		avahi_process_resolved(ddata, address, host_name, port);
+		avahi_process_resolved(ddata, iface, address, host_name, port);
 		IIO_DEBUG("Avahi Resolver : service '%s' of type '%s' in domain '%s':\n",
 			  name, type, domain);
 		break;
@@ -157,7 +170,7 @@ static void avahi_host_resolver(AvahiHostNameResolver *resolver,
 			  host_name, avahi_strerror(err));
 		break;
 	case AVAHI_RESOLVER_FOUND:
-		avahi_process_resolved(ddata, address, host_name, IIOD_PORT);
+		avahi_process_resolved(ddata, iface, address, host_name, IIOD_PORT);
 		break;
 	}
 
--- a/dns_sd_windows.c
+++ b/dns_sd_windows.c
@@ -24,6 +24,9 @@
 #define STRINGIFY(x) _STRINGIFY(x)
 #define MDNS_PORT_STR STRINGIFY(MDNS_PORT)
 
+static uint64_t lasttime;
+#define TIMEOUT 2
+
 #ifdef HAVE_IPV6
 static const unsigned char localhost[] = {
 	0, 0, 0, 0, 0, 0, 0, 0,
@@ -347,6 +350,7 @@ static int query_callback(int sock, const struct sockaddr *from, size_t addrlen,
 		}
 	}
 #endif /* HAVE_IPV6 */
+	lasttime = iio_read_counter_us();
 	iio_mutex_unlock(dd->lock);
 quit:
 	return 0;
@@ -370,6 +374,8 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 	int transaction_id[32];
 	int nfds, res, ret = -ENOMEM;
 	struct timeval timeout;
+	uint64_t nowtime;
+	int64_t diff;
 
 	if (WSAStartup(versionWanted, &wsaData)) {
 		IIO_ERROR("Failed to initialize WinSock\n");
@@ -422,9 +428,11 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 	IIO_DEBUG("Reading mDNS query replies\n");
 
 	records = 0;
+	lasttime = iio_read_counter_us();
+
 	do {
 		nfds = 0;
-		timeout.tv_sec = 2;
+		timeout.tv_sec = TIMEOUT;
 		timeout.tv_usec = 0;
 
 		fd_set readfs;
@@ -448,6 +456,16 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
 				FD_SET(sockets[isock], &readfs);
 			}
 		}
+
+		/* res > 0 even if we didn't process anything :(
+		 * timeout from the last time we successfully added a proper mdns record
+		 */
+		nowtime = iio_read_counter_us();
+
+		/* convert to ms */
+		diff = (nowtime - lasttime) / 1000ull;
+		if (diff > (TIMEOUT * 1000))
+			res = 0;
 	} while (res > 0);
 
 	for (isock = 0; isock < num_sockets; ++isock)
--- a/network.c
+++ b/network.c
@@ -1060,7 +1060,7 @@ static bool msg_trunc_supported(struct iiod_client_pdata *io_ctx)
 }
 #endif
 
-struct iio_context * network_create_context(const char *host)
+struct iio_context * network_create_context(const char *hostname)
 {
 	struct addrinfo hints, *res;
 	struct iio_context *ctx;
@@ -1072,6 +1072,10 @@ struct iio_context * network_create_context(const char *host)
 	char *description, *uri, *end, *port = NULL;
 	char port_str[6];
 	uint16_t port_num = IIOD_PORT;
+	char host_buf[FQDN_LEN + sizeof(":65535") + 1];
+	char *host = hostname ? host_buf : NULL;
+
+	iio_strlcpy(host_buf, hostname, sizeof(host_buf));
 
 #ifdef _WIN32
 	WSADATA wsaData;
--- a/utilities.c
+++ b/utilities.c
@@ -20,6 +20,12 @@
 #include <stdlib.h>
 #include <string.h>
 
+#ifdef _WIN32
+#include <Windows.h>
+#else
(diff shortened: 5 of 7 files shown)


utilities.c
```diff
@@ -359,3 +365,25 @@ ssize_t __iio_printf iio_snprintf(char *buf, size_t len, const char *fmt, ...)
...
+	LARGE_INTEGER freq, cnt;
+
+	QueryPerformanceFrequency(&freq);
+	QueryPerformanceCounter(&cnt);
+
+	value = (1000000 * cnt.QuadPart) / freq.QuadPart;
+#else
+	struct timeval tv;
+
+	gettimeofday(&tv, NULL);
```
nunojsa, line 383: from the man page:

```
APPLICATION USAGE
       Applications should use the clock_gettime() function instead of the obsolescent gettimeofday() function.

RATIONALE
       None.

FUTURE DIRECTIONS
       The gettimeofday() function may be removed in a future version.

```

rgetz:
this can time out too early...

If you are getting `_iio._tcp.local` mdns responses - that's what the timeout should be based on. You should wait 2 seconds after you got the last "good" mdns response. Not just wait 2 seconds (which is what this seems to do).

You need to reset the timer at https://github.com/analogdevicesinc/libiio/blob/b470740ad9f22c8db9fde1e98e9dda94d28a5827/dns_sd_windows.c#L352 (while the lock is being held, to make sure there isn't a threading issue).

-Robin

pcercuei:
Why can't you just wait 2 seconds total?

pcercuei:
@rgetz I reverted to the old behaviour.

And added some commits on top that should fix scanning with IPv6 and a nasty bug in the network backend.

I also changed how the URIs are reported by the scanning mechanism, now it will use the IPs instead  of the hostname in the URIs. I think it makes sense but I'd like feedback on this one.

rgetz:
>Why can't you just wait 2 seconds total?

because sometimes it takes more than 2 seconds for a device to respond. Depending on how many devices are plugged into the network, it can be 4 seconds (or more) for all devices to chirp off with their mdns responses... That's why I set things up to be either 4 seconds from the last valid response, or 2 seconds from any response (so it's as responsive as possible, but still waits long enough).

When I was testing - I had ~15 devices on the network (10 Plutos + USB Ethernet adapters, plus a couple Zedboards, and a couple RPis, and x86 machines).

-Robin

rgetz:
>I also changed how the URIs are reported by the scanning mechanism, now it will use the IPs instead of the hostname in the URIs. I think it makes sense but I'd like feedback on this one.

I'm not sure what you mean. Hostnames are much easier to understand - and tell you something about the machine. On a DHCP managed network - IP addresses change, so you will have no idea what you are talking to, and could be talking to the wrong machine day to day... (I agree that duplicated host names are just as bad with the pluto-1.local, pluto-2.local, pluto-3.local names; but most people have properly configured hostnames, and will rename/reboot things, so they will get the names they want (I have "pluto-adsb", "pluto-fmradio", "pluto-gps"; changing that to 192.168.1.138 - would be meaningless to me).

I think that change would also break some existing applications - since it expects an hostname, and will do a lookup on it...

-Robin

dns_sd.h
```diff
@@ -15,11 +15,15 @@
...
+#include <ntddndis.h>
+#include <netioapi.h>
 #else
+#include <net/if.h>
 #include <sys/param.h>
 #endif
 
-#define DNS_SD_ADDRESS_STR_MAX (40) /* IPv6 Max = 4*8 + 7 + 1 for NUL */
+/* IPv6 Max = 4*8 + 7 + 1 for NUL + 1 for '%' + interface length */
+#define DNS_SD_ADDRESS_STR_MAX (41 + IF_NAMESIZE)
```
rgetz, line 26: `IF_NAMESIZE` already includes the null byte the way I read [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/net_if.h.html) which states "_symbolic constant for the length of a buffer containing an interface name (including the terminating NULL character)_".

rgetz (commented):
This makes sense - thanks for the fix.

It also might make sense to add something like:
`- ipv6       2001:db8:4006:812::200e%eth0` 
in https://github.com/analogdevicesinc/libiio/blob/master/network.c#L1101-L1104

or combinations of things that you need to support. (since I think that its valid to end an ipv6 address with `::` which would make it `::%`, which would make the port logic need to be updated... (I think it's only for multicast, so maybe it's OK?), but I would still add something in there so it's not forgotten in the future.

pcercuei:
> > I also changed how the URIs are reported by the scanning mechanism, now it will use the IPs instead of the hostname in the URIs. I think it makes sense but I'd like feedback on this one.
> 
> I'm not sure what you mean. Hostnames are much easier to understand - and tell you something about the machine. On a DHCP managed network - IP addresses change, so you will have no idea what you are talking to, and could be talking to the wrong machine day to day... (I agree that duplicated host names are just as bad with the pluto-1.local, pluto-2.local, pluto-3.local names; but most people have properly configured hostnames, and will rename/reboot things, so they will get the names they want (I have "pluto-adsb", "pluto-fmradio", "pluto-gps"; changing that to 192.168.1.138 - would be meaningless to me).
> 
> I think that change would also break some existing applications - since it expects an hostname, and will do a lookup on it...

I am just talking about scanning here. Using hostnames e.g. a `ip:analog.local` URI to create a context would still work fine.

The problem with scanning is that you get duplicated URIs for IPv4 vs. IPv6 as shown in my commit message:
```
    Available contexts:
            0: 192.168.1.184 AD-FMCOMMS3-EBZ on Xilinx Zynq ZED [ip:analog.local]
            1: fe80::18e7:703c:611a:f8e8%enp4s0 AD-FMCOMMS3-EBZ on Xilinx Zynq ZED [ip:analog.local]

```

utilities.c
pcercuei: I had no idea. I updated the code for clock_gettime() with the CLOCK_MONOTONIC clock now.

pcercuei:
@rgetz I dropped the commit that changed hostnames to IPs in the URIs returned by the scan. While I think we definitely should fix it one way or another, there's no time to argue about this right now - v0.25 should be released ASAP and this can wait.

dns_sd_windows.c
```diff
@@ -448,6 +456,16 @@ int dnssd_find_hosts(struct dns_sd_discovery_data **ddata)
...
 			}
 		}
+
+		/* res > 0 even if we didn't process anything :(
+		 * timeout from the last time we successfully added a proper mdns record
+		 */
+		nowtime = iio_read_counter_us();
+
+		/* convert to ms */
+		diff = (nowtime - lasttime) / 1000LL;
```
nunojsa, line 466: nit: is the `LL` really needed? In theory, this could actually get wrong, if `lastime > nowtime` but I guess that should be impossible :thinking:

dns_sd_avahi.c
```diff
@@ -83,7 +86,13 @@ static void avahi_process_resolved(struct dns_sd_discovery_data *ddata,
...
 	iio_mutex_unlock(ddata->lock);
 
-	IIO_DEBUG("\t\t%s:%u (%s)\n", host_name, port, ddata->addr_str);
+	ptr = ddata->addr_str + strlen(ddata->addr_str);
+
+	if (addr->proto == AVAHI_PROTO_INET6
+	    && iface != AVAHI_IF_UNSPEC
+	    && if_indextoname((unsigned int)iface, ptr + 1)) {
+		*ptr = '%';
+	}
```
nunojsa, line 95: shouldn't we move `IIO_DEBUG` here? Or at least state in the message that you want to explicitly remove that log

network.c
```diff
@@ -1072,6 +1072,10 @@ struct iio_context * network_create_context(const char *host)
 	char *description, *uri, *end, *port = NULL;
 	char port_str[6];
 	uint16_t port_num = IIOD_PORT;
+	char host_buf[FQDN_LEN + sizeof(":65535") + 1];
+	char *host = hostname ? host_buf : NULL;
+
+	iio_strlcpy(host_buf, hostname, sizeof(host_buf));
```
nunojsa, line 1078: love the first sentence :smile:

nunojsa (commented):
Mostly nits from my side... The most important bit from my side is the `static` stuff

dns_sd_avahi.c
pcercuei: I actually removed it accidentally. Good catch.

dns_sd_windows.c
pcercuei: Should probably be `ULL` yes ;) Not strictly required but I like to avoid implicit type casting when I can.

Both Windows and POSIX implementations of `iio_read_counter_us` use monotonic clocks, which mean that they can never go backwards. And therefore `lasttime >= nowtime` is always true.

network.c
pcercuei: I mean, who thought it was a good idea to return a non-const pointer to an immutable string?

utilities.c
```diff
@@ -359,3 +365,25 @@ ssize_t __iio_printf iio_snprintf(char *buf, size_t len, const char *fmt, ...)
...
+	LARGE_INTEGER freq, cnt;
+
+	QueryPerformanceFrequency(&freq);
+	QueryPerformanceCounter(&cnt);
+
+	value = (1000000 * cnt.QuadPart) / freq.QuadPart;
+#else
+	struct timespec ts;
+
+	clock_gettime(CLOCK_MONOTONIC, &ts);
```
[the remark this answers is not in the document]
pcercuei: It would complicate things quite a bit, I think, if we consider that `iio_read_counter_us` can return an error.

Besides, the Windows implementation is guaranteed not to fail on versions >= XP, and I doubt we want to support anything older than that. And the only way the POSIX implementation can fail, is if monotonic timers are not available, which will never happen on any of the systems that Libiio supports right now.

pcercuei:
Updated with Nuno's comments.

utilities.c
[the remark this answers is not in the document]
nunojsa: ok, fair enough

network.c
nunojsa: yeah, great API design...

nunojsa (approved):
LGTM

pcercuei:
@rgetz All good for you?

rgetz:
> The problem with scanning is that you get duplicated URIs for IPv4 vs. IPv6 as shown in my commit message:
> 
> ```
>     Available contexts:
>             0: 192.168.1.184 AD-FMCOMMS3-EBZ on Xilinx Zynq ZED [ip:analog.local]
>             1: fe80::18e7:703c:611a:f8e8%enp4s0 AD-FMCOMMS3-EBZ on Xilinx Zynq ZED [ip:analog.local]
> ```

yeah, that is how it was designed on purpose... 

applications like iioscope will use `ip:analog.local` and can either decide to show the IP number, or get rid of duplicates and let the OS decide how to resolve things (ipv4 or ipv6).

It's the name that matters, not the IP number - the name shouldn't change, but the IP number might get a new one via DHCP.

-Robin

rgetz (commented):
don't we need the same fix for all OSes? (Windows and Mac?) it looks like we only do this on Linux?

This is something I wasn't completely aware of - but in order to resolve the ambiguity in textual link-local addresses, a zone index must be appended to the address (since the same address may be valid in different zones, and in use by a different host in each of those zones) - which makes sense.

but it's the same issue exists on windows/macos
  * `fe80::1ff:fe23:4567:890a%eth2` this (using an interface name) is customary on most unix-like operating systems (e.g., BSD, Linux, macOS. 
  * `fe80::1ff:fe23:4567:890a%3` This (using an interface number) is the standard syntax on Windows, but as support for this syntax is mandatory, it is also available on other operating systems.

Also the [Wikipedia](https://en.wikipedia.org/wiki/IPv6_address) page warns:
_When used in uniform resource identifiers (URI), the use of the percent sign causes a syntax conflict, therefore it must be escaped via percent-encoding with_ `a%25` _example:_ `http://[fe80::1ff:fe23:4567:890a%25eth0]/`

I don't know if we need that or not?

and if we are going to accept that (which I agree we should) we should update [iio.h](https://github.com/analogdevicesinc/libiio/blob/master/iio.h#L470-L477) to include that in the url examples, so it shows up in the on-line doc.

rgetz:
the others look good to me - thanks.

I will close the other PRs.

dns_sd_avahi.c
```diff
@@ -83,6 +86,14 @@ static void avahi_process_resolved(struct dns_sd_discovery_data *ddata,
 	}
 	iio_mutex_unlock(ddata->lock);
 
+	ptr = ddata->addr_str + strlen(ddata->addr_str);
```
rgetz, line 89: codacy flags this:
strlen does not handle strings that are not \0-terminated; if given one it may perform an over-read

could be:
ptr = ddata->addr_str + strnlen(ddata->addr_str, DNS_SD_ADDRESS_STR_MAX);

pcercuei: But the string is guaranteed to be \0-terminated, isn't it?

rgetz: It should be - as far as I can tell.

Since it's not protected from the app that's using it - a bad pointer somewhere else could over write things.

But you are right - it's low likelihood.

rgetz:
In the "avahi: Fix detection of contexts at IPv6 addresses" - we only need that for ipv6 local link addresses - correct? 

Shouldn't we be checking  link-local prefix fe80::/10? Otherwise we add it to all - and Linux is sort of notorious for changing the order of Ethernet cards boot to boot.. (which card shows up as eth0, which shows up as eth1). 

Right now - many applications just use and save the uri across boots, and this may cause some things that use to work - to start failing if we add it to everything...

I think?

I get this is handled in many modern distributions with a generated /etc/udev/rules.d/70-persistent-net.rules - but it isn't in all...

- Robin

pcercuei:
> In the "avahi: Fix detection of contexts at IPv6 addresses" - we only need that for ipv6 local link addresses - correct?

Yes. I'm pretty sure that Avahi won't discover anything outside the local scope. But I amended the commit to only append the interface name to link-local addresses.

> don't we need the same fix for all OSes? (Windows and Mac?) it looks like we only do this on Linux?

Windows doesn't have any problems discovering contexts over IPv6 here while Linux did, so the issue is less pressing. I did not test OSX, but we didn't get any bug reports there - so maybe it's broken but nobody uses this usecase yet :)

As such I think it is OK to leave it for later, I can't delay the v0.25 release forever.

PR #1009, opened 2023-07-14
CI: Install dependencies in /usr/local/lib/libiio on OSX

pcercuei:
The previous installation directory was /usr/local/lib which is where Homebrew will install its own libraries, which can then conflict with ours.

### DIFF
--- a/CI/azure/macos_tar_fixup.sh
+++ b/CI/azure/macos_tar_fixup.sh
@@ -35,20 +35,20 @@ if [ ! -f "${libserialport_loc}" ]; then
 fi
 
 # Create links to framework files
-mkdir -p usr/local/{lib,include}
+mkdir -p usr/local/include usr/local/lib/libiio
 ln -fs "${libiio_loc}" usr/local/lib/libiio.dylib
 ln -fs "${libiioheader_loc}" usr/local/include/iio.h
 
 # Copy dependent libs to local libs
-cp "${libusb_loc}" usr/local/lib/
-cp "${libxml_loc}" usr/local/lib/
-cp "${libserialport_loc}" usr/local/lib/
-chmod +w usr/local/lib/libusb-1.0.dylib
-chmod +w usr/local/lib/libxml2.dylib
-chmod +w usr/local/lib/libserialport.dylib
-install_name_tool -id @rpath/libusb-1.0.dylib usr/local/lib/libusb-1.0.dylib
-install_name_tool -id @rpath/libxml2.dylib usr/local/lib/libxml2.dylib
-install_name_tool -id @rpath/libserialport.dylib usr/local/lib/libserialport.dylib
+cp "${libusb_loc}" usr/local/lib/libiio/
+cp "${libxml_loc}" usr/local/lib/libiio/
+cp "${libserialport_loc}" usr/local/lib/libiio/
+chmod +w usr/local/lib/libiio/libusb-1.0.dylib
+chmod +w usr/local/lib/libiio/libxml2.dylib
+chmod +w usr/local/lib/libiio/libserialport.dylib
+install_name_tool -id @rpath/libusb-1.0.dylib usr/local/lib/libiio/libusb-1.0.dylib
+install_name_tool -id @rpath/libxml2.dylib usr/local/lib/libiio/libxml2.dylib
+install_name_tool -id @rpath/libserialport.dylib usr/local/lib/libiio/libserialport.dylib
 
 # Update rpath of library
 install_name_tool -change "${libusb_loc}" "@rpath/libusb-1.0.dylib" "${libiio_loc}"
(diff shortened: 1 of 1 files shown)


rgetz:
I'm not sure - all the paths are relative (have no leading slash), so if you are in `foo` when running, things would be `foo/usr/local/lib` ... which is a more "standard" path than `foo/adi/opt/lib`

It's only if you run the script from the root directory that it would conflict. (So, maybe that's the fix? - don't allow that).

pcercuei:
The relative paths are only for fixing the rpath - the installed libiio will find its dependencies in `/opt/adi/lib/`.

rgetz:
>The relative paths are only for fixing the rpath - 
right

> the installed libiio will find its dependencies in /opt/adi/lib/.
so, that's what I'm still missing - shouldn't there be a change to set `DESTDIR` or something then?

There doesn't seem to be anything that installs anything in /opt/adi/lib. all the changes are just changing the path in the tarball right now. (or I'm missing something).

pcercuei:
@rgetz well yes, but you extract the tarball at the root folder, don't you?

pcercuei:
@rgetz the whole point of this patch is to move away the dependencies if you extract the tarball at the root - so that it doesn't conflict with the libraries installed by Homebrew. If you extract them somewhere else it shouldn't matter, as Libiio should still be able to find its dependencies, as the rpath is relative to the library's install location.

rgetz:
> @rgetz well yes, but you extract the tarball at the root folder, don't you?

No that's not the intent - MathWorks installs it off path, as non-root, so that it doesn't conflict with anything in the system directories. That's why we need a completely relative rpath fixup.  It will end up in `/wherever/the/user/installs/matlab/leaf/directory`

That's why I don't think this is necessary.

rgetz:
> if you extract them somewhere else it shouldn't matter, as Libiio should still be able to find its dependencies, as the rpath is relative to the library's install location.

right - but a leaf directory of `usr/local/lib` is more intuitive than `opt/adi/lib` isn't it?
for MATLAB users, they will end up with one of:
`/wherever/the/user/installs/matlab/leaf/directory/usr/local/lib`
or
`/wherever/the/user/installs/matlab/leaf/directory/opt/adi/lib`

if someone wants to install it into `/opt/adi`, they can do what you want with : `tar -xf file.tar -C /opt/adi` ? I think that should work, although you might get `/opt/adi/usr/local/lib` but then `tar xf file.tar --strip-components=2 -C /opt/adi` should work... (all untested, so I might have a syntax issue).

It's just different than previous libiio releases, for no apparent reason that I can see.

rgetz:
If you still want to do it - `opt/libiio` would be better than `opt/adi` for an open source vendor neutral library.

pcercuei:
> if someone wants to install it into `/opt/adi`, they can do what you want with : `tar -xf file.tar -C /opt/adi` ? I think that should work, although you might get `/opt/adi/usr/local/lib` but then `tar xf file.tar --strip-components=2 -C /opt/adi` should work... (all untested, so I might have a syntax issue).

But you don't want to install Libiio to `/opt/adi` - you only want to install the *dependencies* there, and have the library properly installed in `/Library/Frameworks/iio.framework/iio`, as well as a symlink to `/usr/local/lib/libiio.dylib`.

The only thing that changes from previous releases is the location of the *dependencies*.

pcercuei:
@rgetz Alternatively I can install the dependencies to `usr/local/lib/libiio/`.

rgetz:
> @rgetz Alternatively I can install the dependencies to `usr/local/lib/libiio/` or even `usr/local/lib/libiio-deps/`.

sure that makes sense - as long as all the tar balls are created the same (ie - there isn't a libiio-deps in the windows or linux tar balls either - they are just all jammed in the same directory as the libiio libraries - same as this).

pcercuei:
> sure that makes sense - as long as all the tar balls are created the same (ie - there isn't a libiio-deps in the windows or linux tar balls either - they are just all jammed in the same directory as the libiio libraries - same as this).

There are no Linux tarballs. And on Windows, you don't risk overwriting system libs.

pcercuei:
> > @rgetz well yes, but you extract the tarball at the root folder, don't you?
> 
> No that's not the intent - MathWorks installs it off path, as non-root, so that it doesn't conflict with anything in the system directories. That's why we need a completely relative rpath fixup. It will end up in `/wherever/the/user/installs/matlab/leaf/directory`
> 
> That's why I don't think this is necessary.

The OSX tarballs contain the file hierarchy, therefore it is a perfectly valid operation to extract them at the root. The fact that MathWorks doesn't do that doesn't mean that the fix is invalid.

pcercuei:
Closing this because I changed my mind, and it tries to solve a non-issue.

The tarballs should not be extracted at the root and doing that is asking for trouble.

PR #1010, opened 2023-07-14
device: Fix sample size calculation

pcercuei:
If we consider a system where channel 0 delivers 32-bit data and channel 1 delivers 16-bit data, the sample size computed would be (4 + 2) == 6 bytes.

This is not correct, as it does not take into account the padding needed until the next sample, which must be computed in a way that all the elements of the next sample are aligned to their own size.

This new algorithm follows what the Linux kernel does (see: iio_compute_scan_bytes() in drivers/iio/industrialio-buffer.c as of v6.4).

This should fix a practical issue with the ADIS16475 driver, reported by @rbolboac.

### DIFF
--- a/device.c
+++ b/device.c
@@ -442,7 +442,7 @@ ssize_t iio_device_get_sample_size_mask(const struct iio_device *dev,
 		const uint32_t *mask, size_t words)
 {
 	ssize_t size = 0;
-	unsigned int i;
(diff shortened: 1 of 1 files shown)


nunojsa (approved):

rgetz (approved):
looks good to me.

PR #1016, opened 2023-07-24
CI: fix empty Windows.zip release artifact

ccraluca:
Fix Windows zip file in release.


pcercuei:
Fix release zip of what? And what was wrong with it? :)
I know the answer, but that info should be in the commit message.

ccraluca:
Commit message updated. :)

pcercuei:
I think it should also say that we're talking about `Windows.zip` in particular. You could set the commit title to e.g. `CI: fix empty Windows.zip artifact`

pcercuei (approved):

PR #1017, opened 2023-07-26
serial: Pass port name/description as context attributes

### LINKED ISSUE #926: context description with serial backend
1. Given the following string in xml:

```
<context name="xml" version-major="0" version-minor="24" version-git="c4498c2" description="no-OS master-9271c4ee2-modified" >
```
When using the serial backend, I get the following context name and description from libiio:
```
>>> serctx.name
'serial'
>>> serctx.description
'/dev/ttyACM0: DAPLink CMSIS-DAP - 0425170267cb0dfe00000000000000000000000097969906'
```

2. Given the following string in xml:

```
<context name="xml" version-major="0" version-minor="24" version-git="c4498c2" description="Linux (none) 5.4.0-gfd4058d9a174 #121 SMP PREEMPT Wed Apr 14 08:46:13 CEST 2021 armv7l" >
```
When using the usb backend, I get the folloing context name and description from libiio:
```
>>> usbctx.name
'usb'
>>> usbctx.description
'Linux (none) 5.4.0-gfd4058d9a174 #121 SMP PREEMPT Wed Apr 14 08:46:13 CEST 2021 armv7l'
```

At point 2, the description is as expected.
At point 1 however, the description contains serial port info instead of the xml description attribute.

I've traced this to this function:
https://github.com/analogdevicesinc/libiio/blob/e9c801112c28e27ada054bae5ed260d4618f5f21/serial.c#L113

I think it would be better to have consistent behavior across backends and populate context description in libiio with the content from xml and not from libserialport. (Perhaps populate with info from libserialport if the description attribute is missing from xml, but this is optional).

pcercuei:
Instead of overriding the context description that comes from the XML, pass the serial port name and description as context attributes, and keep the context description as-is.

Fixes #926.

### DIFF
--- a/serial.c
+++ b/serial.c
@@ -105,26 +105,6 @@ static int serial_get_version(const struct iio_context *ctx,
 			major, minor, git_tag);
 }
 
-static char * serial_get_description(struct sp_port *port)
-{
-	char *description, *name, *desc;
-	size_t desc_len;
-
-	name = sp_get_port_name(port);
-	desc = sp_get_port_description(port);
-
-	desc_len = sizeof(": \0") + strlen(name) + strlen(desc);
-	description = malloc(desc_len);
-	if (!description) {
-		errno = ENOMEM;
-		return NULL;
-	}
-
-	iio_snprintf(description, desc_len, "%s: %s", name, desc);
-
-	return description;
-}
-
 static int serial_open(const struct iio_device *dev,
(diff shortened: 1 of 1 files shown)


rgetz (approved):

buha (approved):
even tested this with hardware, works as expected

PR #1021, opened 2023-08-01
Rgetz iiod enable ip numbers

rgetz:
print out IP numbers when clients connect.

### DIFF
--- a/iiod/dns-sd.c
+++ b/iiod/dns-sd.c
@@ -256,7 +256,6 @@ static void start_avahi_thd(struct thread_pool *pool, void *d)
 
 	while(true) {
 		ret = gethostname(host, sizeof(host));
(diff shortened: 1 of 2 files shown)


tfcollins:
I'm wondering the downside here. So if we run the stress test will the log explode?

pcercuei:
@tfcollins it shouldn't, the amount of data printed varies very little.

rgetz:
>So if we run the stress test will the log explode?

The "add" is to replace `0.0.0.0` with the actual/proper IP number: `192.168.122.1` or `fe80::fc54:ff:fe57:b2c7%vnet5` so I think the log will fill up no matter what .... :)

-Robin

pcercuei:
No idea why Codacy still complains. It looks OK to me now.

pcercuei (approved):

PR #1022, opened 2023-08-04
Fix issues found when `-DLOG_LEVEL=NoLog`

rgetz:
This is just a quick check to see if there are any debug print issues.


rgetz:
There were some issues, fix them.

pcercuei:
I'll close it in favor of #1023, which should fix the problem at its root. I cherry-picked your second commit there too.

rgetz:
Thanks

PR #1035, opened 2023-08-23
DMABUF API v3

pcercuei:
Follow-up of https://github.com/analogdevicesinc/libiio/pull/928

These commits will modify how the local-DMABUF interface works, because the interface itself in the kernel changed (as it is not yet upstream).

The local backend will now make use of the "udmabuf" kernel driver, which was added in 5.19, to create DMABUF objects from memfd buffers. These DMABUFs can then be attached to the IIO device and/or other devices from other subsystems, and memory-mapped to access the data.

IIOD has also been modified to support enqueueing DMABUFs to the USB (functionfs) interface, and will support passing the buffers between IIO and USB in a zero-copy fashion.

From a Libiio API standpoint, some changes were introduced:
- `iio_block_enqueue` will support partial transfers on RX buffers as well;
- Setting bytes_used=0 in `iio_block_enqueue` will default to transferring the full buffer;
- New API function `iio_block_disable_cpu_access`, to disable CPU synchronization with the buffers when doing zero-copy;
- New API function `iio_block_get_dmabuf_fd` which can be used when doing zero-copy as well.

### DIFF
--- a/README_BUILD.md
+++ b/README_BUILD.md
@@ -109,6 +109,7 @@ Cmake Options       | Default | Description                                    |
 `WITH_IIOD_NETWORK` |  ON | Add network (TCP/IP) support                         |
 `WITH_IIOD_SERIAL`  |  ON | Add serial (UART) support                            |
 `WITH_IIOD_USBD`    |  ON | Add support for USB through FunctionFS within IIOD   |
+`WITH_IIOD_USB_DMABUF` | OFF | Enable DMABUF support on the USB stack            |
 `WITH_IIOD_V0_COMPAT` |  ON | Add support for Libiio v0.x protocol and clients |
 `WITH_LIBTINYIIOD`  | OFF | Build libtinyiiod                                    |
 `WITH_AIO`          |  ON | Build IIOD with async. I/O support                   |
--- a/block.c
+++ b/block.c
@@ -21,6 +21,8 @@ struct iio_block {
 
 	struct iio_task_token *token;
 	size_t bytes_used;
+
+	int dmabuf_fd;
 };
 
 struct iio_block *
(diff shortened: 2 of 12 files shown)


rgetz:
Do we need some CMake and ifdef so that things still build on old kernels (pre 5.19?) 

I think that is what is causing the centos 7 fails.
```[ 31%] Building C object CMakeFiles/iio.dir/local-dmabuf.c.o
/__w/1/s/local-dmabuf.c: In function 'local_create_dmabuf':
/__w/1/s/local-dmabuf.c:100:56: error: 'MFD_ALLOW_SEALING' undeclared (first use in this function)
  memfd = syscall(__NR_memfd_create, "/libiio-udmabuf", MFD_ALLOW_SEALING);
                                                        ^
/__w/1/s/local-dmabuf.c:100:56: note: each undeclared identifier is reported only once for each function it appears in
/__w/1/s/local-dmabuf.c:106:19: error: 'F_ADD_SEALS' undeclared (first use in this function)
  if (fcntl(memfd, F_ADD_SEALS, F_SEAL_SHRINK) < 0) {
                   ^
/__w/1/s/local-dmabuf.c:106:32: error: 'F_SEAL_SHRINK' undeclared (first use in this function)
  if (fcntl(memfd, F_ADD_SEALS, F_SEAL_SHRINK) < 0) {
                                ^
make[2]: *** [CMakeFiles/iio.dir/local-dmabuf.c.o] Error 1
make[1]: *** [CMakeFiles/iio.dir/all] Error 2
make: *** [all] Error 2
```

pcercuei:
Updated the PR; it now uses dma-heaps instead of udmabuf to create the DMABUF object.

pcercuei:
Rebased and added a fix so that the DMABUFs are correctly detached from USB after #1136.

pcercuei:
@mhennerich reminder to merge this :)

Note that I cannot update the branch anymore as I'm not in ADI's organization, so I cannot rebase it.

pcercuei:
Nice! @nunojsa don't hesitate to ping me if you have problems or questions.

nunojsa:
@pcercuei,

I finally got to do some testing with this with an ADI updated kernel. Things look good and working. Just some hick ups on the MMAP interface (I guess you're more focused on testing the new API :)). While I'm eager to remove that code from ADI kernel, we still need to support it for at least one release.

Anyways, first patch can be squashed in the fixup (just put it like this so it's easier for you to see the change) and the second I guess it needs to be in it's own patch.

pcercuei:
@nunojsa forgot to answer, sorry. LGTM!

cristina-suteu (approved):

nunojsa (approved):

cristina-suteu:
did another round of testing on this, everything seems good to go. 
will wait another day before merging, if there are any other issues, please let me know.

pcercuei:
🥳

rgetz:
I think we missed adding `WITH_IIOD_USB_DMABUF` to [`Cmake`](https://github.com/analogdevicesinc/libiio/blob/main/CMakeLists.txt#L696-L720) so someone knows what they are building... (and turning it on for CI/CD).

PR #1044, opened 2023-08-31
local-mmap: Fix block allocation

### LINKED ISSUE #1042: Error when opening tx buffer with libiio 0.25 on host and 1.0 on device
When I try to open a buffer with `buffer = iio_device_create_buffer(dev, 7680, false);` I get this error
```
ERROR: Open unlocked: -12
```

host version: 0.25 (git tag:b6028fde)
device version: 1.0 (git tag:v1.0)

It does not show the git hash when I do `iio_info -V` on the device, is that intended (tag v1.0 does not exist yet)?

pcercuei:
The algorithm for allocating new blocks was broken, as it was requesting an allocation of (nb_blocks + 1) blocks instead of a single block.

Fixes #1042.

### DIFF
--- a/local-mmap.c
+++ b/local-mmap.c
@@ -119,16 +119,11 @@ local_create_mmap_block(struct iio_buffer_pdata *pdata,
 		req.id = 0;
 		req.type = 0;
 		req.size = size;
-		req.count = priv->idx + 1;
+		req.count = 1;
 
 		ret = ioctl_nointr(pdata->fd, BLOCK_ALLOC_IOCTL, &req);
 		if (ret < 0)
 			goto out_free_priv;
-
-		if (req.count < priv->idx + 1) {
-			ret = -ENOMEM;
-			goto out_free_priv;
-		}
 
 		ppdata->nb_blocks++;
 	} else {


local-mmap.c
```diff
@@ -119,13 +119,13 @@ local_create_mmap_block(struct iio_buffer_pdata *pdata,
...
 		req.size = size;
-		req.count = priv->idx + 1;
+		req.count = 1;
 
 		ret = ioctl_nointr(pdata->fd, BLOCK_ALLOC_IOCTL, &req);
 		if (ret < 0)
 			goto out_free_priv;
 
-		if (req.count < priv->idx + 1) {
+		if (req.id != ppdata->nb_blocks) {
```
nunojsa, line 128: This is a subtle change that might deserve some mention in the commit message. Furthermore, I don't really think ENOMEM (not sure what would be either) is the best error code in here or even if we should care in doing any check. Looking at the kernel side, the only way for `req.id` not to be the older number of blocks is if we error out already on the `ioctl()` call so I'm not so sure this brings any value...

One check that could make sense before doing the `ioctl()` is `if (ppdata->nb_blocks == 64)` in which case we would not do any `ioctl()` as it's meaningless (and maybe print a warning message stating the maximum number of blocks is reached).

I know 64 is an hardcoded value but that value is also hardcoded in the kernel and I guess that is very unlikely to change as we don't care much in extending the mmap interface (in fact, as you know, we want to deprecate it)

nunojsa: > There's also no need to check if (ppdata->nb_blocks == 64)... a second time, because it's done right above this code.

Oh, I failed to see that...

nunojsa (approved):

PR #1047, opened 2023-09-01
Remove API function iio_device_identify_filename()

pcercuei:
This function can be implemented with other libiio functions. The
callers that need it should implement it themselves.

### DIFF
--- a/include/iio/iio.h
+++ b/include/iio/iio.h
@@ -1606,23 +1606,6 @@ iio_device_debug_attr_write_raw(const struct iio_device *dev,
 		 double: iio_device_debug_attr_write_double)(dev, attr, val)
 
 
-/** @brief Identify the channel or debug attribute corresponding to a filename
- * @param dev A pointer to an iio_device structure
- * @param filename A NULL-terminated string corresponding to the filename
- * @param chn A pointer to a pointer of an iio_channel structure. The pointed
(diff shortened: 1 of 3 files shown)


rgetz (approved):

rgetz:
looks good to me. Thanks

include/iio/iio.h
```diff
@@ -1606,23 +1606,6 @@ iio_device_debug_attr_write_raw(const struct iio_device *dev,
...
- * pointer will be set to point to the name of the attribute corresponding to
- * the filename.
- * @return On success, 0 is returned, and *chn and *attr are modified.
- * @return On error, a negative errno code is returned. *chn and *attr are not
- * modified. */
-__api __check_ret int iio_device_identify_filename(const struct iio_device *dev,
-		const char *filename, struct iio_channel **chn,
-		const char **attr);
-
-
```
nunojsa, line 1625: I guess this function is not that simple simple to deserve an helper. But it was added so, IMHO, we now have to leave with it (or at least give time - using deprecated or so - for users of this to remove it). So, NAK from my side as I do think breaking our users should be a no go...

(BTW, I'm also not a big fan of the deprecated __attribute__)

nunojsa (approved):

tfcollins (approved):

PR #1056, opened 2023-09-21
New IIO attributes API

pcercuei:
This set of patches change how attributes work in Libiio.

Before, each one of `iio_context`, `iio_device`, `iio_channel` had its own set of API functions to read / write their attributes, for a total of 50+ different API functions.

This was maybe slightly easier to use, but had some drawbacks; notably the high number of API functions, but also a lack of standardization of the accesses:
- context attributes did not have any read_bool / read_longlong / read_double functions;
- buffer attributes required the hardware buffer index to be specified.
This lack of standardization also meant that it was impossible for language bindings to wrap attributes into a common "attribute" object.

For these reasons, a new API is proposed.

Accessors are still present, but will now return an opaque pointer to an `iio_attr`:
```c
unsigned int iio_context_get_attrs_count(const struct iio_context *ctx);
const struct iio_attr * iio_context_get_attr(const struct iio_context *ctx, unsigned int idx);
const struct iio_attr * iio_context_find_attr(const struct iio_context *ctx, const char *name);

unsigned int iio_device_get_attrs_count(const struct iio_device *dev);
const struct iio_attr * iio_device_get_attr(const struct iio_device *dev, unsigned int idx);
const struct iio_attr * iio_device_find_attr(const struct iio_device *dev, const char *name);

unsigned int iio_device_get_debug_attrs_count(const struct iio_device *dev);
const struct iio_attr * iio_device_get_debug_attr(const struct iio_device *dev, unsigned int idx);
const struct iio_attr * iio_device_find_debug_attr(const struct iio_device *dev, const char *name);

unsigned int iio_channel_get_attrs_count(const struct iio_channel *chn);
const struct iio_attr * iio_channel_get_attr(const struct iio_channel *chn, unsigned int idx);
const struct iio_attr * iio_channel_find_attr(const struct iio_channel *chn, const char *name);
```

Note that the accessors for buffer attributes changed. They are now specific to an `iio_buffer` object, and not to an `iio_device` object as it used to be. The reason behind that, is that in a multi-buffer situation, different buffers will have different attributes.
```c
unsigned int iio_buffer_get_attrs_count(const struct iio_buffer *buf);
const struct iio_attr * iio_buffer_get_attr(const struct iio_buffer *buf, unsigned int idx);
const struct iio_attr * iio_buffer_find_attr(const struct iio_buffer *buf, const char *name);
```

Once you obtain the `iio_attr` object, you can get its name, filename and static value (only for context attributes) with these functions:
```c
const char * iio_attr_get_name(const struct iio_attr *attr);
const char * iio_attr_get_filename(const struct iio_attr *attr);
const char * iio_attr_get_static_value(const struct iio_attr *attr);
```

The functions to read / write attributes are now:
```c
ssize_t iio_attr_read_raw(const struct iio_attr *attr, char *dst, size_t len);
ssize_t iio_attr_write_raw(const struct iio_attr *attr, const void *src, size_t len);

int iio_attr_read_bool(const struct iio_attr *attr, bool *val);
int iio_attr_read_longlong(const struct iio_attr *attr, long long *val);
int iio_attr_read_double(const struct iio_attr *attr, double *val);

int iio_attr_write_string(const struct iio_attr *attr, const char *str);
int iio_attr_write_bool(const struct iio_attr *attr, bool val);
int iio_attr_write_longlong(const struct iio_attr *attr, long long val);
int iio_attr_write_double(const struct iio_attr *attr, double val);
```
You also have the `iio_attr_read` and `iio_attr_write` C11 generic macros which will automatically pick the right API function according to the type.

Emulating the old behaviour is very easy (except for buffer attributes - more on that below), and client applications who wish to keep using the old behaviour can create themselves wrappers like this:
```c
ssize_t channel_write_raw(const struct iio_channel *chn, const char *name, const void *src, size_t len)
{
   const struct iio_attr *attr = iio_channel_find_attr(ch, name);
   if (!attr)
       return -ENOENT;
   return iio_attr_write_raw(attr, src, len);
}
```

The most intrusive change is maybe the one with the buffer attributes, because now they can only be accessed when a `iio_buffer` has been created. Granted, in v1.0 API creating a `iio_buffer` does not automatically start the stream process like it did in v0.x, but it will still try to open the /dev node. If that fails, then the buffer attributes cannot be accessed, but that should be okay - you probably don't need to change buffer attributes if the buffer can't be created in the first place.

The problem is more with the v0.x compatibility layer; as emulating `iio_device_buffer_attr_read_raw` for instance would mean creating a "fake" `iio_buffer` just to be able to read its attributes. The compatibility layer (and IIOD) will now return an error code when an old application tries to read a buffer attribute before an iio_buffer has been created. This definitely affects the old "iio_info" for instance, but it *should* be fine as very few applications in v0.x used buffer attributes in the first place.

Feedback welcome.

### DIFF
--- a/utils/gen_code.c
+++ b/utils/gen_code.c
@@ -124,6 +124,7 @@ void gen_start(const char *gen_file)
 
 		fprintf(fd, "int main(int argc, char **argv)\n{\n"
 			"\tstruct iio_context *ctx;\n\tstruct iio_device *dev;\n\tstruct iio_channel *ch;\n"
+			"\tstruct iio_attr *attr;\n"
 			"\tconst char* val_str;\n\tssize_t ret;\n\tchar buf[256];\n\n");
 
 	} else if (lang == PYTHON_LANG) {
@@ -198,7 +199,8 @@ void gen_context_attr(const char *key)
 
 	if (lang == C_LANG) {
 		fprintf(fd, "\t/* Read IIO Context attribute and return result as string */\n");
-		fprintf(fd, "\tval_str = iio_context_get_attr_value(ctx, \"%s\");\n", key);
+		fprintf(fd, "\tattr = iio_context_find_attr(ctx, \"%s\");\n", key);
+		fprintf(fd, "\tval_str = iio_attr_get_static_value(ctx, attr);\n");
 		fprintf(fd, "\tprintf(\"%s : %%s\\n\", val_str);\n", key);
 	} else if (lang == PYTHON_LANG) {
 		fprintf(fd, "    # Read IIO Context attribute and return result as string\n");
@@ -255,7 +257,7 @@ void gen_context_timeout(unsigned int timeout_ms)
 }
 
 void gen_function(const char* prefix, const char* target,
-		const char* attr, const char* wbuf)
+		  const struct iio_attr *attr, const char *wbuf)
 {
 	char *rw = wbuf ? "write" : "read";
 
@@ -265,27 +267,31 @@ void gen_function(const char* prefix, const char* target,
 	if (lang == C_LANG) {
 		if (wbuf) {
 			fprintf(fd, "\t/* Write null terminated string to %s attribute: */\n", prefix);
-			fprintf(fd, "\tRET_ASSERT(ret = iio_%s_attr_write(\n"
-				    "\t\t\t%s, \"%s\", \"%s\"));\n",
-				prefix, target, attr, wbuf);
+			fprintf(fd, "\tattr = iio_%s_find_attr(%s, \"%s\");\n"
+				    "\tRET_ASSERT(ret = iio_attr_write_string(attr, \"%s\"));\n",
+				    prefix, target, iio_attr_get_name(attr), wbuf);
 		} else {
 			fprintf(fd, "\t/* Read IIO %s attribute, and put result in string */\n", prefix);
-			fprintf(fd, "\tRET_ASSERT(ret = iio_%s_attr_read(\n"
-				    "\t\t\t%s, \"%s\", buf, sizeof(buf)));\n",
-				prefix, target, attr);
+			fprintf(fd, "\tattr = iio_%s_find_attr(%s, \"%s\");\n"
+				    "\tRET_ASSERT(ret = iio_attr_read_raw(attr, buf, sizeof(buf)));\n",
+				    prefix, target, iio_attr_get_name(attr));
 		}
 		fprintf(fd, "\t/* For other types, use:\n");
-		fprintf(fd, "\t *  ret = iio_%s_attr_%s_bool(%s, \"%s\", v_bool);\n",
-			prefix, rw, target, attr);
-		fprintf(fd, "\t *  ret = iio_%s_attr_%s_double(%s, \"%s\", v_double);\n",
-			prefix, rw, target, attr);
-		fprintf(fd, "\t *  ret = iio_%s_attr_%s_longlong(%s, \"%s\", v_ll);\n",
-			prefix, rw, target, attr);
+		fprintf(fd, "\t *  attr = iio_%s_find_attr(%s, \"%s\");\n"
+			    "\t    ret = iio_attr_%s_bool(attr, v_bool);\n",
+			prefix, rw, target, iio_attr_get_name(attr));
+		fprintf(fd, "\t *  attr = iio_%s_find_attr(%s, \"%s\");\n"
+			    "\t    ret = iio_attr_%s_double(attr, v_double);\n",
+			prefix, rw, target, iio_attr_get_name(attr));
+		fprintf(fd, "\t *  attr = iio_%s_find_attr(%s, \"%s\");\n"
+			    "\t    ret = iio_attr_%s_longlong(attr, v_ll);\n",
+			prefix, rw, target, iio_attr_get_name(attr));
 		fprintf(fd, "\t *******************************************************************/\n");
 		if (wbuf) {
 			fprintf(fd, "\tprintf(\"Wrote %%zi bytes\\n\", ret);\n\n");
 		} else {
-			fprintf(fd, "\tprintf(\"%s : %%s\\n\", buf);\n\n", attr);
+			fprintf(fd, "\tprintf(\"%s : %%s\\n\", buf);\n\n",
+				iio_attr_get_name(attr));
 		}
 	} else if (lang == PYTHON_LANG) {
 		if (wbuf) {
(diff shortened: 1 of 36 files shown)


pcercuei:
Pinging @catkira and @sternmull for feedback :)

rgetz:
>The compatibility layer (and IIOD) will now return an error code when an old application tries to read a buffer attribute before an iio_buffer has been created. This definitely affects the old "iio_info" for instance, but it should be fine as very few applications in v0.x used buffer attributes in the first place.

let me check some other applications to see if this is true. I'm not sure...

sternmull:
I only looked at the comments here on GitHub and not at the code. Looks good to me! Happy to see the API becoming smaller and more elegant.

mhennerich (approved):
LGTM

pcercuei:
@rgetz, final words?

catkira:
I think it's a big improvement. Removing +400 lines of code at the same time is also very good 👍

rgetz:
> This definitely affects the old "iio_info" for instance,

I didn't see anything that did this - but it will cause issues with the compatibility layer.

two questions:
- Are you "fixing" iio_info info, so that the output is the same (ie. it opens a buffer before checking the attributes?)
- Are you giving a specific error number in the compatibility layer, so people can catch this easier in their old apps?

Thanks
-Robin

pcercuei:
>     * Are you "fixing" iio_info info, so that the output is the same (ie. it opens a buffer before checking the attributes?)

Yes, that's what the new iio_info does.

>     * Are you giving a specific error number in the compatibility layer, so people can catch this easier in their old apps?

Right now it's ENOENT (no such file or directory) as in "the attribute does not exist" which I agree is not the best. I can make it return `ENOSYS` which in Libiio always corresponds to an unimplemented feature.

utils/iio_attr.c
```diff
@@ -129,12 +132,12 @@ static int dump_device_attributes(const struct iio_device *dev,
 		}
 	}
 	if (wbuf) {
-		gen_function("device", "dev", attr, wbuf);
-		ret = iio_device_attr_write_string(dev, attr, wbuf);
+		gen_function(type, var, attr, wbuf);
+		ret = iio_attr_write_string(attr, wbuf);
 		if (ret > 0) {
 			if (quiet == ATTR_VERBOSE)
 				printf("wrote %li bytes to %s\n", (long)ret, attr);
```
tfcollins, line 139: attr needs to be wrapped with  iio_attr_get_name

utils/gen_code.c
```diff
@@ -265,21 +267,24 @@ void gen_function(const char* prefix, const char* target,
 	if (lang == C_LANG) {
 		if (wbuf) {
 			fprintf(fd, "\t/* Write null terminated string to %s attribute: */\n", prefix);
-			fprintf(fd, "\tRET_ASSERT(ret = iio_%s_attr_write(\n"
-				    "\t\t\t%s, \"%s\", \"%s\"));\n",
-				prefix, target, attr, wbuf);
+			fprintf(fd, "\tattr = iio_%s_find_attr(%s, \"%s\");\n"
+				    "\tRET_ASSERT(ret = iio_attr_write_string(attr, \"%s\"));\n",
+				    prefix, target, attr, wbuf);
```
tfcollins, line 272: attr needs to be wrapped with  iio_attr_get_name

utils/gen_code.c
```diff
@@ -265,21 +267,24 @@ void gen_function(const char* prefix, const char* target,
...
 		} else {
 			fprintf(fd, "\t/* Read IIO %s attribute, and put result in string */\n", prefix);
-			fprintf(fd, "\tRET_ASSERT(ret = iio_%s_attr_read(\n"
-				    "\t\t\t%s, \"%s\", buf, sizeof(buf)));\n",
-				prefix, target, attr);
+			fprintf(fd, "\tattr = iio_%s_find_attr(%s, \"%s\");\n"
+				    "\tRET_ASSERT(ret = iio_attr_read_raw(attr, buf, sizeof(buf)));\n",
+				    prefix, target, attr);
```
tfcollins, line 277: attr needs to be wrapped with  iio_attr_get_name

utils/gen_code.c
```diff
@@ -265,21 +267,24 @@ void gen_function(const char* prefix, const char* target,
...
+			    "\t    ret = iio_attr_%s_bool(attr, v_bool);\n",
 			prefix, rw, target, attr);
-		fprintf(fd, "\t *  ret = iio_%s_attr_%s_double(%s, \"%s\", v_double);\n",
+		fprintf(fd, "\t *  attr = iio_%s_find_attr(%s, \"%s\");\n"
+			    "\t    ret = iio_attr_%s_double(attr, v_double);\n",
 			prefix, rw, target, attr);
-		fprintf(fd, "\t *  ret = iio_%s_attr_%s_longlong(%s, \"%s\", v_ll);\n",
+		fprintf(fd, "\t *  attr = iio_%s_find_attr(%s, \"%s\");\n"
+			    "\t    ret = iio_attr_%s_longlong(attr, v_ll);\n",
 			prefix, rw, target, attr);
```
tfcollins, line 288: attr needs to be wrapped with  iio_attr_get_name

utils/gen_code.c
```diff
@@ -265,21 +267,24 @@ void gen_function(const char* prefix, const char* target,
...
 		}
 		fprintf(fd, "\t/* For other types, use:\n");
-		fprintf(fd, "\t *  ret = iio_%s_attr_%s_bool(%s, \"%s\", v_bool);\n",
+		fprintf(fd, "\t *  attr = iio_%s_find_attr(%s, \"%s\");\n"
+			    "\t    ret = iio_attr_%s_bool(attr, v_bool);\n",
 			prefix, rw, target, attr);
-		fprintf(fd, "\t *  ret = iio_%s_attr_%s_double(%s, \"%s\", v_double);\n",
+		fprintf(fd, "\t *  attr = iio_%s_find_attr(%s, \"%s\");\n"
+			    "\t    ret = iio_attr_%s_double(attr, v_double);\n",
 			prefix, rw, target, attr);
```
tfcollins, line 285: attr needs to be wrapped with  iio_attr_get_name

tfcollins (changes requested):
Minor issues with the utils and use of attr (I believe)

utils/iio_attr.c
pcercuei: Yes, it's strange that GCC doesn't warn about that...

tfcollins (approved):

PR #1064, opened 2023-10-05
Enhance blocks API

pcercuei:
This is a subset of #1035, which does not depend on the DMABUF IIO interface being merged into the upstream Linux kernel.

I create a separate PR so that these can be merged sooner.


rgetz (approved):
Did not test - but makes sense.

mhennerich (approved):

PR #1070, opened 2023-10-12
CI/build_win.ps1: Fix cmake using the wrong python version

bia1708:
Added a flag for cmake which sets the python version to the one used by the system (configured in azure-pipelines.yml).

### DIFF
--- a/CI/build_win.ps1
+++ b/CI/build_win.ps1
@@ -11,7 +11,8 @@ mkdir build-x64
 cp .\libiio.iss.cmakein .\build-x64
 cd build-x64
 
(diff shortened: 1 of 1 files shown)


SRaus (approved):

SRaus (approved):

rgetz:
your commit message says what you did - but not why.

bia1708:
@rgetz the purpose of this commit was to fix the CIs. The official Azure windows images got updated, which introduced a bug in our Azure Pipelines CIs. The new version of the windows images contain a faulty installation of python 3.12, introduced by [this PR](https://github.com/actions/runner-images/pull/8451/checks?check_run_id=17360712485). I figured out that cmake was not using the version of python that is configured globally in azure_pipelines.yml, therefore I made this change.
I am expecting that the next version of the windows images will not contain said bug ([this PR](https://github.com/actions/runner-images/pull/8510) reverts those changes, but it hasn't been included in a release yet), so I didn't find it necessary to include all of these details in the commit description. Nevertheless, it is good practice to have cmake use the python version defined globally (as it will be a lot easier to debug in the future).

PR #1079, opened 2023-10-25
Cherry pick commits from libiio-v0 branch

AAndrisa:


pcercuei:
Why not just merge `libiio-v0` into `next_stable`?

SRaus:
> Why not just merge `libiio-v0` into `next_stable`?

Hi Paul.
The 'next_stable' and 'libiio-v0' branches started to diverge. Previous commits in 'next_stable' where added also with cherry-picks from other branches. Since we are so close to Kuiper release, we just don't want to do mass cherry-picking or rebase, but to add only the strictly necessary commits. 
In this case, there are only CI changes, so they won't broke anything. but in general, we are trying to follow the above described flow.

SRaus (approved):

pcercuei:
@SRaus OK. I think you could have merged, as everything that goes into `libiio-v0` now are just fixes. The problem with cherry-picking individual patches is that it may be a pain to switch `next_stable` to the future v1.0 release later if it diverges too much.

PR #1081, opened 2023-11-06
Add support for IIO events

pcercuei:
Add new API functions to allow receiving IIO events.

- `iio_device_create_event_stream` returns a `iio_event_stream` object, which can then be used to read IIO events;
- `iio_event_stream_destroy` to destroy a previously created stream;
- `iio_event_stream_read` to read an `iio_event`, either non-blocking (in which case -EAGAIN is returned if there is no event) or blocking (if you want to run it in a thread with a callback mechanism).

The `iio_event` is the exact same as the `iio_event_data` object from Linux' `<linux/iio/events.h>` so the data can be extracted similarly. A few functions were added to make it easier to decode events: `iio_event_get_type`, `iio_event_get_direction`, `iio_event_get_channel`.

Events can be configured (to set thresholds etc.) and enabled/disabled using the standard IIO attribute mechanism, as their settings now appear as standard iio_channel attributes.


rgetz:
don't we want to list that events are possible in iio_info?

pcercuei:
I think I will update iio_info in a separate PR. There are a few things I want to do there anyway, notably handle attribute reads errors better, and add some color.

local.c
```diff
@@ -1541,6 +1579,92 @@ int local_dequeue_block(struct iio_block_pdata *pdata, bool nonblock)
...
+			.events = POLLIN,
+		}
+	};
+	int ret, timeout_rel = nonblock ? 0 : -1;
+
+	do {
+		ret = poll(pollfd, 2, timeout_rel);
+	} while (ret == -1 && errno == EINTR);
+
+	if ((pollfd[1].revents & (POLLIN | POLLNVAL)))
```
dlech, line 1645: Should this be moved after the `ret` checks below?

dlech: Also, don't we need to read from the eventfd to reset it?

iio-private.h
```diff
@@ -34,7 +34,7 @@
 #define is_little_endian() (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
 #endif
 
-#define BIT(x) (1 << (x))
+#define BIT(x) (1ull << (x))
```
dlech, line 37: Seems odd to make this 64-bit but not the other `BIT` macros.

events.c
```diff
@@ -0,0 +1,139 @@
...
+			iio_channel_get_id(chn));
+	} else {
+		dev_dbg(dev, "Unable to find channel for event\n");
+	}
+
+	return chn;
+}
+
+struct iio_event_stream *
+iio_device_create_event_stream(const struct iio_device *dev)
```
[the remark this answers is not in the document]
pcercuei: That wouldn't be consistent with what we already have, e.g. `iio_device_create_buffer`, `iio_buffer_create_block`, `iio_buffer_create_stream`.

include/iio/iio.h
```diff
@@ -1290,6 +1295,81 @@ static inline bool iio_device_is_hwmon(const struct iio_device *dev)
...
+};
+
+/**
+ * @brief Get the type of a given IIO event
+ * @param event A pointer to an iio_event structure
+ * @return An enum iio_event_type. */
+static inline enum iio_event_type
+iio_event_get_type(const struct iio_event *event)
+{
+	return (enum iio_event_type)((event->id >> 56) & 0xff);
```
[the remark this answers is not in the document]
pcercuei: Ok, will do.

local.c
pcercuei: We don't actually want to reset it; otherwise the `local_read_event` function would only fail the first time.

As for the order - the `pollfd[1].revents` check is done before the check for `ret < 0` just to be consistent with what's already in `buffer_check_ready`. It shouldn't matter much anyway.

iio-private.h
pcercuei: I only make the `BIT` macro able to create 64-bit masks, the `BIT_MASK` and `BIT_WORD` still need to work on 32-bit words. I *could* make them work with 64-bit words, but that means changing the `iio_channels_mask` struct, which is a bit outside this PR.

pcercuei: Thanks for reviewing btw.

events.c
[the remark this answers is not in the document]
dlech: fair enough

local.c
```diff
@@ -1541,6 +1579,92 @@ int local_dequeue_block(struct iio_block_pdata *pdata, bool nonblock)
...
+
+	return pdata;
+}
+
+static void local_close_events_fd(struct iio_event_stream_pdata *pdata)
+{
+	uint64_t event = 1;
+	int ret;
+
+	ret = write(pdata->cancel_fd, &event, sizeof(event));
```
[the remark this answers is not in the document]
dlech: ah yes, of course

local.c
dlech: Looking back, I see where the docs say that you have to call `iio_event_stream_destroy()` to cancel the read, so makes sense now that there is no reset to read again..

bindings/python/iio.py
```diff
@@ -1081,6 +1146,54 @@ def __next__(self):
...
+
+        chn1 = None
+        chn2 = None
+
+        try:
+            dev = self._parent
+            chn1 = Channel(dev, _ev_get_channel(_byref(event), dev._device, False))
+            chn2 = Channel(dev, _ev_get_channel(_byref(event), dev._device, True))
+        except Exception:
+            pass
```
tfcollins, line 1193: Need to be more explicit for the exception

pcercuei: I changed it to `OSError`.

rgetz:
@tfcollins should look at the Codacy issues, and point out any that should be fixed before a merge - since they are all Python.

pcercuei:
> @tfcollins should look at the Codacy issues, and point out any that should be fixed before a merge - since they are all Python.

I fixed some of them, some other I will leave as-is. The "parent `__init__` not called" one because the children class' `__init__` just returns an exception (as we don't want to permit direct instantiation), the super() one to be consistent with the rest of the script file.

mhennerich (approved):
LGTM, Thanks!

PR #1084, opened 2023-11-15
Add offset to iio_data_format, rework scale

pcercuei:
Add an `offset` field in the `iio_data_format` structure alongside the "scale". Make sure those two are initialized properly for all channels, not just the scan-element ones.


include/iio/iio.h
```diff
@@ -1328,6 +1328,10 @@ struct iio_data_format {
 
 	/** @brief Number of times length repeats (added in v0.8) */
 	unsigned int repeat;
+
+	/** @brief Contains a value to be added to the raw sample before
+	 *  applying the scale. */
+	double offset;
```
nunojsa, line 1334: Just a side note... `structs` in header files are always nice to get some ABI breakage down the road. I wonder if we shouldn't have some `reserved` bits in here? At least for when v1.0 is out...

context.c
```diff
@@ -398,6 +398,52 @@ const struct iio_backend * const iio_backends[] = {
...
+static int iio_context_update_scale_offset(struct iio_context *ctx)
+{
+	const struct iio_attr *attr;
+	struct iio_channel *chn;
+	struct iio_device *dev;
+	unsigned int i, j;
+	int err;
+
+	if (!ctx->ops->read_attr) {
+		/* read_attr callback not implemented - nothing to do. */
```
nunojsa, line 410: nit: don't think the comment adds much value. The code is very self explanatory...

context.c
```diff
@@ -398,6 +398,52 @@ const struct iio_backend * const iio_backends[] = {
...
+				chn->format.with_scale = true;
+			}
+
+			attr = iio_channel_find_attr(chn, "offset");
+			if (attr) {
+				err = iio_attr_read_double(attr,
+							   &chn->format.offset);
+				if (err) {
+					chn_perror(chn, err, "Unable to read offset");
+					return err;
```
nunojsa, line 438: didn't checked the code so just sanity checking... There's really no `with_offset` like in the scale case?

nunojsa (approved):
Just some nits that you ignore if you want...

context.c
pcercuei: Well, there's no need, having no offset is the same as having offset == 0.0 (just like having no scale is the same as having scale == 1.0, but it's too late for that).

include/iio/iio.h
pcercuei: Yeah I was thinking that this is not a problem because old applications won't try to access the "offset" field. Apps built with Libiio > 1.0 may try to access new fields that are yet to be added, but we never really cared about forward compatibility.

With that said - I can add some padding, it's not that much more work.

nunojsa: Yeah, if the new fields are always added in the end of the struct but we will always leave the door open for subtle things and like you said... it's not much work at all

pcercuei: Done, as a separate commit. Thanks for the review btw.

PR #1085, opened 2023-11-16
Fix compatibility layer

pcercuei:
After the new attributes API was introduced, the compatibility layer started to fail, but only on non-local backends.

This was caused by local Libiio symbols being incorrectly linked to the compatibility layer. The call to `iio_channel_get_attr` in `iiod_client_read_attr_new` would resolve to the compatibility layer's `iio_channel_get_attr`, which returns a string instead of a `struct iio_attr *`, and caused all kinds of weird problems further down.

This happened because by default the linker will search for any symbol in the global symbol table, instead of searching in priority in the shared library that contains the code. Since the v0.25 "iio_info" (used in this test) would call `iio_channel_get_attr` before calling `iio_channel_attr_read`, the compat layer's `iio_channel_get_attr` ended up in the global symbol table, and was being chosen as the symbol to be called from the v1.0 Libiio.

Fix this by telling the linker to bind Libiio's references to the definitions within Libiio, if present, so that those have priority over the symbols found in the global symbol table.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -646,6 +646,7 @@ string(REPLACE ";" "," LIBIIO_SCAN_BACKENDS "${LIBIIO_SCAN_BACKENDS}")
 
 configure_file(iio-config.h.cmakein ${CMAKE_CURRENT_BINARY_DIR}/iio-config.h @ONLY)
 
+toggle_iio_feature("${LIBIIO_COMPAT}" compat)
 toggle_iio_feature("${WITH_XML_BACKEND}" xml)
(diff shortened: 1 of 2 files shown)


nunojsa (approved):

pcercuei:
It seems to fail to build on OSX, however it is using Clang 14 and it was compiling fine for me with Clang 17. Maybe we can try updating the Clang compiler on OSX.

pcercuei:
Updated the CMake script to not pass -Bsymbolic on OSX, since OSX already has the behaviour we want by default.

PR #1091, opened 2023-11-30
WIP Add test infrastructure

tfcollins:
The PR adds core infrastructure to introduce unit tests. It adds
- CMake + CTest pieces for testing targets
- Basic example test
- Coverage target for LCOV based testing
- GH action to run non-hardware tests


tfcollins:
> You need to define
> 
> ```
>  ./CMakeLists.txt:option(TESTS "Enabling unit testing" OFF)
> ./CMakeLists.txt:option(TESTS_DEBUG "Enable test debugging" OFF)
> ./CMakeLists.txt:option(TESTS_COVERAGE "Enable coverage tracing when testing" OFF)
> ```
> 
> in the `README_BUILD" file.

Added

tfcollins:
> Just a question about this unit testing framework.
> 
> Why does GitHub Action was chosen?
> 
> Do we have in plan to add local runners for those unit tests? If not, we can check here what are the costs https://confluence.analog.com/display/DEVTOOLS/Github+Actions%3A+Usage+and+Cost (*The page was not updated since a while - now we are a bit over the "free limit")
> 
> In other repos, we have implemented an extra step in Azure Pipeline for unit tests - still docker containers on cloud, any OS, fit in the existing payment plan. Check the [example for No-OS](https://dev.azure.com/AnalogDevices/no-OS/_build/results?buildId=37211&view=logs&j=739e1b7c-752b-5dd3-b8f8-24f95e240c64).

@SRaus GHA is the general direction of the company and even Microsoft. Its better integrated and doesn't require multiple logins. As for costs you missed the top paragraph of that page:

https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

_GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the account's plan. Any usage beyond the included amounts is controlled by spending limits._

Other reference: https://github.com/orgs/community/discussions/70492


Since libiio is public, GHA is free.

As for self hosted runners, this shouldn't be necessary for libiio as it doesn't require special tools (MATLAB, Vivado, Quartus) to build. Hardware testing is a different story and isn't even covered by Azure Pipelines.

rgetz:
>As for self hosted runners, this shouldn't be necessary for libiio as it doesn't require special tools (MATLAB, Vivado, Quartus) to build. Hardware testing is a different story and isn't even covered by Azure Pipelines.

I think this is why it's super important to integrate/squash https://github.com/analogdevicesinc/iio-emu into the main libiio repo.

Otherwise testing is going to continue to be pretty adhoc, and hardware testing is nearly non-existent (random manual tests that devs do, and random tests that people do for M2k and Pluto).

PR #1092, opened 2023-12-01
Serial backend fixes

pcercuei:
- Make sure the serial backend will properly exit and not lock up when iio_context_destroy is called;
- Handle clients sending the BINARY command in ASCII when IIOD is already in binary mode (which happens when a new client connects)
- Fix iio_context_clone() in the compat layer, which should return -ENOSYS with the serial backend, as only one context can be created at the same time. This fixes iio-osc running on top of the compat layer when using the serial backend.

### DIFF
--- a/serial.c
+++ b/serial.c
@@ -262,6 +262,18 @@ static int serial_enable_buffer(struct iio_buffer_pdata *buf,
 	return iiod_client_enable_buffer(buf->pdata, nb_samples, enable);
 }
 
+static ssize_t
+serial_readbuf(struct iio_buffer_pdata *buf, void *dst, size_t len)
+{
+	return iiod_client_readbuf(buf->pdata, dst, len);
+}
+
+static ssize_t
+serial_writebuf(struct iio_buffer_pdata *buf, const void *src, size_t len)
+{
+	return iiod_client_writebuf(buf->pdata, src, len);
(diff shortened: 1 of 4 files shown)


serial.c
```diff
@@ -145,21 +152,44 @@ static ssize_t serial_write_data(struct iiod_client_pdata *io_data,
...
+#else
+	const struct timespec ts = {
+		/* One millisecond */
+		.tv_nsec = 1 * 1000 * 1000,
+	};
+
+	nanosleep(&ts, NULL);
+#endif
+}
+
```
rgetz, line 168: should this be in utils.c rather than here? (I guess it's a single use case?)

rgetz (approved):

serial.c
pcercuei: utils.c is for the high-level API, we'd have to export it as a public symbol (as the serial backend can be a module) and I prefer to avoid that as much as possible.

rgetz: makes sense.

tfcollins (approved):

pcercuei:
Added a fix for a lockup that only @buha could trigger.

buha (approved):
cool, the last commit does fix the issue of iio_info hanging after printing all the info

PR #1108, opened 2023-12-19
iio.py: fix Trigger instantiation

### LINKED ISSUE #1107: Trigger instantiation missing argument in iio.Device
See: https://github.com/analogdevicesinc/libiio/blob/1b1258d8d33025e2de5581b472fa8ca6e0c7c26e/bindings/python/iio.py#L1349

There should be two arguments passed here, but only the trigger device (value.contents) is passed. Without it, when trying to check a device's trigger property, you get something like:

```
>>> adc._ctrl.trigger
  File "/home/baylibre/git/pyadi-iio/venv/lib/python3.10/site-packages/iio.py", line 1281, in _get_trigger
    trig = Trigger(value.contents)
TypeError: Trigger.__init__() missing 1 required positional argument: '_device'
```

threexc:
Trigger() is being passed device contents, but no context, and therefore fails like so:

|>>> adc._ctrl.trigger
|Traceback (most recent call last):
|  File "<stdin>", line 1, in <module>
|  File "/home/baylibre/git/pyadi-iio/venv/lib/python3.10/site-packages/iio.py", line 1281, in _get_trigger
|    trig = Trigger(value.contents)
|TypeError: Trigger.__init__() missing 1 required positional argument: '_device'

Pass self.ctx as the first argument so that the trigger is properly instantiated.

Fixes #1107.

### DIFF
--- a/bindings/python/iio.py
+++ b/bindings/python/iio.py
@@ -1346,7 +1346,7 @@ class Device(_DeviceOrTrigger):
     def _get_trigger(self):
         value = _DevicePtr()
         _d_get_trigger(self._device, _byref(value))
-        trig = Trigger(value.contents)
+        trig = Trigger(self.ctx, value.contents)
 
         for dev in self.ctx.devices:
             if trig.id == dev.id:


pcercuei:
Looks valid. Thanks for the fix!

pcercuei (approved):

Issue #1046, opened 2023-08-31
C++ bindings are broken


avalerobiot:
I am trying to compile libiio following the instructions in the README_BUILD file and I get the same error both in MacOs and in Linux Centos:

Fixup Current symbolic link
[ 56%] Built target iio-compat
[ 59%] Linking C shared library iio.framework/iio
[ 59%] Built target iio
[ 65%] Building CXX object bindings/cpp/CMakeFiles/iiopp-enum.dir/examples/iiopp-enum.cpp.o
[ 65%] Building C object tests/CMakeFiles/iio_tests_helper.dir/gen_code.c.o
[ 65%] Building C object tests/CMakeFiles/iio_tests_helper.dir/iio_common.c.o
In file included from libiio/bindings/cpp/examples/iiopp-enum.cpp:12:
libiio/bindings/cpp/./iiopp.h:17:10: fatal error: 'iio.h' file not found
#include <iio.h>
         ^~~~~~~
[ 68%] Linking C static library libiio_tests_helper.a
[ 68%] Built target iio_tests_helper
[ 72%] Building C object tests/CMakeFiles/iio_attr.dir/iio_attr.c.o
[ 72%] Building C object tests/CMakeFiles/iio_info.dir/iio_info.c.o
[ 75%] Building C object tests/CMakeFiles/iio_stresstest.dir/iio_stresstest.c.o
[ 77%] Building C object tests/CMakeFiles/iio_rwdev.dir/iio_rwdev.c.o
[ 79%] Building C object tests/CMakeFiles/iio_genxml.dir/iio_genxml.c.o
[ 81%] Building C object tests/CMakeFiles/iio_reg.dir/iio_reg.c.o
[ 84%] Linking C executable iio_genxml
[ 86%] Linking C executable iio_info
[ 88%] Linking C executable iio_reg
[ 90%] Linking C executable iio_rwdev

Am I missing something?
Thanks!

rgetz:
I don't know if @sternmull wants to have a look at this one?

pcercuei:
I'm working on it.

illiw99:
Any updates on this? Encountered the same problem!

pcercuei:
Still not done - it gets pushed down the priority list constantly.

pcercuei:
The updated C++ bindings are in the main branch now, closing.

PR #1112, opened 2023-12-23
C++ API updated for new C-API

sternmull:
This updates iiopp.h for the latest changes of the C-API. It should fix issue #1046.

But be warned: I currently have no system with iio devices available for testing. So this is completely untested code!

The small example compiles and in general the code should be easy to read and review. I hope i didn't introduce any bugs. Please keep that in mind and review and test it carefully.


rgetz:
since: the cpp bindings are off by default - https://github.com/analogdevicesinc/libiio/blob/61fdbe78c298570bbc0c9e9ec3ce1a7c1fa3742e/CMakeLists.txt#L651C1-L651C48 

now things are working - we should turn them on for CI where we want to...

likely here for linux: https://github.com/analogdevicesinc/libiio/blob/61fdbe78c298570bbc0c9e9ec3ce1a7c1fa3742e/azure-pipelines.yml#L75

and here for mac: https://github.com/analogdevicesinc/libiio/blob/61fdbe78c298570bbc0c9e9ec3ce1a7c1fa3742e/azure-pipelines.yml#L337C4-L337C4

and here for windows: https://github.com/analogdevicesinc/libiio/blob/61fdbe78c298570bbc0c9e9ec3ce1a7c1fa3742e/CI/build_win.ps1#L14

and here for linux on ARM: https://github.com/analogdevicesinc/libiio/blob/61fdbe78c298570bbc0c9e9ec3ce1a7c1fa3742e/CI/azure/ci-ubuntu.sh

that way everything will get at least build tested on CI...

sternmull:
Static code analysis passes. The Linux builds fail due to a missing C++ compiler. The windows build fails because boost is not installed.
I don't know how to fix that. I think the Linux CI had a C++ compiler back when I added the bindings. Back then the C++ bindings where not activated for the windows CI-build.

Maybe someone who has experience with this CI-stuff can take a look at it?

rgetz:
Centos fails since `Target "iiopp-enum" requires the language dialect "CXX17" ` I think it's too old to support C++17... so turning that off makes sense. I think the same for Fedora34. Maybe @tfcollins knows more when he gets back in Jan.

If BOOST is a requirement - we should check for it in the CMAKE, and tell people to install it... something like:

```
  find_package(Boost QUIET NO_MODULE)

  if (Boost_FOUND)
    message(STATUS "Boost ${Boost_FIND_VERSION} found.")
    if (Boost_FIND_COMPONENTS)
      message(STATUS "Found Boost components:  ${Boost_FIND_COMPONENTS}")
    endif()
  else()
    message(SEND_ERROR "You need to install BOOST, or turn off C++ bindings")
  endif()
```

If you need specific modules - add them rather than `NO_MODULE`.

I think Travis can add BOOST to the docker image in Jan.

rgetz:
So, boost is not a hard requirement for windows...

The [code](https://github.com/analogdevicesinc/libiio/blob/main/bindings/cpp/iiopp.h#L20-L24): 
```
#if __cplusplus < 201703L
#include <boost/optional.hpp>
#else
#include <optional>
#endif
```

However - Visual Studio only fills out the `__cplusplus` var if you tell it to. (according to the [doc](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170), we need to add `/Zc:__cplusplus` compiler flag to enable the `__cplusplus` macro). I think that is the actual solution...

With maybe some more CMAKE to see if it needs to check for boost...

sternmull:
> So, boost is not a hard requirement for windows...

Yes. Actually it is not required with VS 2017 and later. Fixed that and a few minor problems (warnings that showed up with MS build).

rgetz:
So, then we are left with:

LinuxBuilds fedora34:
```
-- The CXX compiler identification is unknown
CMake Error at bindings/cpp/CMakeLists.txt:3 (project):
  No CMAKE_CXX_COMPILER could be found.
```
debug : Install a C++ compiler in fedora34 image. I think @tfcollins needs to do this.


LinuxBuilds centos_7:
```
CMake Error in bindings/cpp/CMakeLists.txt:
  Target "iiopp-enum" requires the language dialect "CXX17" (with compiler
  extensions).  But the current compiler "GNU" does not support this, or
  CMake does not know the flags to enable it.
```

```
-- The C compiler identification is GNU 4.8.5
```
which was released June 23, 2015, before the cpp 2017 spec was written. Either figure out how to make it work without, and have BOOST installed, or (my suggestion) just turn it off in the `azure-pipelines.yml` file - there is a seperate line for Centos already...

pcercuei:
@sternmull thanks a lot! I'll have a proper look next week.

tfcollins:
Container updated https://github.com/sdgtt/Dockerfiles/commit/5ade2941e7a606bbfffcd3195d1a628af7611f64  Will be ready in 30 minutes

rgetz:
@tfcollins - thanks - can someone with admin rights re-start the fedora build? or @sternmull can you squash things and push so things re-build?

pcercuei:
All checks pass now.

bindings/cpp/iiopp.h
```diff
@@ -309,221 +389,193 @@ class Channel
...
+class Stream
+{
+    iio_stream * const p;
+public:
+
+    Stream() = delete;
+    Stream(iio_stream * s) : p(s){}
+    operator iio_stream * () const {return p;}
+
+    Block next_block() {return const_cast<iio_block *>(impl::check(iio_stream_get_next_block(p), "iio_stream_get_next_block")); }
```
pcercuei, line 440: Won't the iio_block be deleted when it gets out of scope?

sternmull: No, the Block-class does not manage the lifetime of the referenced object. It is just a shallow wrapper for calling its functions.
See the comment at the top of the header. This should explain how the C++ classes manage resources.

pcercuei: Ok, thanks.

bindings/cpp/iiopp.h
```diff
@@ -309,221 +389,193 @@ class Channel
...
+    Attr find_attr(cstr name) {return iio_buffer_find_attr(p, name);}
     void set_data(void * data){iio_buffer_set_data(p, data);}
-    void * start() {return iio_buffer_start(p);}
-    ptrdiff_t step() const {return iio_buffer_step(p);}
+    void * data() {return iio_buffer_get_data(p);}
+    void cancel() {iio_buffer_cancel(p);}
+    void enable() {impl::check(iio_buffer_enable(p), "iio_buffer_enable");}
+    void disable() {impl::check(iio_buffer_disable(p), "iio_buffer_disable");}
+    ChannelsMask channels_mask() {return iio_buffer_get_channels_mask(p);}
+    BlockPtr crate_block(size_t size) { return BlockPtr{impl::check(iio_buffer_create_block(p, size), "iio_buffer_create_block")}; }
```
pcercuei, line 504: `crate_block` -> `create_block`

sternmull: Oops. Committed a fix.

pcercuei: Can you merge it into the first commit maybe?

I'll also need you to sign all the commits (currently only the first one is signed).

sternmull:
Done

pcercuei (approved):

pcercuei:
Merged, thanks a lot!

PR #1118, opened 2024-01-11
local: Avoid keeping buffer FD opened at all times

### LINKED ISSUE #1111: iio_device_create_buffer() is broken on local
commit aba77ca961414bde29c9856ae3f6797bc6eab9c0 causes `iio_device_create_buffer()` to return `Device or resource busy (16)` when I run it on local.

pcercuei:
The problem of leaving it opened at all times is that other instances of Libiio cannot be ran in parallel.

Fixes #1111.

### DIFF
--- a/local.c
+++ b/local.c
@@ -15,6 +15,7 @@
 #include <dirent.h>
 #include <errno.h>
 #include <iio/iio-debug.h>
+#include <iio/iio-lock.h>
 #include <limits.h>
 #include <poll.h>
 #include <stdbool.h>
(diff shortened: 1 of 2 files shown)


mhennerich (approved):

local.c
```diff
@@ -117,6 +122,10 @@ static void local_shutdown(struct iio_context *ctx)
 			close(dev->pdata->fd);
 		local_free_pdata(dev);
 	}
+
+	if (ctx->pdata && ctx->pdata->lock)
```
nunojsa, line 126: do we really need the condition?

nunojsa (approved):
minor comment that likely does not matter

local.c
pcercuei: Yes, because local_create_context() can fail before the pdata is allocated or the mutex is created. The error path calls iio_context_destroy() which in turn calls local_shutdown().

PR #1120, opened 2024-01-12
Better support for enqueue/dequeue errors

### LINKED ISSUE #1109: cannot recover when iio_block_dequeue() timeouts
I think we briefly talked about this already.

When a call of `iio_block_dequeue()` with nonblock set to false timeouts, the return error is `ETIMEDOUT`. This is fine. What I would expect is that the next call to `iio_block_dequeue()` then tries again for the timeout period that has been set with `iio_context_set_timeout()`. However what happens is that all calls to `iio_block_dequeue()` now immediately return with error code `EPERM` even when new blocks are available to dequeue. 

Can this be changed so that a read timeout is recoverable without having to tear down the whole connection and start everything from new? The current behavior prevents my data link application from doing fast reconnects in case of a connection loss.

pcercuei:
Given that the IIOD protocol has a single command for enqueue + dequeue, in case of an error we need to have a way to know which operation failed, so that the client can retry without having to tear down the whole IIO buffer.

Fixes #1109.


pcercuei:
The Codacy report looks completely bogus to me, or do I not know C?

mhennerich (approved):

PR #1121, opened 2024-01-12
Misc. fixes

pcercuei:
Fix some out-of-scope references, memory leaks, use-after-free, among other things.

### DIFF
--- a/context.c
+++ b/context.c
@@ -546,7 +546,7 @@ iio_create_context_from_xml(const struct iio_context_params *params,
 			    const char **ctx_values, unsigned int nb_ctx_attrs)
 {
 	struct iio_context *ctx;
-	char *new_description;
+	char *new_description = NULL;
 	unsigned int i;
 	ssize_t len;
 	int ret;
@@ -596,11 +596,10 @@ iio_create_context_from_xml(const struct iio_context_params *params,
 			prm_err(params, "Unable to alloc memory\n");
 			goto err_context_destroy;
 		}
-
-		free(ctx->description);
-		ctx->description = new_description;
 	}
 
+	free(ctx->description);
+	ctx->description = new_description;
 	ctx->params = *params;
 
 	return ctx;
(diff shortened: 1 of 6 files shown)


mhennerich (approved):

context.c
```diff
@@ -596,11 +596,10 @@ iio_create_context_from_xml(const struct iio_context_params *params,
 			prm_err(params, "Unable to alloc memory\n");
 			goto err_context_destroy;
 		}
-
-		free(ctx->description);
-		ctx->description = new_description;
 	}
 
+	free(ctx->description);
+	ctx->description = new_description;
```
pamolloy, line 602: If `description == null` and `ctx->description` is a valid pointer this will free the valid pointer and then assign `new_description` (which is `null`) to `ctx-description`.

nunojsa: correct... good catch!

pcercuei: Well, I disagree. If `description` is NULL it makes sense that `ctx->description` is set to NULL, no?

pcercuei: Nevermind, I think you are right. It *would* make sense if the description was replaced, but the code tries to append the old description to the new one.
So please send a PR.

pamolloy: Thanks for the quick replies! Just out of curiosity, how is `description` here normally populated? In my case I'm running `iio_info -u "serial:..."` and IIOD on an MCU.

pcercuei: Here it is populated by the XML backend from the XML string sent by IIOD, which itself will get the description from the libiio backend's `iio_backend` struct.

pamolloy: Sorry that wasn't phrased clearly. :sweat_smile:  I understand how it works in my case.

I was curious in what cases the `description` that is passed in as an argument to `iio_create_context_from_xml()` will not be `null`, which I assume is the "normal" use case.

pcercuei: It happens e.g. with the network backend, which prepends the IP address to the description found in the XML string.

PR #1128, opened 2024-01-24
Embedded libiio + libtinyiiod

pcercuei:
This set of patches enables support for compiling Libiio + IIOD on no-OS embedded devices.

Libiio itself can now be compiled without threads, using the `NO_THREADS` CMake option. This conflicts with all current backends (local, serial, usb, network) so those have to be disabled.

For embedded use, the backend used should be the "external" backend - enabled using the `WITH_EXTERNAL_BACKEND` CMake option. The application then has the responsability to provide the Libiio backend itself (or through a board-specific external library).

The embedded applications can now also implement a IIOD interface, using "libtinyiiod" that can be enabled with the `WITH_LIBTINYIIOD` CMake option. The API is described in `iiod/libtinyiiod.h`; the application is responsible for providing the callbacks to read/write on the communication bus (which can also be provided by a board-specific external library).


rgetz:
I'm assuming that when this is included - we would deprecate
- https://github.com/analogdevicesinc/libtinyiiod
- https://github.com/analogdevicesinc/no-OS/tree/main/iio
and use this instead?

How does this implemntation compare in size (text and data, stripped) to the existing libtinyiiod? 

What are the requirements for this implmentions "no-OS embedded devices." I assume any 32-bit MCU with enough internal memory?

Has this been tested to replace the use cases where the existing tinyiiod is being used? Maybe @dbogdan has some good ideas where this should be tested before merged.

pcercuei:
Don't know yet; @buha will play with it in the coming days/weeks.

buha:
So regarding size, a simple embedded app that just initializes the UART and configures this new tinyiiod to use the UART for communicating with the client, uses up:
```
   text	   data	    bss	    dec	    hex	filename
  95820	   1860	   2052	  99732	  18594	build/tinyiiod_demo.elf
```
This is with optimizations turned off and with debugging at `-g3`. (With optimizations on and debugging symbols off, I suspect flash usage would be reduced by a factor of 2x maybe 3x, didn't try.)

This would mean that the flash size needs to be 100 kB or more to fit such a program which is a good start, given that most micros that we use have 256 kB or more flash.

Ram usage at initialization is negligible. Ram usage at runtime after library init and a few mallocs I have not investigated.

My interpretation is that in any case, the benefits of using a common tinyiiod code base outweigh the cost of not being to run on the tinyest microcontrollers. It should also add that the iiod that was embedded into no-OS was lacking features compared to this one.

buha (approved):
I'm approving this since I can already run this on an embedded target and have it communicate with a client. Not all features are yet tested, but it's a good start.

PR #1133, opened 2024-01-25
Rework iio_context_get_xml()

pcercuei:
Instead of returning a string allocated alongside the IIO context, it
will now return a string allocated on the heap, that has to be freed
with free() afterwards. The function can now also return errors.

The advantage of this change is multiple:
- IIO contexts now don't have a huge XML string allocated during their
  whole lifetime;
- If the application does not call iio_context_get_xml(), then all the
  internal functions used to generate the XML string in the first place
  are dead code, and when building statically, can be garbage-collected.

C# bindings will be updated by @AlexandraTrifan  in #1122.

### DIFF
--- a/bindings/python/iio.py
+++ b/bindings/python/iio.py
@@ -31,6 +31,7 @@ from ctypes import (
     c_double,
     cast,
     sizeof,
+    string_at,
     POINTER as _POINTER,
     CDLL as _cdll,
     memmove as _memmove,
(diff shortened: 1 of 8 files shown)


bindings/python/iio.py
```diff
@@ -300,6 +301,7 @@ class ChannelType(Enum):
     _iiolib = "iio"
 
 _lib = _cdll("libiio.so.1", use_errno=True, use_last_error=True)
+_libc = _cdll(find_library("c"))
```
tfcollins, line 304: This will fail in Windows. Need to use something like ```ctypes.util.find_msvcrt```

Probably need to fix the _cdll call above as well since that breaks Windows too.

pcercuei: If Python bindings were broken on Windows before, and they are still broken on Windows after, then I'd argue it's not a problem of this commit and it can be fixed later in a subsequent PR. I'm kind of running against time here :)

mhennerich (approved):

tfcollins (approved):

PR #1138, opened 2024-01-31
Better backend API

pcercuei:
Modify/add functions:
- `iio_context_add_device`, which will create a `iio_device` and add it to the context;
- `iio_device_add_channel`, which will create a `iio_channel` and add it to the device;
- The functions `iio_{context,device,channel}_add_attr` to add attributes to contexts/devices/channels are now backend API
- `iio_create_context_from_backend` now takes a few extra parameters that will be used to create the iio_context.

With these changes, the XML backend was updated to be fully implemented on top of the public API (be it iio.h or iio-backend.h) and does not use iio-private.h anymore.

In theory, the local backend could do that too but it's a lot more work (due to how it currently creates the context).

The code should be considered very experimental, it was not thoroughly tested.

### DIFF
--- a/attr.c
+++ b/attr.c
@@ -7,6 +7,7 @@
  */
 
 #include "iio-private.h"
+#include "sort.h"
 
 #include <inttypes.h>
 #include <errno.h>
@@ -213,6 +214,8 @@ int iio_add_attr(union iio_pointer p, struct iio_attr_list *attrs,
 	}
 
 	attrs->num++;
+
+	iio_sort_attrs(attrs);
 
 	return 0;
 }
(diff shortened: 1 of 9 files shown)


buha (approved):
i've played with a few of these functions and they provide a good alternative to the `iio_create_context_from_xml` function which uses libxml2 which grows the .text area of the binary about 3 times in my tests (100k -> 300k)

what Paul did here is very good for embedding tinyiiod on small embedded devices

mhennerich (approved):

buha:
@pcercuei you may merge this

pcercuei:
Only people from ADI can merge 🙃

pcercuei:
Gentle ping.

Issue #1139, opened 2024-01-31
c++: Replace all asserts with throws


ekigwana:
https://github.com/analogdevicesinc/libiio/blob/7ae483696affb64ab6e1766f55e47f00456593d0/bindings/cpp/iiopp.h#L256

The asserts as utilized especially in the example above result in the program terminating. This is reasonable for test program accessing a particular device but when the code is part of a larger system this is highly undesirable. There are a couple ways of dealing with this:

1. Define NDEBUG prior to including iiopp.h
2. Replace asserts with a MACRO that a definition can convert to NO OP
3. Replace asserts with throws since the happy path should not incur the penalties of error checking

As an example of this undesirable behavior, if there is a hardware communication problem with a particular IIO device, it will not get enumerated by the kernel driver. Proceeding to then use the context to find the device by name then results in a terminal condition:
```
#include <iiopp.h>
#include <iostream>

int main(int argc, char ** argv)
{
    try
    {
        auto iio_ctx = iiopp::create_context(nullptr, nullptr);
        auto dev = iio_ctx->find_device("aaxi_fan_control"); // This results in an assert. Note misspelled name to simulate error
    }
    catch (iiopp::error & e)
    {
        std::cerr << "ERROR " << e.code().value() << ": " << e.what() << std::endl;
        return EXIT_FAILURE;
    }

    return EXIT_SUCCESS;
}
```

With above program producing the following output:
```
test: /usr/include/iio/iiopp.h:256: iiopp::impl::AttrSeqT<obj_T, get_attrs_count_T, get_attr_T, find_attr_T>::AttrSeqT(const obj_T*) [with obj_T = iio_device; unsigned int (* get_attrs_count_T)(const obj_T*) = iio_device_get_attrs_count; const iio_attr* (* get_attr_T)(const obj_T*, unsigned int) = iio_device_get_attr; const iio_attr* (* find_attr_T)(const obj_T*, const char*) = iio_device_find_attr]: Assertion `obj' failed.
Aborted
```

Please share your thoughts on this so I can work on a patch to resolve this issue. Also, is there a code format  spec for libiio? Like a clang-format or something? I have noticed different styles within iiopp.h.

pcercuei:
@sternmull, want to chime in?

sternmull:
The idea is that the wrapper classes always refer to non-null pointers of the C-API object, just like C++ references would do. This is the general assumption that is checked by the `assert`.

[...]

PR #1143, opened 2024-02-02
Solve #1139 by returning optionals where C-API returns NULL for objects

sternmull:
This should solve #1139

### DIFF
--- a/bindings/cpp/iiopp.h
+++ b/bindings/cpp/iiopp.h
@@ -40,7 +40,7 @@
  * - Classes that model the major types, providing member functions for easy usage.
  * - Error handling (errors are checked and turned into exceptions).
  * - Simplified resource management via smart pointers.
- * - Functions that may return \c NULL for "no string" are explicit by returning an <tt>std::optional</tt>.
+ * - Functions that may return \c NULL for "no string" or "no object" are explicit by returning an <tt>std::optional</tt>.
  * - Iterators and array subscription operator for idiomatic access to sequences.
  * - Implicit conversion to the wrapped C-types, so C++ instances can easily be passed to the C-API. Arguments C++-API take the raw C-types.
  *
@@ -109,6 +109,11 @@ namespace impl
 inline optstr opt(char const * s)
 {
     return s ? optstr{{s}} : optstr{};
+}
+
+template <class T, class C> optional<T> maybe(C * obj)
+{
+    return obj ? optional<T>{{obj}} : optional<T>{};
 }
 
 inline std::string err_str(int err)
@@ -206,7 +211,7 @@ class Attr
     iio_attr const * p;
 public:
     Attr() = delete;
-    Attr(iio_attr const * attr) : p(attr){}
+    Attr(iio_attr const * attr) : p(attr){assert(attr);}
     operator iio_attr const * () const {return p;}
 
     cstr name() {return iio_attr_get_name(p);}
@@ -281,19 +286,13 @@ public:
 template <class obj_T, iio_attr const * find_attr_T(obj_T const *, char const *)>
 optional<Attr> attr(obj_T const * obj, cstr name)
 {
-    iio_attr const * a = find_attr_T(obj, name);
-    if (a)
-        return {Attr(a)};
-    return {};
+    return maybe<Attr>(find_attr_T(obj, name));
 }
 
 template <class obj_T, iio_attr const * get_attr_T(obj_T const *, unsigned int)>
 optional<Attr> attr(obj_T const * obj, unsigned int index)
 {
-    iio_attr const * a = get_attr_T(obj, index);
-    if (a)
-        return {Attr(a)};
-    return {};
+    return maybe<Attr>(get_attr_T(obj, index));
 }
 
 } // namespace impl
@@ -341,7 +340,7 @@ class ChannelsMask
 public:
 
     ChannelsMask() = delete;
-    ChannelsMask(iio_channels_mask const * mask) : p(mask){}
+    ChannelsMask(iio_channels_mask const * mask) : p(mask){assert(mask);}
     operator iio_channels_mask const * () const {return p;}
 };
 
@@ -360,7 +359,7 @@ class Block
 public:
 
     Block() = delete;
-    Block(iio_block * block) : p(block){}
+    Block(iio_block * block) : p(block){assert(block);}
     operator iio_block * () const {return p;}
 
     void * start() {return iio_block_start(p);}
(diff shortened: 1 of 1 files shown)


ekigwana:
Thank you @sternmull, I'll continue our evaluation of this interface.

As an aside, we need to install iiopp.h when bindings are enabled
https://github.com/analogdevicesinc/libiio/blob/7ae483696affb64ab6e1766f55e47f00456593d0/CMakeLists.txt#L670

bindings/cpp/iiopp.h
```diff
@@ -561,9 +562,9 @@ class Device : public impl::IndexedSequence<Device, Channel>
     optstr label() const { return impl::opt(iio_device_get_label(p));}
     unsigned int channels_count() const {return iio_device_get_channels_count(p);}
     unsigned int attrs_count() const {return iio_device_get_attrs_count(p);}
-    Channel channel(unsigned int idx) const {return Channel{iio_device_get_channel(p, idx)};}
+    optional<Channel> channel(unsigned int idx) const {return impl::maybe<Channel>(iio_device_get_channel(p, idx));}
```
ekigwana, line 565: From my review this is a good solution. However we need to extend the design concept all the way to iiopp::Context. E.g. ```Device iiopp::Context::find_device(cstr name);``` Still suffers from the assertion bug as shown by the example below:
```
#include <iiopp.h>
#include <iostream>

int main(int argc, char ** argv)
{
    try
    {
        auto iio_ctx = iiopp::create_context(nullptr, nullptr);
        auto dev = iio_ctx->find_device("aaxi_fan_control"); // This results in an assert
    }
    catch (iiopp::error & e)
    {
        std::cerr << "ERROR " << e.code().value() << ": " << e.what() << std::endl;
        return EXIT_FAILURE;
    }

    return EXIT_SUCCESS;
}
```
g++ -ggdb3 -I/usr/include/iio -o cpp_test cpp_test.cpp -liio

ekigwana (changes requested):

sternmull:
Somehow I missed the `Context` when looking for the functions that needed to be changed. Updated the code.

ekigwana:
I have updated my test program (Note that we are dealing with an ADRV9009-ZU11EG for our project and it has over 1000 attributes)
```
#include <cassert>
#include <cstring>
#include <iiopp.h>
#include <iostream>

/*
 * hwmon3: axi_fan_control
 *         3 channels found:
 *                 fan1:  (input)
 *                 3 channel-specific attributes found:
 *                         attr  0: fault (fan1_fault) value: 0
 *                         attr  1: input (fan1_input) value: 1906
 *                         attr  2: label (fan1_label) value: FAN
 *                 temp1:  (input)
 *                 2 channel-specific attributes found:
 *                         attr  0: input (temp1_input) value: 49642
 *                         attr  1: label (temp1_label) value: SYSMON4
 *                 pwm1:  (input)
 *                 1 channel-specific attributes found:
 *                         attr  0: pwm1 value: 64
 *         No trigger on this device
 */

int main(int argc, char ** argv)
{
    try
    {
        auto iio_ctx = iiopp::create_context(nullptr, nullptr);
        {
                auto dev = iio_ctx->find_device("aaxi_fan_control");
                assert(!dev);
        }

        auto dev_name = "axi_fan_control";
        auto dev = iio_ctx->find_device(dev_name);
        assert(dev);

        {
                auto attr = dev->find_attr("non-existent");
                assert(!attr);
                auto ch1 = dev->find_channel("non-existent", false);
                assert(!ch1);
                auto ch2 = dev->find_channel("fan1", true);
                assert(!ch2);
        }

        auto ch_id = "fan1";
        auto ch = dev->find_channel(ch_id, false);
        assert(ch);

        {
                auto attr = ch->find_attr("non-existent");
                assert(!attr);
        }

        auto attr_name = "input";
        auto attr = ch->find_attr(attr_name);
        assert(attr);

        {
                auto name = dev->name();
                std::cout << "dev name: " << ((name) ? *name : "unknown") << std::endl;
                assert(name);
                assert(std::strcmp(*name, dev_name) == 0);
        }

        {
                auto id = ch->id();
                std::cout << "ch id: " << id << std::endl;
                assert(std::strcmp(id, ch_id) == 0);

                auto name = ch->name();
                std::cout << "ch name: " << ((name) ? *name : "unknown") << std::endl;
                assert(!name);
        }

        {
                auto name = attr->name();
                std::cout << "attr name: " << ((name) ? name : "unknown") << std::endl;
                assert(name);
                assert(std::strcmp(name, attr_name) == 0);
        }
    }
    catch (iiopp::error & e)
    {
        std::cerr << "ERROR " << e.code().value() << ": " << e.what() << std::endl;
        return EXIT_FAILURE;
    }

    return EXIT_SUCCESS;
}
```

So far that works. I am not fond of the optional deleted objects that cannot be reused in a loop. A pointer would have had the same effect and can be reassigned but i understand the design principal per the issue text.

ekigwana (approved):

sternmull:
> I am not fond of the optional deleted objects that cannot be reused in a loop. A pointer would have had the same effect and can be reassigned but i understand the design principal per the issue text.

I don't want to force any design decision. But I also don't see the problem. `std::optional` and the wrapper classes themself are very lightweight objects. I would expect any reasonable optimizing compiler to make them basically zero cost. Also I don't understand why you would not be able to reuse them the same way as you could with pointers. Maybe you can give a small example where it is inconvenient or inefficient?

A pattern that I find very useful with optionals (works for pointers too):

```
if (auto attr = dev->find_attr("foo")) // Makes sure attr is only visible where it is safe to dereference
{
  attr->bar(); // safe to use attr
}
```

But I am not sure if this is related to your critique.

ekigwana:
> > I am not fond of the optional deleted objects that cannot be reused in a loop. A pointer would have had the same effect and can be reassigned but i understand the design principal per the issue text.
> 
> I don't want to force any design decision. But I also don't see the problem. `std::optional` and the wrapper classes themself are very lightweight objects. I would expect any reasonable optimizing compiler to make them basically zero cost. Also I don't understand why you would not be able to reuse them the same way as you could with pointers. Maybe you can give a small example where it is inconvenient or inefficient?
> 
> A pattern that I find very useful with optionals (works for pointers too):
> 
> ```
> if (auto attr = dev->find_attr("foo")) // Makes sure attr is only visible where it is safe to dereference
> {
>   attr->bar(); // safe to use attr
> }
> ```
> 
> But I am not sure if this is related to your critique.

After further review, I do not expect it to be a problem at all please disregard that comment.

dNechita (approved):

cristina-suteu (approved):

cristina-suteu:
@mhennerich @dNechita can I merge this ?

mhennerich (approved):

mhennerich:
> @mhennerich @dNechita can I merge this ?

Yes

PR #1146, opened 2024-02-23
iiod: Add default env var to service file

### LINKED ISSUE #1142: Systemd warns about unset variables
During startup of iiod, we get the following warning:
`iiod.service: Referenced but unset environment variable evaluates to an empty string: IIOD_EXTRA_OPTS`

This is because since v254 of systemd, unset environmental variables will result in a warning (see [commit](https://github.com/systemd/systemd-stable/commit/f331434d13488425ccd8485327085d15f8f92aea)).

I could make a PR where I add this to the service file:
`Environment=IIOD_EXTRA_OPTS=''`

Would this be a good solution?

dennis-ju:
Systemd now gives a warning when IIOD_EXTRA_OPTS is unset.

This change sets it to an empty string by default while still allowing it be overridden by /etc/default/iiod.

Fixes #1142.

### DIFF
--- a/iiod/init/iiod.service.cmakein
+++ b/iiod/init/iiod.service.cmakein
@@ -2,7 +2,7 @@
 #
 # iiod - Systemd init script
 #
-# Copyright (C) 2016-2021 Analog Devices Inc.
+# Copyright (C) 2016-2024 Analog Devices Inc.
 
 [Unit]
 Description=IIO Daemon
@@ -11,6 +11,7 @@ After=network.target systemd-udev-settle.service
 ConditionPathExists=/sys/bus/iio
 
 [Service]
+Environment=$IIOD_EXTRA_OPTS=''
 EnvironmentFile=-/etc/default/iiod
 ExecStart=@CMAKE_INSTALL_FULL_SBINDIR@/iiod $IIOD_EXTRA_OPTS
 KillMode=process


iiod/init/iiod.service.cmakein
```diff
@@ -2,7 +2,7 @@
 #
 # iiod - Systemd init script
 #
-# Copyright (C) 2016-2021 Analog Devices Inc.
+# Copyright (C) 2016-2024 Analog Devices Inc.
```
nunojsa, line 5: Kind on unrelated change. You could have mentioned the change in the commit message....

nunojsa (commented):
Just one question.... Does this also affects the main branch? If so, the natural flow is to change the target branch to main (in this PR) and then cherry pick this into libiio-v0

Thanks for the contributing!

nunojsa (approved):

cristina-suteu:
can I merge this ?

nunojsa:
> can I merge this ?

@mhennerich ?

dennis-ju:
> > Just one question.... Does this also affects the main branch? If so, the natural flow is to change the target branch to main (in this PR) and then cherry pick this into libiio-v0
> 
> Forget about this... just saw that the commit is already in main. Next time run cherry-pick with -x and then it's clear to me :)

Alright, I'll do that next time :)

PR #1154, opened 2024-03-21
readme: Switch to using the term 'main' branch

dNechita:


README.md
```diff
@@ -32,9 +32,9 @@ If you have a question about libiio and an Analog Devices IIO kernel driver plea
...
 |  | [![Build Status](https://dev.azure.com/AnalogDevices/OpenSource/_apis/build/status/analogdevicesinc.libiio?branchName=master&stageName=Builds&jobName=WindowsBuilds&configuration=WindowsBuilds%20VS2022)](https://dev.azure.com/AnalogDevices/OpenSource/_build/latest?definitionId=9&branchName=master) | Windows-64 Server 2022 | (libiio-setup.exe works for both Windows Server 2019 and Windows Server 2022) | [![Latest Windows zip](https://raw.githubusercontent.com/wiki/analogdevicesinc/libiio/img/win_box.png)](https://swdownloads.analog.com/cse/azure_builds/Windows-VS-2022-x64-latest_master_libiio.zip) |
```
cristina-suteu, line 40: should the download links use  main instead of master as well ? (https://dev.azure.com/AnalogDevices/OpenSource/_build/latest?definitionId=9&branchName=master -->  main here)

dNechita: I didn't want to mess with the links from Azure. Maybe @SRaus can give us some advice.

SRaus: In theory, it should work by just replacing 'master' with 'main' in links:
- and this will work for Status Badges
- but not for installers/packages. Most of the artifacts do not exists with _main_ in name (i.e. Windows-VS-2022-x64-latest_master_libiio.zip exists, but Windows-VS-2022-x64-latest_main_libiio.zip not). 

It looks that the name of the installers, packages, or other artifacts needs to be updated also in /[CI/azure/prepare_assets.sh](https://github.com/analogdevicesinc/libiio/blob/main/CI/azure/prepare_assets.sh#L77C1-L81C2), and then you'll need to create a release so they will get uploaded on SWDownloads with the new names.

SRaus: This become a bit more complex than it looks, if there will be different releases (so implicitly different sets of artifacts) from different branches - in the sense that CI pipeline and scripts will be different from one branch to another.

cristina-suteu: so, could we keep the links as they are and merge this ? and then maybe handle the download links in a separate PR? would that make sense ?

SRaus: The links on the lines that starts with "[Build Status]" won't show anything now (actually will point to a ["Never built" ]( https://dev.azure.com/AnalogDevices/OpenSource/_apis/build/status/analogdevicesinc.libiio?branchName=master&stageName=Builds&jobName=WindowsBuilds&configuration=WindowsBuilds%20VS2019) status), because they still look for 'master' builds. 
Since they can be fixed easily, by just replacing 'master' with 'main', I would go and change them in this PR.

The other links - download links - are functional if nothing is changed on them, with single dropdown that they still contain 'master' keyword in their names. 
I agree to postpone this to another PR, because this imply touching a lot of files, like pipelines, building scripts, links in ReadMe and so on.

dNechita: I have added a new commit that updates the links about "[Build Status]". This PR should be good to merge.

SRaus: > I have added a new commit that updates the links about "[Build Status]". This PR should be good to merge.

From my point of view, you can squash them in one single commit.

SRaus (approved):

README.md
dNechita: Sounds good to me

dNechita:
V2:
 - squash the 2nd commit and meld into the 1st one

SRaus (approved):

SRaus:
You can go on and merge this PR.
DevOps will take care of updating artifacts name (to contain 'main' instead of 'master' in their names), update download links to new files and add in documentation the recently added builds (i.e. for Debian Bookworm).

PR #1159, opened 2024-04-11
iio_info.py example: pick uri automatically when there's only one

royalmo:
## PR Description

I was testing v0.25's iio_info.py and saw it didn't do anything when no command-line arguments were given. If there are more than one available uri a message is shown before exiting the program (with `sys.exit(0)`). However, this line is also executed when there's only one or zero of them, and without printing any message previously. This causes confusion, at least in my opinion. I've seen that v1 didn't had any changes in that file, meaning that this issue still persists.

For reference, the "oficial" C-programmed iio_info (the one that comes built in the releases) selects the first uri if only one exists by default.

What I've done in this PR is to mimic what it's being done in the iio_info executable. I'm not sure if the examples should change in v1, but at least it should be useful for v0.25 (this PR is on the main branch, but I can move it to the v0.25 branch if you prefer). I've also fixed a typo some lines below.

It's my first PR ever to an open-source project, so don't hesitate on giving me constructive feedback, it's all a little new to me!

Tested on v0.25, with Python3.6 and Python2.7 on a Ubuntu 18.04.

## PR Type

## PR Checklist
(due to the simplicity of the fix, and given the fact that it is one of the example codes I consider that some of the checks don't apply, I left them unchecked)

### DIFF
--- a/bindings/python/examples/iio_info.py
+++ b/bindings/python/examples/iio_info.py
@@ -31,7 +31,11 @@ def _create_context():
             print("Multiple contexts found. Please select one using --uri:")
             for uri, description in contexts.items():
                 print("\t%s: %s" % (uri, description))
-        sys.exit(0)
+            sys.exit(0)
+        if len(contexts) == 0:
+            print("No contexts found.")
+            sys.exit(0)
+        uri = list(contexts.keys())[0]
 
     return iio.Context(uri)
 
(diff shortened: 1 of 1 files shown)


bindings/python/examples/iio_info.py
```diff
@@ -31,7 +31,11 @@ def _create_context():
             print("Multiple contexts found. Please select one using --uri:")
             for uri, description in contexts.items():
                 print("\t%s: %s" % (uri, description))
-        sys.exit(0)
+            sys.exit(0)
```
dNechita, line 34: The change on this line doesn't make sense to me as it prints only the first item, then immediately exits. What we want is to print all contexts.

royalmo: Hello! Maybe I'm missing something, but I see the new sys.exit outside the for loop, so we are still printing all contexts. Before it was outside the if statement (which we don't want anymore since the code continues).

dNechita: Oh, you're right. I haven't carefully counted the spaces before `sys.exit(0)` and I thought it's inside the loop.

Issue #1174, opened 2024-06-04
[Windows] Error when running iio_info


SrikanthPagadarai:
I grabbed the latest Azure builds for Windows and ran iio_info -s. It opens an error window with the message "The procedure entry point sp_blocking_read could not be located in the dynamic link library C:\WINDOWS\system32\libiio1.dll."

tfcollins:
A few things:
- Based on the iss file libzstd is not included but its built against it in CI
- Serial support seems broken on Windows
- Python bindings are also broken on Windows

cristina-suteu:
@tfcollins regarding the serial support , a fix was added here https://github.com/analogdevicesinc/libiio/pull/1163 , is serial support still broken after this ? 
as for the other things, i'm working on fixing them

rgetz:
I think this is fixed, and can be closed?

PR #1183, opened 2024-06-28
Fix Windows Builds

cristina-suteu:
## PR Description

As things was before, for the windows deps an archive was downloaded from swdownloads. Said archive contains windows deps compiled with MinGW. The installer is created based on the MSVC builds. Both MSVC and MinGW builds were using deps from this archive compiled with MinGW. While the project seemed to be building okay, this caused problems which were visible in both installer and MSVC libiio zips at runtime. 
By adding a new archive with deps compiled with MSVC to be used when building libiio with MSVC, the issue is resolved. 
Moreover, a check for the compiler used was added in the windows build script. This ensures the correct paths to libiio's deps according to the compiler used. 
This fixes issue https://github.com/analogdevicesinc/libiio/issues/1174

## PR Type

## PR Checklist
- [ ] I have checked that I did not introduced new warnings or errors (CI output)

### DIFF
--- a/CI/publish_deps.ps1
+++ b/CI/publish_deps.ps1
@@ -9,7 +9,10 @@ cd $src_dir
 mkdir dependencies
 cd dependencies
 wget http://swdownloads.analog.com/cse/build/libiio-deps-20220517.zip -OutFile "libiio-win-deps.zip"
+wget https://swdownloads.analog.com/cse/build/libiio-deps-20240627.zip -OutFile "libiio-msvc-deps.zip"
+
 7z x -y "libiio-win-deps.zip"
+7z x -y "libiio-msvc-deps.zip"
 
 # Version numbers inside this directory change all the time; print what's
 # currently in the folder to make it easier to debug CI breakages on MinGW.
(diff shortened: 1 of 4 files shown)


CI/publish_deps.ps1
```diff
@@ -9,7 +9,10 @@ cd $src_dir
 mkdir dependencies
 cd dependencies
 wget http://swdownloads.analog.com/cse/build/libiio-deps-20220517.zip -OutFile "libiio-win-deps.zip"
+wget https://swdownloads.analog.com/cse/build/libiio-deps-20240627.zip -OutFile "libiio-msvc-deps.zip"
```
dNechita, line 12: It would help (in the future) to rename libiio-deps-20200627 to a name that doesn't change over time. For example: libiio-deps-msvc. If at some point in time, there will be a new libiio-deps-yyyymmdd ,then there will be only one place to update the name. Now there are extra 4 places besides this one.

cristina-suteu: i used the yyyy-mm-dd format, to match the naming used in the other archive. 
as a side note, the old archive with mingw compiled deps should also be updated to contain all required files. would it make sense to update the names of both archives to leave out the yyyy-mm-dd suffix after fixing the mingw archive as well ?

rgetz:
As things was before - was sort of fragile and is sort of iffy from a licensing standpoint...

replicating it might not be the right path forward.

tfcollins:
Providing the complete build would address the licensing issue. The installer probably needs to be checked that it does reference anything packaged as well

rgetz:
I think this is replaced by #1197 and this can be closed (without merging)?

PR #1229, opened 2025-01-16
Channel label support

dNechita:
## PR Description

Add support for labels in IIO channels, similar to IIO devices.
These changes add `iio_channel_get_label(const struct iio_channel *chn)` to the existing API.

This is a draft. The core functionality is already implemented. Updating all bindings is currently under development.

## PR Type
- [ *] New feature (a change that adds new functionality)

## PR Checklist


device.c
```diff
@@ -139,6 +139,7 @@ struct iio_channel * iio_device_find_channel(const struct iio_device *dev,
 			continue;
 
 		if (!strcmp(chn->id, name) ||
+				(chn->name && !strcmp(chn->label, name)) ||
```
mhennerich, line 142: Shouldn't this be ... (chn->label && !strcmp(chn->label, name)) ...

dNechita: Yes, it should. Thanks!

dNechita:
V2:

- fixed 'name' being used instead of 'label' (based on review fidings)
- added new commits with updates for iio_info and iio_attr
- rebased on latest main branch

nunojsa:
Hi @dNechita,

I plan to give a better review on this one next week. For now, some points from me:

* The git title in the first two patches don't follow the typical style. If you're touching lots of files (really because you have too), you can do `treewide: ...`;
* The order of the patches does not look good to me. For example, exposing an API (the first patch) which will do nothing makes no sense to me. What about reversing the order of the first 3 patches?

rgetz:
reversing the order of the patches will cause build failures (won't it?) - since you will be using label, before it's added/initialized.

I can see swapping the order of:
- [Enhance iio_device_find_channel() to use labels](https://github.com/analogdevicesinc/libiio/pull/1229/commits/d5c326d905e790d6360ac2e9f5e229ccefb020dd)
- [local: Init channel label when found](https://github.com/analogdevicesinc/libiio/pull/1229/commits/74c0b97fcdac3d099645a1d83db8a5ac5ff99bef)

so that local is filled out when you start using them - but it shouldn't really matter too much...

nunojsa:
> reversing the order of the patches will cause build failures (won't it?) - since you will be using label, before it's added/initialized.

I didn't meant literally. Likely, as you pointed out, it needs some rework in order to change the order. But it should be doable to code all the foundation and the last patch to just be exposing the API to the "outside world". As it stands, I do not think it makes to much sense to expose an API without being implemented.

dNechita:
> I think the patch set is incomplete/wrong somewhere:
> 
> running it I get (on iio_info):
> 
> ```
>         hwmon18: max15301 (label: VCCPSINT)
>                 5 channels found:
>                         curr1:  (label: iout1) (input)
>                         4 channel-specific attributes found:
>                                 attr  0: crit (curr1_crit) value: 11015
>                                 attr  1: crit_alarm (curr1_crit_alarm) value: 0
>                                 attr  2: input (curr1_input) value: 1396
>                                 attr  3: label (curr1_label) value: iout1
>                         temp2:  (input)
>                         3 channel-specific attributes found:
>                                 attr  0: crit (temp2_crit) value: 115000
>                                 attr  1: crit_alarm (temp2_crit_alarm) value: 0
>                                 attr  2: input (temp2_input) value: 47125
>                         in1:  (label: vin) (input)
>                         3 channel-specific attributes found:
>                                 attr  0: alarm (in1_alarm) value: 0
>                                 attr  1: input (in1_input) value: 12000
>                                 attr  2: label (in1_label) value: vin
>                         temp1:  (input)
>                         3 channel-specific attributes found:
>                                 attr  0: crit (temp1_crit) value: 115000
>                                 attr  1: crit_alarm (temp1_crit_alarm) value: 0
>                                 attr  2: input (temp1_input) value: 43125
>                         in2:  (label: vout1) (input)
>                         3 channel-specific attributes found:
>                                 attr  0: alarm (in2_alarm) value: 0
>                                 attr  1: input (in2_input) value: 849
>                                 attr  2: label (in2_label) value: vout1
> ```
> 
> the issue I want to point out is that the lable still shows up as an attribute, when it should be eaten as a label...
> 
> We see:
> 
> ```
>                         curr1:  (label: iout1) (input)
>                         4 channel-specific attributes found:
>                                 attr  0: crit (curr1_crit) value: 11015
>                                 attr  1: crit_alarm (curr1_crit_alarm) value: 0
>                                 attr  2: input (curr1_input) value: 1396
> ------->                        attr  3: label (curr1_label) value: iout1
>                         temp2:  (input)
> ```
> 
> rather than:
> 
> ```
>                         curr1:  (label: iout1) (input)
>                         4 channel-specific attributes found:
>                                 attr  0: crit (curr1_crit) value: 11015
>                                 attr  1: crit_alarm (curr1_crit_alarm) value: 0
>                                 attr  2: input (curr1_input) value: 1396
>                         temp2:  (input)
> ```
> 
> It doesn't make sense to expose the `curr_label` - now that we have a proper way of doing it...
> 
> I think this just means it's missing an early return somewhere in local.c for example, this is what the early return does here:
> 
> ```
> static int add_attr_to_device(struct iio_device *dev, const char *attr)
> {
> 	unsigned int i;
> 
> 	for (i = 0; i < ARRAY_SIZE(device_attrs_denylist); i++)
> 		if (!strcmp(device_attrs_denylist[i], attr))
> 			return 0;
> 
> 	if (!strcmp(attr, "name") || !strcmp(attr, "label"))
> 		return 0;
> ```

Yes, I have missed adding the implementation that would consider the label not to be a simple attribute. I will include it. Thanks!

dNechita:
> > reversing the order of the patches will cause build failures (won't it?) - since you will be using label, before it's added/initialized.
> 
> I didn't meant literally. Likely, as you pointed out, it needs some rework in order to change the order. But it should be doable to code all the foundation and the last patch to just be exposing the API to the "outside world". As it stands, I do not think it makes to much sense to expose an API without being implemented.

Regarding the order of the commits, this time I just used the order presented in issue that was requesting the feature. But if it wasn't for that order I would have still started with the end in mind, meaning adding API, and even updating the clients of the API and then going for the implementation. This is more of an order of how I would develop, and I can keep it while changing the commits order when creating a PR if that is easier to review.

nunojsa:
> This is more of an order of how I would develop, and I can keep it while changing the commits order when creating a PR if that is easier to review.

It's not it is easier to review. It just does not make sense to me to have a public API without being implemented. Naturally you're thinking on just having the complete PR merged and then this does not matter... I'm more looking at the series on a patch by patch basis and what makes sense to me. Sometimes, one also should have in mind things like `git bisect` for trying to figure what patch introduced some bug and so we should make sure that a specific patch won't willingly break something. 

All the above said, this is a nitpick and a minor detail

dNechita:
V3:

- Reorder the commits such that the implementation goes in the 1st commit
- Reworded messages of commits that touch multiple files by adding 'treewide:' prefix
- added the commit for the C# bindings

include/iio/iio.h
```diff
@@ -876,6 +876,12 @@ __api __check_ret __pure const char * iio_channel_get_id(const struct iio_channe
  * <b>NOTE:</b> if the channel has no name, NULL is returned. */
 __api __check_ret __pure const char * iio_channel_get_name(const struct iio_channel *chn);
 
+/** @brief Retrieve the channel name (e.g. <b><i>anglY</i></b>)
```
rgetz, line 879: copy/paste? Shouldn't this say `Retrieve the channel label`?

local.c
```diff
@@ -835,6 +835,23 @@ static int add_attr_to_channel(struct iio_channel *chn,
...
 	struct iio_attr_list *attrs;
 	int ret;
+	const char *lbl_suffix = "_label";
+	size_t lbl_len = strlen(lbl_suffix);
+	const char *chn_label_ptr = NULL;
+	char label[512];
+	const char *dev_id;
+
+	if (strlen(name) >= lbl_len &&
+		!strcmp(name + strlen(name) - lbl_len, lbl_suffix)) {
```
rgetz, line 845: This is more subjective - Since `lbl_suffix` is only used once - why not just use the string constant `"_label"` in the code? In theory - it should be done by the cpp - but this will ensure that. Same for the use of `strlen(lbl_suffix)` just use `strlen("_label")` in two places. Makes it easier to read IMHO.

utils/iio_attr.c
```diff
@@ -172,8 +172,12 @@ static int dump_channel_attributes(const struct iio_device *dev,
...
 					type_name);
 		}
-		if (iio_channel_get_name(ch) && quiet == ATTR_VERBOSE)
-			printf("id '%s', ", iio_channel_get_name(ch));
+		if (quiet == ATTR_VERBOSE) {
+			if (iio_channel_get_name(ch))
+				printf("id '%s', ", iio_channel_get_name(ch));
+			if (iio_channel_get_label(ch))
+				printf("label '%s', ", iio_channel_get_label(ch));
+		}
```
rgetz, line 180: could have just left the existing code, and added:

```
if (iio_channel_get_label(ch)) && quiet == ATTR_VERBOSE)
        printf("label '%s', ", iio_channel_get_label(ch));
```
would be less changes/churn.

rgetz:
I think the iio_info issue is fixed...

```
       hwmon2: dell_smm
                6 channels found:
                        fan1:  (label: Processor Fan) (input)        <- looks cool
                        1 channel-specific attributes found:
                                attr  0: input (fan1_input) value: 1009
```
and main branch:
```
rgetz@brain:~/github/libiio/build$ sudo ./utils/iio_attr -u local: -c hwmon2 fan1
dev 'dell_smm', channel 'fan1' (input), attr 'input', value '1008'
dev 'dell_smm', channel 'fan1' (input), attr 'label', value 'Processor Fan'
``` 

turns into:
```
rgetz@brain:~/github/libiio/build$ sudo ./utils/iio_attr -u local: -c hwmon2 fan1
dev 'dell_smm', channel 'fan1' (input), label 'Processor Fan', attr 'input', value '1010'
```

which I think is what we are looking for...

:)

local.c
```diff
@@ -835,6 +835,23 @@ static int add_attr_to_channel(struct iio_channel *chn,
...
+		dev_id = iio_device_get_id(iio_channel_get_device(chn));
+		ret = local_do_read_dev_attr(dev_id, 0, name,
+			label, sizeof(label), IIO_ATTR_TYPE_CHANNEL);
+		if (ret > 0) {
+			chn_label_ptr = label;
+			chn->label = iio_strdup(chn_label_ptr);
+		}
+		return 0;
+	}
```
nunojsa, line 854: To add on Robin's comments I wonder if we could not do this simpler:

```
name = get_short_attr_name(chn, name);

if (!strcmp(name, "label") {
    const char *dev_id = iio_device_get_id(iio_channel_get_device(chn));

    ret = local_do_read_dev_attr(dev_id, 0, name,
			label, sizeof(label), IIO_ATTR_TYPE_CHANNEL);
    /* But I actually wonder if we should ignore errors in this case?!! */
    if (ret > 0) {
        ...
        return 0;
    }
}
```

include/iio/iio.h
dNechita: yes, it should. And yes, it was a copy/paste side effect :)

local.c
dNechita: I created `lbl_suffix` because `"_label"` is used in 2 places and I believed it would be easier to change that string only once if needed. But now that I think better, it's unlikely that "_label" will change in the future. I agree that your suggestion is just easier to read.

utils/iio_attr.c
dNechita: I see you point. I updated this.

local.c
nunojsa: >  But now that I think better, it's unlikely that "_label" will change in the future

That would be an ABI breakage that won't happen in the kernel :)

local.c
dNechita: By simpler, do you mean doing the entire if() after `name = get_short_attr_name(chn, name);`? The thing is that `name `will never be equal to "label" since at this point, the channel name isn't known yet, therefore name will still resemble to something like: "in_voltage0_label".

For the other thing, I could add an error message for situations where label reading fails but I am not sure what else can be done. We will end up with a channel with no label and no attribute called 'label' and that is it. Are you considering doing more in this scenario?

dNechita:
V4:

- Use 'label' instead of 'name' in the description of iio_channel_get_label() in iio.h
- Use "_label" instead of variable for better readability in local.c
- Add new code without touching existing code in iio_attr.c

local.c
nunojsa: > By simpler, do you mean doing the entire if() after name = get_short_attr_name(chn, name);? The thing is that name will never be equal to "label" since at this point, the channel name isn't known yet, therefore name will still resemble to something like: "in_voltage0_label".

I meant using existing code already. I would expect `get_short_attr_name()` to just give us the attr name otherwise the name is misleading. But looking at it's implementation and given your example (in_voltage0_label), `get_short_attr_name()` should actually return "label" shouldn't it ?

> For the other thing, I could add an error message for situations where label reading fails but I am not sure what else can be done. We will end up with a channel with no label and no attribute called 'label' and that is it. Are you considering doing more in this scenario?

Not really (at least I do not think so)... If you fail to read it, you should actually fail in creating the device.... `foreach_in_dir()` should stop which should lead to freeing the device.

But thinking a bit more on it, maybe let's put this on hold and just log (with debug) the failure. The reason is that on the kernel side, as soon as your device has a .read_label() callback, all channels will have the label attribute. So far I think than when this happens, we have a label for every channel but I do not think this is an hard rule. Therefore, we might have channels with the _label attribute but when you read it, you get nothing out of it.

rgetz:
Looks good to me - check on the issues Codacy is pointing out, and decide which ones you want to ignore/fix.

dNechita:
V5:

- add error message if it fails to read channel label in local.c

rgetz (approved):
looks good to me.

rgetz:
```
index 1500c39d..29c14e58 100644
--- a/examples/iio-monitor.c
+++ b/examples/iio-monitor.c
@@ -165,7 +165,7 @@ static void * read_thd(void *d)

        while (!stop) {
                struct iio_device *dev;
-               const char *name;
+               const char *name, *label;
                int row, col, len, align, line = 2;
                unsigned int i, nb_channels, nb = 0;
                char buf[1024];
@@ -208,9 +208,12 @@ static void * read_thd(void *d)

                        nb++;
                        name = iio_channel_get_name(chn);
+                       label = iio_channel_get_label(chn);
                        id = iio_channel_get_id(chn);
                        if (!name)
                                name = id;
+                       if (label)
+                               name = label;
                        unit = id_to_unit(id);
```

also helps

dNechita:
> if (label)

Will add. Thanks for the code snippet.

local.c
```diff
@@ -835,6 +835,24 @@ static int add_attr_to_channel(struct iio_channel *chn,
 	union iio_pointer p = { .chn = chn, };
 	struct iio_attr_list *attrs;
 	int ret;
+	size_t lbl_len = strlen("_label");
+	const char *chn_label_ptr = NULL;
+	char label[512];
+	const char *dev_id;
+
+	if (strlen(name) >= lbl_len &&
+		!strcmp(name + strlen(name) - lbl_len, "_label")) {
```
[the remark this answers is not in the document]
dNechita: I forgot to address your concern.

With these changes:

```
diff --git a/local.c b/local.c
index 2b28912c..28fe25b7 100644
--- a/local.c
+++ b/local.c
@@ -851,6 +851,8 @@ static int add_attr_to_channel(struct iio_channel *chn,
                } else {
                        chn_perror(chn, ret, "Unable to read channel label");
                }
+               fprintf(stderr, "initial_name: %s\n", name);
+               fprintf(stderr, "name_after_get_short: %s\n", get_short_attr_name(chn, name));
                return 0;
        }

```

I am getting the following:

analog@analog:~/workspace-dan/libiio/build_v1 $ utils/iio_info 
initial_name: out_altvoltage1_TX1_I_F2_label
name_after_get_short: TX1_I_F2_label
initial_name: out_altvoltage0_TX1_I_F1_label
name_after_get_short: TX1_I_F1_label
initial_name: out_altvoltage7_TX2_Q_F2_label
name_after_get_short: TX2_Q_F2_label
initial_name: out_altvoltage6_TX2_Q_F1_label
name_after_get_short: TX2_Q_F1_label
initial_name: out_altvoltage3_TX1_Q_F2_label
name_after_get_short: TX1_Q_F2_label
initial_name: out_altvoltage2_TX1_Q_F1_label
name_after_get_short: TX1_Q_F1_label
initial_name: out_altvoltage5_TX2_I_F2_label
name_after_get_short: TX2_I_F2_label
initial_name: out_altvoltage4_TX2_I_F1_label
name_after_get_short: TX2_I_F1_label
initial_name: out_altvoltage1_TX_LO_label
name_after_get_short: TX_LO_label
initial_name: out_altvoltage0_RX_LO_label
name_after_get_short: RX_LO_label
initial_name: in_voltage7_vrefn_label
name_after_get_short: vrefn_label
initial_name: in_voltage1_vccaux_label
name_after_get_short: vccaux_label
initial_name: in_voltage2_vccbram_label
name_after_get_short: vccbram_label
initial_name: in_voltage3_vccpint_label
name_after_get_short: vccpint_label
initial_name: in_voltage0_vccint_label
name_after_get_short: vccint_label
initial_name: in_voltage6_vrefp_label
name_after_get_short: vrefp_label
initial_name: in_voltage5_vccoddr_label
name_after_get_short: vccoddr_label
initial_name: in_voltage4_vccpaux_label
name_after_get_short: vccpaux_label

Which seems that it strips away the input/output information and the channel id.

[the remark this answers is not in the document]
nunojsa: Yeah, I was now just getting more familiar with libiio code for populating channels and attrs. Indeed for legacy devices which make use of extended_names (which are now exposed as labels) the short name function won't work.

out_altvoltage1_TX1_I_F2_label - TX1_I_F2 is the extended name and will be seen as the channel name by libiio. This stuff seems to be addressed later on in `set_channel_name()` where we set the proper attr name. So, even if we handle things later and not in `add_attr_to_channel()`, not sure it would be simpler. So yeah, looks good as-is

local.c
```diff
@@ -835,6 +835,24 @@ static int add_attr_to_channel(struct iio_channel *chn,
...
+		dev_id = iio_device_get_id(iio_channel_get_device(chn));
+		ret = local_do_read_dev_attr(dev_id, 0, name,
+			label, sizeof(label), IIO_ATTR_TYPE_CHANNEL);
+		if (ret > 0) {
+			chn_label_ptr = label;
+			chn->label = iio_strdup(chn_label_ptr);
```
nunojsa, line 850: can this simply be `chn->label = iio_strdup(label);`?

device.c
```diff
@@ -139,6 +139,7 @@ struct iio_channel * iio_device_find_channel(const struct iio_device *dev,
 			continue;
 
 		if (!strcmp(chn->id, name) ||
+				(chn->label && !strcmp(chn->label, name)) ||
```
nunojsa, line 142: nothing wrong with the patch but commit subject does not match the style :)

nunojsa (commented):
LGTM, just some minor nits...

device.c
dNechita: do you mean that there is an extra tab on this line?

nunojsa: no.. the commit title itself. nothing wrong with the code. It's a nitpick but consistency across git log is a good thing. Also good for grepping the history

dNechita: Oh, I got it now. I need to grab another coffee.

local.c
dNechita: it can be. Thanks!

dNechita:
V6:

- Simplify code by using the `label `array and no longer use a pointer that was pointing to the same array
- Update the 3rd commit title to preserve the style (add hint of the files being affected by change)

nunojsa (approved):

PR #1237, opened 2025-01-29
Wrong IIO device attribute (and a iio_rwdev fix)

nunojsa:
## PR Description

Fix a device attribute being wrongly displayed and make sure iio_rwdev does not crash if we hit ctrl-c while streaming with a command as `iio_rwdev -u ip:192.168.1.79 -b 1048576 axi-adrv9002-rx-lpc | pv > /dev/null`

There's also a stylistic patch on local.c

## PR Type

## PR Checklist

### DIFF
--- a/local.c
+++ b/local.c
@@ -134,6 +134,7 @@ static void local_shutdown(struct iio_context *ctx)
 static void strcut(char *str, int nb)
 {
 	char *ptr = str + nb;
+
 	while (*ptr)
 		*str++ = *ptr++;
 	*str = 0;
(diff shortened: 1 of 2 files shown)


rgetz:
I would rather fix all the style issues in one massive commit with a tool - rather than patching things by hand piecemeal (one file at a time). - using git blame will be harder when we need to ignore multple random commits over time...

nunojsa:
Hmm, don't really agree. At least me, 99% (if not 100) of the time, I'll just blame one file to see what commit changed some line and I prefer `git show` giving me a small diff only related with that particular file rather than a huge diff with lots of unrelated changes.

Hardly applies in this case but if for some reason we wanted to revert the changes in just this file, we could easily revert the commit but if we touch multiple files at a time...

Finally, I didn't really wanted to go over all the other files or messing around with a tool to change the style everywhere. But I don't think that should be a show stopper for improving things in the sources I was working with.

PR #1238, opened 2025-01-31
Staging/unix thread names

nunojsa:
## PR Description

This adds support for naming threads in Unix (marked as TODO). The first commits are about fixing the names as they cannot be bigger than 16 chars. And one of the patches (on the iiod side) is actually a fix since the name of the main network thread was not being set because of the above.

This is kind of a RFC, since it depends on  pthread_setname_np() which is non portable (and needs _GNU_SOURCE to be defined) . Still, it should have no effect on windows and was tested under linux. We also check for the API availability in cmake but it would still be a good thing if someone could test this under MAC.

If it is yet not clear, this only adds support for Unix and does not touches on Windows builds.

## PR Type

## PR Checklist


nunojsa:
OK, this looks to be not reliable... Not sure if it will be doable at all because it looks like the API just has a different footprint in MAC and that is already discouraging.

iiod-responder.c
```diff
@@ -670,14 +670,14 @@ iiod_responder_create(const struct iiod_responder_ops *ops, void *d)
...
-					   "iiod-responder-writer-task");
+					   "writer-task");
 	err = iio_err(priv->write_task);
 	if (err)
 		goto err_free_io;
 
 	if (!NO_THREADS) {
 		priv->read_thrd = iio_thrd_create(iiod_responder_reader_thrd, priv,
-						  "iiod-responder-reader-thd");
+						  "reader-thd");
```
dNechita, line 680: Given the length constraint, I suggest removing "-thd" since it's already clear that we're naming threads. Instead, those four characters could be used to better describe the thread or distinguish it from others in relevant situations.

nunojsa: Hmm, it does make sense. I'll see if i can thing on something better... I'm reworking things a bit as I just realized (stupid me) that we do need a "pure" define in here and not something that evaluates to 0 or 1. If the API is not available and place inside an if(0) it will still fail to compile. Dead code removal happens in a later point in time...

nunojsa: Hmm thinking a bit more I can replace it for something like job (we do sabe 1 char). Or "client-reader" but here your argument also applies (we already now it's the client). One thing that could be beneficial would be to differentiate between the clients created on behave of buffer and the main client. But the trouble for that seemed not worth it for me.

nunojsa:
Ok, this should be now ready for review. The fact that pthreas_set_name_np() is non portable is far from great but I'm not sure how it can be done in way that tools like ps or top see the names (I mean, I know how I could do it on linux but not on other platforms). I tried anyways to keep things as self contained as possible but it would still be nice to see if this works on MAC (freebsd would also be nice but I guess more difficult)

dNechita:
I will try to do a test on MAC this week.

nunojsa:
> I will try to do a test on MAC this week.

@dNechita could you test this on Mac?

lock.c
```diff
@@ -27,9 +27,24 @@ struct iio_cond {
 struct iio_thrd {
 	pthread_t thid;
 	void *d;
+	/*
+	 * pthread_setname_np(3) does not allow names bigger than 16
+	 * (at least on linux).
+	 */
+	char name[16];
```
rgetz, line 34: I think you should be using `PTHREAD_MAX_NAMELEN_NP` rather than `16`. (its defined in `pthread.h`)

nunojsa: Hmm, actually with `PTHREAD_MAX_NAMELEN_NP` my build fails and it is not very clear why... pthread.h (though I indeed do not find the define in the header) is definitely included and __GNU_SRC is also defined (if it's needed). I might look a bit more into this but I do not plan in wasting too much time on it. Using raw 16 + the comment is fairly acceptable IMO.

AlexandraTrifan:
> > I will try to do a test on MAC this week.
> 
> @dNechita could you test this on Mac?

Tested this on a Mac M2 by attaching lldb to the iio_rwdev process and got:

`thread #4: tid = 0x10f89, 0x000000018568eb70 libsystem_kernel.dylib poll + 8, name = 'enqueue-worker'`

The same test on the **main** branch version didn't display any name for this thread:

`thread #4: tid = 0x11aaa, 0x000000018568eb70 libsystem_kernel.dylib poll + 8`

Is there anything else that needs to be verified on Mac for this PR?

-Alexandra

lock.c
rgetz: https://cwe.mitre.org/data/definitions/547.html

this is still shorter than your comment:
```
#ifndef PTHREAD_MAX_NAMELEN_NP
#define PTHREAD_MAX_NAMELEN_NP 16
#endif 
```

in the local iio-private.h or somewhere that is suitable.

rgetz: `PTHREAD_MAX_NAMELEN_NP` is part of [pthreads-win32 aka pthreads4w](https://sourceforge.net/projects/pthreads4w/) (only if `#if defined (__PTW32_COMPATIBILITY_BSD) || defined (__PTW32_COMPATIBILITY_TRU64)`) and on [netBSD](https://cvsweb.netbsd.org/bsdweb.cgi/~checkout~/src/lib/libpthread/pthread.h) (where it is defined as 32).

It's not defined in the GNU Source (that's why it was failing on locally on Linux), or the Ming implementation. 

That's why I was suggesting the ifdef - it's defined in some systems, some differently (NetBSD), and in the GNU Source, it's fixed as 16. (for now - as you point out).

There are already 128 #ifdef, and 66 #ifndef in the existing codebase - I don't think one more will hurt. You are already adding a bunch (which is the right way to do things - I'm not complaining).

```
#if defined(HAS_PTHREAD_SETNAME_NP)
#if defined(__APPLE__)
#define iio_thrd_create_set_name(thid, name)   pthread_setname_np(name)
#else
#define iio_thrd_create_set_name(thid, name)   pthread_setname_np(thid, name)
#endif
#else
#define iio_thrd_create_set_name(thid, name)   0
#endif
```

Moving that up in the file, so it's right after the `#includes`, and adding:

```
+#if defined(HAS_PTHREAD_SETNAME_NP)
+#ifndef PTHREAD_MAX_NAMELEN_NP
+#define PTHREAD_MAX_NAMELEN_NP 16
+#endif
+#if defined(__APPLE__)
...
```

I don't think is that bad to keep things compatible with NetBSD. (where we do have some users - there have been issues reported before).

I agree - it's a pretty minor things overall - If you don't want to do it - that's your/Dan's call.

nunojsa: > I don't think is that bad to keep things compatible with NetBSD. (where we do have some users - there have been issues reported before).

Alright, I did not went as deep as you but since it is already defined in platforms we support, that changes my mind over this... I will update the code

nunojsa:
The new version adds the changes suggested by @rgetz

dNechita (approved):

PR #1240, opened 2025-02-07
Add C++ compiler support for generic attr

Andrei-Fabian-Pop:
The C11 generic macros iio_attr_{read,write} cannot be accesed from a program built with C++. Instead C++ makes use of function overload to achieve the same behaviour.

This change does not influence the behaviour of libiio in C environments, but it allows the use of generic attr functions in C++ programs linked to libiio.

## PR Description

- Please replace this comment with a summary of your changes, and add any context
necessary to understand them. List any dependencies required for this change.
- To check the checkboxes below, insert a 'x' between square brackets (without
any space), or simply check them after publishing the PR.
- If you changes include a breaking change, please specify dependent PRs in the
description and try to push all related PRs simultaneously.

## PR Type

## PR Checklist

### DIFF
--- a/include/iio/iio.h
+++ b/include/iio/iio.h
@@ -362,11 +362,13 @@ iio_attr_read_raw(const struct iio_attr *attr, char *dst, size_t len);
  * @param ptr A pointer to a variable where the value should be stored
  * @return On success, 0 is returned
  * @return On error, a negative errno code is returned */
+#ifndef __cplusplus
 #define iio_attr_read(attr, ptr)				\
 	_Generic((ptr),						\
 		 bool *: iio_attr_read_bool,			\
 		 long long *: iio_attr_read_longlong,		\
 		 double *: iio_attr_read_double)(attr, ptr)
+#endif /* __cplusplus */
 
 /** @brief Set the value of the given attribute
  * @param attr A pointer to an iio_attr structure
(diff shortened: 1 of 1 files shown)


rgetz:
I'm confused - there is already a iiopp.h header - what's missing there? (if something is missing - all the c++ code should be kept in one place).

https://github.com/analogdevicesinc/libiio/blob/main/bindings/cpp/iiopp.h

dNechita:
> I'm confused - there is already a iiopp.h header - what's missing there? (if something is missing - all the c++ code should be kept in one place).
> 
> https://github.com/analogdevicesinc/libiio/blob/main/bindings/cpp/iiopp.h

He is not using the C++ bindings at all.
With libiio v0.x API, users could directly use the iio.h header in their C++ applications.
But this flexibility has been lost with v1.x API due to the use of C11 generic.

The proposed changes would only bring back that flexibility and not opening the gate to adding more C++ code in iio.h.

rgetz:
Still confused.

> He is not using the C++ bindings at all.
> With libiio v0.x API, users could directly use the iio.h header in their C++ applications.

then why do we continue to have the C++ bindings? If we aren't going to tell people to use them - the first patch (which is missing) should be to remove them completely - shouldn't it?

If we are going to continue to use the C++ bindings - then this patch doesn't make sense.

dNechita:
> Still confused.
> 
> > He is not using the C++ bindings at all.
> > With libiio v0.x API, users could directly use the iio.h header in their C++ applications.
> 
> then why do we continue to have the C++ bindings? If we aren't going to tell people to use them - the first patch (which is missing) should be to remove them completely - shouldn't it?
> 
> If we are going to continue to use the C++ bindings - then this patch doesn't make sense.

I think it makes sense not to add C++ code in the header of library written in C, especially since C++ bindings are available.
It would make sense to hide the API that uses the Generic feature for users attempting to use libiio directly in their C++ application.

Andrei-Fabian-Pop:
I created a new PR for this fix: #1244
This one can be closed.

PR #1243, opened 2025-02-11
Staging/iiod segfault

nunojsa:
## PR Description

This series fixes a bug where iiod crashes (very often actually). To reproduce the issue one just needs to start a buffering session (I tested with the network backend) with iio_rwdev and hit CTRL-C. IIOD crashed fairy easy and the reason is that blocks could be enqueued in the (d)enqueue tasks, then freed and only then the tasks would process the already freed block.

the first patch is a bit unrelated but it fixes some issues regarding locking of the iiod_io write_token. Then, it's all about preparing the path (re-working task.c) for the real fix in the last commit.

## PR Type

## PR Checklist

### DIFF
--- a/task.c
+++ b/task.c
@@ -355,6 +355,12 @@ bool iio_task_is_done(struct iio_task_token *token)
 	return token->done;
 }
 
+int iio_task_cancel_sync(struct iio_task_token *token, unsigned int timeout_ms)
+{
+	iio_task_cancel(token);
+	return iio_task_sync_core(token, timeout_ms, false);
+}
+
 void iio_task_cancel(struct iio_task_token *token)
 {
 	struct iio_task *task = token->task;
--- a/task.c
+++ b/task.c
@@ -184,10 +184,43 @@ err_free_task:
 	return iio_ptr(err);
 }
 
+static int iio_task_token_do_enqueue(struct iio_task *task, struct iio_task_token *token,
+				     bool autoclear)
+{
+	struct iio_task_token *tmp;
+
+	iio_mutex_lock(task->lock);
+
+	if (task->stop) {
+		iio_mutex_unlock(task->lock);
+		return -EBADF;
+	}
+
+	if (!task->list) {
+		task->list = token;
+	} else {
+		for (tmp = task->list; tmp->next; ) {
+			tmp = tmp->next;
+		}
+
+		tmp->next = token;
+	}
+
+	token->autoclear = autoclear;
+
+	iio_cond_signal(task->cond);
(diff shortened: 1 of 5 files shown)


nunojsa:
v2:

* Fix some indentation issues on iio_task_token_find()

task.c
```diff
@@ -94,6 +87,54 @@ static int iio_task_run(void *d)
 
 	return 0;
 }
+static int __iio_task_sync(struct iio_task_token *token, unsigned int timeout_ms, bool token_destroy)
```
dNechita, line 90: I believe it's best to avoid using a double underscore at the beginning of an identifier name, because it is reserved for system use.

dNechita:
The solution looks complex but at the same time necessary to make the IIOD stable. LGTM!

nunojsa:
> The solution looks complex but at the same time necessary to make the IIOD stable. LGTM!

Not seeing any other way. We could go with refcounts but I fear that would be more complex... Also looks like a good practice to let any ongoing block to finish the current operation... But IIOD actually blocks (takes a lot of time) sending a reply back to the client (the dequeued block IIRC) which seems there might be something to be improved on the client side.

task.c
nunojsa: Yeah, I can change that. Though I find it very unlikely to have any kind of conflict with the above symbol :)

dNechita:
I have also tested your changes with iio_rwdev and ad9361-iiostream and haven't encountered any issues. The IIOD seems stable.

nunojsa:
> I have also tested your changes with iio_rwdev and ad9361-iiostream and haven't encountered any issues. The IIOD seems stable.

Make sure to hit ctrl-c for exiting... That was triggering the issue for me. You'll likely see that it now takes sometime to exit the app and the blocking point is io_submit() on IIOD. Not really sure why... Anyways, there's a "fix" that can be done in iio_stream_destroy() on the client side which makes sense to me... I'll send it in another PR. But I'll still dig a bit more just to make sure thereś nothing fundamentally wrong in IIOD causing io_submit() to block for whatever reason.

nunojsa:
v3:

* Renamed __iio_task_sync();
* Fixed codacy complain (an actual deadlock);

nunojsa:
Anything missing on this one?

dNechita:
> Anything missing on this one?
No, this should be bine.
I wanted to test it on Windows but I am unable to successfully create a context. I'm currently debugging the issue. Since this PR doesn't touch source code that is only specific to Windows, I would go ahead and merge these changes.

dNechita (approved):

PR #1245, opened 2025-02-14
Misc improvements

nunojsa:
## PR Description

This improves iio_rwdev exit and is a combination of two things:

1. When destroying the stream, make sure to dequeue any possible in flight block before destroying it. This makes for a much cleaner exit specially in cases where freeing the blocks is deferred to the default client which for some reason will make iiod choke for sometime on a io_submit() call related to in flight block dequeue;
2. With 1), we don't really want to cancel the buffer as soon as we hit ctrl-c since then we would timeout in dequeing the block. And if we don't do 1), then we would still have the delay caused by iiod (since with #1243, iiod now makes sure it's safe to free a block before freeing it).

So this is really a combination of the above two points which improves things a lot when cancelling a buffer session with iio_rwdev.

Note this is only tested on linux so it would be great if someone could do a quick test on windows.

## PR Type

## PR Checklist

### DIFF
--- a/stream.c
+++ b/stream.c
@@ -74,8 +74,17 @@ void iio_stream_destroy(struct iio_stream *stream)
 	size_t i;
 
 	for (i = 0; i < stream->nb_blocks; i++)
-		if (stream->blocks[i])
+		if (stream->blocks[i]) {
+			/* Make sure to dequeue any possible in flight block
+			 * before destroying it. Note that if the block is not
+			 * enqueued, this returns an error but we really don't
+			 * care about it. Making sure the block is dequeued makes
+			 * for a much cleaner exit specially in cases where freeing
+			 * the blocks is deferred to the default client.
+			 */
+			iio_block_dequeue(stream->blocks[i], false);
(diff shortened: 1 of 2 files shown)


nunojsa:
Ups, I was testing this with other patches (part of open PRs)... But this should still work without them. I'll drop the unneeded commits.

dNechita:
I have been running a "quick" test on Windows for a while now. However, I haven't been able to test your changes yet due to some issues with the MSVC 2019 build. The issue occurs both with my local build and with the artifacts from Azure and I am trying to figure out what is causing this.

dNechita:
> > I have been running a "quick" test on Windows for a while now. However, I haven't been able to test your changes yet due to some issues with the MSVC 2019 build. The issue occurs both with my local build and with the artifacts from Azure and I am trying to figure out what is causing this.
> 
> @dNechita is this the same issue you mentioned on the other pull?

Yes.

nunojsa:
> Yes.

Alright, then I'll wait on this one until we can actually test it on windows

dNechita (approved):

Issue #1259, opened 2025-03-19
local backend with cma


xR3b0rn:
In the current version of **libiio**, the local backend allocates DMA buffers using [[/dev/dma_heap/system](https://github.com/analogdevicesinc/libiio/blob/9f24ce5700bc9f13959a830efd10723a427875de/local-dmabuf.c#L87)](https://github.com/analogdevicesinc/libiio/blob/9f24ce5700bc9f13959a830efd10723a427875de/local-dmabuf.c#L87), which results in a scattered DMA buffer using `sg_table`. However, some hardware requires **contiguous** DMA buffers. In such cases, it would be preferable to use `/dev/dma_heap/cma,linux` instead. Therefore, there should be a way to configure `libiio` to use CMA-based DMA buffers when needed.

dNechita:
Hi,
Would it be sufficient if this configuration were available at the time libiio is being built?

xR3b0rn:
For my current needs, however, it would be helpful to have a quick upstream solution where a simple CMake option would be sufficient.

dNechita:
Hi @xR3b0rn,

I made a draft with a possible implementation: https://github.com/analogdevicesinc/libiio/pull/1260
Can you please have a look and let me know if this works for you?
Thanks!

PR #1260, opened 2025-03-19
local: Expose ENV to allow switching the DMA buffers allocator

dNechita:
The default allocator is a generic allocator that does not garantee continous memory. Some hardware require contiguous DMA buffers. See #1259.

Set the environment variable CMA_FOR_DEVICES with a string that contains the name of the iio device for which it is desired the CMA allocator to be used. Multiple device names can be set.

## PR Type

## PR Checklist

### DIFF
--- a/local.c
+++ b/local.c
@@ -1484,12 +1484,20 @@ local_create_buffer(const struct iio_device *dev, unsigned int idx,
 	const struct iio_channel *chn;
 	int err, cancel_fd, fd;
 	unsigned int i;
+	const char *use_cma_for_device;
 
 	pdata = zalloc(sizeof(*pdata));
 	if (!pdata)
 		return iio_ptr(-ENOMEM);
 
 	pdata->dev = dev;
+
+	pdata->dmabuf_alloc_type = DMABUF_ALLOC_SC;
+	use_cma_for_device = iio_getenv("CMA_FOR_DEVICE");
+	if (use_cma_for_device) {
+		if (strstr(dev->name, use_cma_for_device))
+			pdata->dmabuf_alloc_type = DMABUF_ALLOC_CMA;
+	}
 
 	if (WITH_LOCAL_MMAP_API) {
 		pdata->pdata = local_alloc_mmap_buffer_impl();
--- a/local-dmabuf.c
+++ b/local-dmabuf.c
@@ -84,7 +84,11 @@ local_create_dmabuf(struct iio_buffer_pdata *pdata, size_t size, void **data)
 	if (!priv)
 		return iio_ptr(-ENOMEM);
 
-	devfd = open("/dev/dma_heap/system", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+	if (pdata->dmabuf_alloc_type == DMABUF_ALLOC_SC) {
+		devfd = open("/dev/dma_heap/system", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+	} else {
+		devfd = open("/dev/dma_heap/cma,linux", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+	}
 	if (devfd < 0) {
 		ret = -errno;
 
--- a/local.h
+++ b/local.h
@@ -17,12 +17,18 @@ struct iio_block_impl_pdata;
 struct iio_device;
 struct timespec;
 
+enum dma_allocator_type {
+	DMABUF_ALLOC_SC,  /* System allocator (generic - uses Scatter-Gather) */
+	DMABUF_ALLOC_CMA, /* Continuous memory allocator */
+};
+
 struct iio_buffer_pdata {
 	const struct iio_device *dev;
 	struct iio_buffer_impl_pdata *pdata;
 	int fd, cancel_fd;
 	unsigned int idx;
 	bool dmabuf_supported;
+	enum dma_allocator_type dmabuf_alloc_type;
 	bool mmap_supported;
 	size_t size;
 };


local.c
```diff
@@ -1484,13 +1484,21 @@ local_create_buffer(const struct iio_device *dev, unsigned int idx,
...
 		return iio_ptr(-ENOMEM);
 
 	pdata->dev = dev;
 
+	pdata->dmabuf_alloc_type = DMABUF_ALLOC_SC;
+	use_cma_for_device = iio_getenv("CMA_FOR_DEVICE");
+	if (use_cma_for_device) {
+		if (strstr(dev->name, use_cma_for_device))
+			pdata->dmabuf_alloc_type = DMABUF_ALLOC_CMA;
+	}
```
nunojsa, line 1500: This is now just a draft so I guess we can use the general ideal for testing purposes. I would be interested to understand if the underlying kernel code works (as-is) with the CMA dma-heap (It should but still...). 

Having said the above, I do not think this is the proper way... I'm afraid we need a dedicated API for choosing the heap. The reason is that this env variable is selecting devices while this setting should be per buffer. Note that the kernel supports multiple buffers (which can be DMA) per device and while very unlikely, it should be perfectly fine to have a buffer connected to a DMA controller doing CMA and the other doing SG.

Also, this only assumes two kind of heaps while there's nothing stopping us to add more in the future. This approach won't scale that nicely.

dNechita: I had a feeling that these changes will ask for an API change, but I didn't know about the possibility of having multiple buffers.
I have tested this but open() fails on my setup (zc706 with ADI image). The `errno` gets set to ENOENT even though running this shows:
```
analog@analog:~ $ dmesg | grep -i cma
[    0.000000] cma: Reserved 128 MiB at 0x38000000
[    0.000000] Memory: 881944K/1048576K available (12288K kernel code, 818K rwdata, 10376K rodata, 1024K init, 470K bss, 35560K reserved, 131072K cma-reserved, 131072K highmem)
```

I will be working on defining a proper API but in the meantime, as you said, this can be used for testing things.
Thanks @nunojsa

nunojsa: > I have tested this but open() fails on my setup (zc706 with ADI image). The errno gets set to ENOENT even though running this shows:

Likely because that module is not being compiled atm

nunojsa:
> From the driver's perspective, I always receive an sg_table. However, when allocating 4MB of DMA memory using /dev/dma_heap/system, the sg_table contains 4 entries of 1MB each. In contrast, when using /dev/dma_heap/cma,linux, the sg_table contains a single entry of 4MB.

Yeah, I figured the dma-heap driver would do something like that to keep the interface consistent for dma buffers. Thanks for testing this @xR3b0rn

dNechita:
> Is the API for version 1.0 finalized yet? It's important for us to have this feature in the mainline. I might find some time to implement it myself.
> 
> I'm thinking of something similar to the `params` logic used by `create_context`, which would effectively change the API for `create_buffer`.

It's either a structure similar to `iio_context_params` or another parameter for the call creating a buffer.
With the structure approach, the buffer index could also be included in the structure. This method appears to be more scalable for accommodating additional parameters that may arise in the future.
I was planning to start working on this at the beginning of next week. But let me know by then if you want to give it a try.

dNechita:
> When adding an `iio_buffer_params` struct, it might also make sense to introduce similar structs for all other `iio_*_create_*` functions before version 1.0 — just for the sake of consistency. :3

It would be nice, but it's not sufficient. I prefer to add them when they are actually needed.

nunojsa:
> It would be nice, but it's not sufficient. I prefer to add them when they are actually needed.

I agree with this... Always better to introduce an API with actual users of it.

xR3b0rn:
My initial thought wasn't just about consistency—it was also about avoiding the need to increment the major version when adding new options, by introducing such structs for the other *_create_* functions as well. If you don’t do that, you’ll have to bump the major version every time a new option is added.

rgetz:
>I will be working on defining a proper API but in the meantime, as you said, this can be used for testing things.

Hmm - I'm hesitant to encourage such hardware specific parts in libiio. I get why it's needed - but you don't want to have different applications failing to work because they were targeted at one hardware vs a different hardware (ie. exposing different APIs)... (ie libALSA always works - it abstracts the hardware for people).

my 2 cents.

nunojsa:
> Hmm - I'm hesitant to encourage such hardware specific parts in libiio. I get why it's needed - but you don't want to have different applications failing to work because they were targeted at one hardware vs a different hardware (ie. exposing different APIs)... (ie libALSA always works - it abstracts the hardware for people).

But that's already the case AFAICT... In this case, while the same application works in our platforms, it does not work for @xR3b0rn. As it stands, it's already very HW dependent. I agree we should keep the default choice the same as of today but I'm also not seeing any problem in trying to extend the API to support other platforms (agree it can be dangerous as it allows misuse). 

But I agree that ideally we should be able to detect HW capabilities and decide accordingly. But I'm not sure of a sane way for doing that...

xR3b0rn:
> > I will be working on defining a proper API but in the meantime, as you said, this can be used for testing things.
> 
> Hmm - I'm hesitant to encourage such hardware specific parts in libiio. I get why it's needed - but you don't want to have different applications failing to work because they were targeted at one hardware vs a different hardware (ie. exposing different APIs)... (ie libALSA always works - it abstracts the hardware for people).
> 
> my 2 cents.

If your driver or hardware isn't able to handle all DMA buffer sizes, a libiio application with buffer size of 256 might run on one platform, but will fail on another. libiio is already HW dependent.

rgetz:
> But I agree that ideally we should be able to detect HW capabilities and decide accordingly.

Then we just need to define something in the kernel - that is similar to lib v4l2 inferface ...

```
v4l2-ctl --device=/dev/video0 --list-ctrls
v4l2-ctl --device=/dev/video0 --list-formats-ext
```

we could have an interface that exposes capabilities from the kernel if needed. Other subsystems do it.

https://www.kernel.org/doc/html/v4.9/media/uapi/v4l/vidioc-g-fmt.html

dNechita:
Closing as the full implementation was done in PR: https://github.com/analogdevicesinc/libiio/pull/1264

PR #1264, opened 2025-03-25
task: Implement iio_buffer_params struct and make CMA selectable at runtime

xR3b0rn:
See #1260

Therefor this PR impelments the possibility to decide which block is using CMA at runtime using a newly introduce `struct iio_block_params` struct.

Reviewing of the general idea requested.

### DIFF
--- a/include/iio/iio.h
+++ b/include/iio/iio.h
@@ -156,6 +156,34 @@ struct iio_context_params {
 	char __rsrv[32];
 };
 
+/**
+ * @enum iio_buffer_dma_allocator
+ * @brief Provides low-level control over DMA buffer allocation.
+ */
+enum iio_buffer_dma_allocator {
+	/** @brief May use scattered DMA memory (default). */
+	IIO_DMA_ALLOCATOR_SYSTEM = 0,
+
+	/** @brief Enforces contiguous DMA memory (for expert use only). */
+	IIO_DMA_ALLOCATOR_CMA_LINUX = 1,
+};
+
+/**
+ * @struct iio_buffer_params
+ * @brief Used to pass parameters to the iio_device_create_buffer function
+ */
+struct iio_buffer_params {
+	/** @brief The index of the hardware buffer. Should be 0 in most cases.
+	 * (default: 0) */
+	unsigned int idx;
+
+	/** @brief Use the allocator to allocate a DMA buffer.
+	 * In most cases, this should be set to IIO_DMA_ALLOCATOR_SYSTEM.
+	 * ONLY change this to a different option if you fully understand the
+	 * implications. (default: IIO_DMA_ALLOCATOR_SYSTEM) */
+	enum iio_buffer_dma_allocator dma_allocator;
+};
+
 /*
  * <linux/iio/types.h> header guard to protect these enums from being defined
  * twice
--- a/include/iio/iio.h
+++ b/include/iio/iio.h
@@ -1080,12 +1080,15 @@ iio_buffer_find_attr(const struct iio_buffer *buf, const char *name);
 
 /** @brief Create an input or output buffer associated to the given device
  * @param dev A pointer to an iio_device structure
- * @param idx The index of the hardware buffer. Should be 0 in most cases.
- * @param mask A pointer to an iio_channels_mask structure
+ * @param params A pointer to an iio_buffer_params structure.
+ * @param mask A pointer to an iio_channels_mask structure. If NULL is passed,
+ *        default values are used, for more information see comments in
+ *        struct iio_buffer_params.
  * @return On success, a pointer to an iio_buffer structure
  * @return On failure, a pointer-encoded error is returned */
 __api __check_ret struct iio_buffer *
-iio_device_create_buffer(const struct iio_device *dev, unsigned int idx,
+iio_device_create_buffer(const struct iio_device *dev,
+			 struct iio_buffer_params *params,
 			 const struct iio_channels_mask *mask);
 
 
--- a/local-dmabuf.c
+++ b/local-dmabuf.c
@@ -78,13 +78,20 @@ local_create_dmabuf(struct iio_buffer_pdata *pdata, size_t size, void **data)
 		.fd_flags = O_CLOEXEC | O_RDWR,
 	};
 	struct iio_block_pdata *priv;
-	int ret, fd, devfd;
+	int ret, fd, devfd = -1;
 
 	priv = zalloc(sizeof(*priv));
 	if (!priv)
 		return iio_ptr(-ENOMEM);
 
-	devfd = open("/dev/dma_heap/system", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+	switch (pdata->params->dma_allocator) {
+	case IIO_DMA_ALLOCATOR_SYSTEM:
+		devfd = open("/dev/dma_heap/system", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+		break;
+	case IIO_DMA_ALLOCATOR_CMA_LINUX:
+		devfd = open("/dev/dma_heap/linux,cma", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+		break;
+	}
 	if (devfd < 0) {
 		ret = -errno;
 
--- a/utils/iio_rwdev.c
+++ b/utils/iio_rwdev.c
@@ -223,6 +223,7 @@ int main(int argc, char **argv)
 	struct iio_stream *stream;
 	const struct iio_block *block;
 	struct iio_channels_mask *mask;
+	struct iio_buffer_params buffer_params = {0};
 	const struct iio_channels_mask *hw_mask;
 	const struct iio_attr *uri, *attr;
 	struct option *opts;
@@ -437,7 +438,7 @@ int main(int argc, char **argv)
 		goto err_free_mask;
 	}
 
-	buffer = iio_device_create_buffer(dev, 0, mask);
+	buffer = iio_device_create_buffer(dev, &buffer_params, mask);
 	ret = iio_err(buffer);
 	if (ret) {
 		dev_perror(dev, ret, "Unable to allocate buffer");
--- a/utils/iio_rwdev.c
+++ b/utils/iio_rwdev.c
@@ -41,12 +41,13 @@ static const struct option options[] = {
 	  {"write", no_argument, 0, 'w'},
 	  {"cyclic", no_argument, 0, 'c'},
 	  {"benchmark", no_argument, 0, 'B'},
+	  {"cma", no_argument, 0, 'C'},
 	  {0, 0, 0, 0},
 };
 
 static const char *options_descriptions[] = {
 	"[-t <trigger>] [-b <buffer-size>]"
-		"[-s <samples>] <iio_device> [<channel> ...]",
+		"[-s <samples>] <iio_device> [<channel> ...] [-C]",
 	"Use the specified trigger.",
 	"Set the trigger to the specified rate (Hz). Default is 100 Hz.",
 	"Size of the transfer buffer. Default is 256.",
@@ -56,6 +57,7 @@ static const char *options_descriptions[] = {
 	"Use cyclic buffer mode.",
 	"Benchmark throughput."
 		"\n\t\t\tStatistics will be printed on the standard input.",
+	"Use CMA-Linux allocator for DMA buffer."
 };
 
 static struct iio_context *ctx;
@@ -205,7 +207,7 @@ static ssize_t transfer_sample(const struct iio_channel *chn,
 	return (ssize_t) nb;
 }
 
-#define MY_OPTS "t:b:s:T:r:wcB"
+#define MY_OPTS "t:b:s:T:r:wcBC"
 
 int main(int argc, char **argv)
 {
@@ -294,6 +296,9 @@ int main(int argc, char **argv)
 			break;
 		case 'w':
 			is_write = true;
+			break;
+		case 'C':
+			buffer_params.dma_allocator = IIO_DMA_ALLOCATOR_CMA_LINUX;
 			break;
 		case '?':
 			printf("Unknown argument '%c'\n", c);
--- a/utils/iio_stresstest.c
+++ b/utils/iio_stresstest.c
@@ -226,6 +226,7 @@ static void *client_thread(void *data)
 	const struct iio_device *dev;
 	const struct iio_channel *ch;
 	struct iio_channels_mask *mask;
+	struct iio_buffer_params buffer_params = {0};
 	struct timeval start, end;
 	int id = -1, stamp, r_errno;
 	ssize_t ret;
@@ -307,10 +308,11 @@ static void *client_thread(void *data)
 		if (info->verbose == VERYVERBOSE)
 			printf("%2d: Running\n", id);
 
+		buffer_params.idx = info->buffer_size;
 		i = 0;
 		while (threads_running || i == 0) {
 			info->buffers[id]++;
-			buffer = iio_device_create_buffer(dev, info->buffer_size, false);
+			buffer = iio_device_create_buffer(dev, &buffer_params, false);
 			ret = iio_err(buffer);
 			if (ret) {
 				struct timespec wait;
@@ -398,6 +400,7 @@ int main(int argc, char **argv)
 	struct iio_buffer *buffer;
 	struct iio_context *ctx;
 	struct iio_channel *ch;
+	struct iio_buffer_params buffer_params = {0};
 	struct iio_channels_mask *mask;
 	const char *name;
 	int c, pret, option_index, err;
@@ -492,7 +495,7 @@ int main(int argc, char **argv)
 							iio_channel_enable(ch, mask);
 					}
 
-					buffer = iio_device_create_buffer(dev, 0, mask);
+					buffer = iio_device_create_buffer(dev, &buffer_params, mask);
 					if (!iio_err(buffer)) {
 						iio_buffer_destroy(buffer);
 
--- a/utils/iio_stresstest.c
+++ b/utils/iio_stresstest.c
@@ -103,19 +103,21 @@ static const struct option options[] = {
 	{"samples", required_argument, 0, 's' },
 	{"duration", required_argument, 0, 'd'},
 	{"threads", required_argument, 0, 't'},
+	{"cma", no_argument, 0, 'C'},
 	{"verbose", no_argument, 0, 'v'},
 	{0, 0, 0, 0},
 };
 
 static const char *options_descriptions[] = {
 	("[-n <hostname>] [-u <vid>:<pid>] [-t <trigger>] [-b <buffer-size>] [-s <samples>]"
-		"<iio_device> [<channel> ...]"),
+		"<iio_device> [<channel> ...] [-C]"),
 	"Show this help and quit.",
 	"Use the context at the provided URI.",
 	"Size of the capture buffer. Default is 256.",
 	"Number of samples to capture, 0 = infinite. Default is 0.",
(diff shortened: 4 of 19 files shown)


mhennerich:
Hi Julian,

This PR touches some files which are subject to relicensing -
Please see here:
https://github.com/analogdevicesinc/libiio/discussions/1210

If we're going to accepts this contribution, which I think is great.
We need your consent as well.

-Michael

include/iio/iio.h
```diff
@@ -1142,7 +1151,8 @@ iio_buffer_get_channels_mask(const struct iio_buffer *buf);
  * @return On success, a pointer to an iio_block structure
  * @return On failure, a pointer-encoded error is returned */
 __api __check_ret struct iio_block *
-iio_buffer_create_block(struct iio_buffer *buffer, size_t size);
+iio_buffer_create_block(struct iio_buffer *buffer,
+          struct iio_block_params *params);
```
dNechita, line 1155: While it made sense to group `continuous_dma_memory_required `with `idx `(belonging to iio buffer) it is not the same for the `size `(belonging to the iio block). `size` is a mandatory parameter to be used each time a block is created which is not the case for `continuous_dma_memory_required`.
Please add a new parameter to the `iio_buffer_create_block()` and `iio_buffer_create_stream()` that is an enumeration. The first item in the enumeration should be the default system allocator and the 2nd should be the CMA allocator.
This way we can extend to more allocators if we need to.

local-dmabuf.c
```diff
@@ -84,7 +84,11 @@ local_create_dmabuf(struct iio_buffer_pdata *pdata, size_t size, void **data)
 	if (!priv)
 		return iio_ptr(-ENOMEM);
 
-	devfd = open("/dev/dma_heap/system", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+  if (params->continuous_dma_memory_required) {
+	  devfd = open("/dev/dma_heap/cma,linux", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+  } else {
+	  devfd = open("/dev/dma_heap/system", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+  }
```
dNechita, line 91: This behavior should be enhanced such if CMA doesn't work it then falls back to using the `system` and an error message is printed.

utils/iio_rwdev.c
```diff
@@ -46,10 +46,16 @@ static const struct option options[] = {
 
 static const char *options_descriptions[] = {
 	"[-t <trigger>] [-b <buffer-size>]"
+#if WITH_LOCAL_DMABUF_API == 1
+    "[-C] "
+#endif
```
dNechita, line 51: It makes more sense to eliminate the `if` directives from both iio_rwdev.c and iio_stresstest.c. This PR only addresses the local context, but at some point, these changes need to be extended for the other contexts (network, usb, serial, xml). And for that uses cases, when running iio_rwdev on a system where local context is not available, then WITH_LOCAL_DMABUF_API won't be defined (it's not applicable there) but it could be defined on the target device where iiod is running.
Therefore, it's best to keep the cmdline argument all the time and users will receive an error message in the scenario where CMA is not available.

dNechita (changes requested):

nunojsa:
@xR3b0rn, does it really make sense to have this per block? I mean, can we attach a block with a CMA backing storage and another block with SG on the same buffer? If the DMA device backing that buffer supports both, then I guess we're fine but if doesn't, won't we have issues? All to say that, shouldn't this be a per buffer parameter? I would say that it does not make sense  to have blocks with different heaps on the same buffer anyways (at least I'm not seeing the point of that)...

xR3b0rn:
> @xR3b0rn, does it really make sense to have this per block? I mean, can we attach a block with a CMA backing storage and another block with SG on the same buffer? If the DMA device backing that buffer supports both, then I guess we're fine but if doesn't, won't we have issues? All to say that, shouldn't this be a per buffer parameter? I would say that it does not make sense to have blocks with different heaps on the same buffer anyways (at least I'm not seeing the point of that)...

I have no strong opinion on this topic. The main reason I implemented it on a per-block basis is your comment [[here](https://github.com/analogdevicesinc/libiio/pull/1260#discussion_r2005014750)](https://github.com/analogdevicesinc/libiio/pull/1260#discussion_r2005014750). But since you're asking me this question, I guess I misunderstood something.

If you ask me whether it makes sense, I’d say it completely depends on the hardware and how the driver is implemented. In theory, you could pass scattered DMA buffers with different scatter granularities, including a granularity of one. However, I’m not sure why you would. That said, I’m not very experienced in this area, so my opinion may not carry much weight.

nunojsa:
> But since you're asking me this question, I guess I misunderstood something

"...The reason is that this env variable is selecting devices while this setting should be per buffer..." - Per buffer :)

> If you ask me whether it makes sense, I’d say it completely depends on the hardware and how the driver is implemented. In theory, you could pass scattered DMA buffers with different scatter granularities, including a granularity of one. However, I’m not sure why you would. That said, I’m not very experienced in this area, so my opinion may not carry much weight.

Well, but your experience can also be important because you were the one needing CMA. So why did you needed it in your system? Does the DMA controller on your system only handles CMA memory? From the below I would assume that's the case:

"However, some hardware requires contiguous DMA buffers"

So in the driver each IIO DMA buffer will be associated with a DMA channel and then IIO blocks are attached to this buffer. So you see my point that if the DMA channel backing a buffer needs contiguous memory there's no real point in being able to configure things per block. In fact, for such a scenario it should not even work... It would also be very strange to have blocks with different backing storage attached to the same buffer. IOW, I believe that the *possible* restriction on the memory allocation type is per buffer...

xR3b0rn:
My underlying hardware clearly doesn't support scattered DMA lists, so from my perspective, a per-buffer solution works fine. However, if your hardware does support scattered DMA buffers, it is possible to pass scattered buffers with a size of one. Therefore, if you need to pass scattered lists for one block and a contiguous buffer for another, it is entirely feasible.

Since I prefer not to make assumptions about people's intentions, even if it may seem dumb af, the longer I think about it, the more I would suggest a per-block solution, simply because it is possible.

nunojsa:
> Therefore, if you need to pass scattered lists for one block and a contiguous buffer for another, it is entirely feasible

It __might__ be possible but why would you do that in practice? And the tricky part is for HW that only support one type of backing storage then you open the door to shoot yourself in the foot. 

> Since I prefer not to make assumptions about people's intentions, even if it may seem dumb af, the longer I think about it, the more I would suggest a per-block solution, simply because it is possible.

I'm really not convinced about that... The key distinction is that it **might** be possible but we have no guarantee of that. Also, I'm really not seeing any point of having this kind of flexibility as I cannot see any usecase where one would want to have blocks with different backing storage types. So, I would start from the basics and only care about something more fine tuned if we ever really need it. 

In fact, the only sane way to ever support a per block approach would be if we could somehow detect what the underlying HW can handle and construct a mask so we could match against it when creating a new block. Right now, we should just stick to one kind of memory (that we need to know in advance) and have all blocks stick with it. There's nothing stopping us form enhancing this in the future so instead of starting with the more flexible approach, I would start with a simpler (and stricter) one.

dNechita:
I agree that the simpler solution is the one to go with.

xR3b0rn:
> There's nothing stopping us form enhancing this in the future so instead of starting with the more flexible approach
I think this implies an API-breaking change, so how easy it is to implement this after 1.0 becomes a strategic decision.

But apart from that I'm totally find with that.

local-dmabuf.c
xR3b0rn: Why Should There Be a Fallback?

1. I couldn’t find anything like that in **libiio** (nothing in **local-dmabuf** prints anything; everything either succeeds or returns an error code).  
2. If the underlying hardware requires CMA, there is no point in attempting to handle misuse — the whole process should simply fail.

dNechita: Your points make sense. I am ok with them.
To answer your question, a fallback seems useful due to the fact that I see the "/dev/dma_heap/system" as the default option and "cma,linux" as a configuration that is used less frequent. Therefore, I thought some users will be tempted to set it without knowing it could brake stuff and the fallback would act both as a warning and prevent things from breaking.

nunojsa: Assuming we have a function with the params stuff, we need to have the main api like `create_foo_with_params(struct foo *foo, struct iio_foo_params *params)` and I would definetly suggest to have an helper like:

```
static inline int create_foo(struct foo *foo)
{
    struct iio_foo_params params = { DEFAULT };

    return create_foo_with_params(foo, &params);
}
```

And just use the default API option as of today. Then, we **really** need to document that one really needs to know what he's going when messing with this since that might heavily depend on HW capabilities.

So I do understand your point about a fallback but this is such a low level thing that if someone explicitly uses it, better to know what he's doing.... To sum up, my 2 cents: If we properly document this as a very "specialized" thing, I would not bother with the extra effort of a fallback.

xR3b0rn:
In this version, I have done my best to apply all requested changes. I also attempted to adapt the (incomplete?) bindings to the new API as well as possible.

Before merging, please give me a chance to clean up the commit history.

dNechita:
Could you, please, split the commit into several commits to better group the changes you are making as they are touching lots of files. Take a look at the title style of other commits as well. In addition, we need you to sign off the commits (`git commit -s`). And last, please give your consent on the page mentioned early in this PR.
Thank you!

xR3b0rn:
> Could you, please, split the commit into several commits to better group the changes you are making as they are touching lots of files. Take a look at the title style of other commits as well. In addition, we need you to sign off the commits (`git commit -s`). And last, please give your consent on the page mentioned early in this PR. Thank you!

I had to modify many files solely due to the API change, as its nature affected a large number of them. Splitting this into multiple commits would result in intermediate commits that do not build properly, which I see no benefit in. In my opinion, this is the smallest self-contained change that remains buildable.

xR3b0rn:
@dNechita done

utils/iio_stresstest.c
```diff
@@ -307,10 +310,11 @@ static void *client_thread(void *data)
 		if (info->verbose == VERYVERBOSE)
 			printf("%2d: Running\n", id);
 
+		buffer_params.idx = info->buffer_size;
```
dNechita, line 313: What is the purpose of this?

xR3b0rn: Idk, the original line lookedl ike this: `buffer = iio_device_create_buffer(dev, info->buffer_size, false);`

I just kept it that way.

dNechita: Oh, ok then. I will check it out at some point. But for this PR it's fine.

xR3b0rn:
@dNechita I have just given my consent.

dNechita:
> @dNechita I have just given my consent. Furthermore, the CMA functionality has been tested on my machine and works as expected with both `iio_rwdev` and when linking against `libiio`. Please let me know if you need anything else.

Thank you for contributing!

dNechita (approved):

nunojsa:
@dNechita, @xR3b0rn I see this is merged and I'm late to the party but I think we may have issues with this... @xR3b0rn how did you tested this? With a local test app? As this stands I'm not sure how things will properly work if we use a backend other than the local... I mean, we are still not able to select CMA.

rgetz:
Just going on record - I think this is a bad design decision, and it going to make things more complicated to use going forward. The goal of the kernel is to abstract hardware from userspace, and allow common code to be run on any architecture. This is the opposite of that.

xR3b0rn:
@nunojsa I’ve only tested it with the local backend. Honestly, I really think you should consider splitting the core parts of `libiio` from all the extra fancy stuff. I’m seriously thinking about switching to a different library — or even writing my own — simply because `libiio` violates the “do one thing and do it well” principle. This discussion clearly highlights what needs to be considered when adapting to new kernel features. Just look at the amount of files I had to edit to only implement one simple `if`, that's cary.

@rgetz A bit late to the party.

nunojsa:
> Just going on record - I think this is a bad design decision, and it going to make things more complicated to use going forward. The goal of the kernel is to abstract hardware from userspace, and allow common code to be run on any architecture. This is the opposite of that.

I understand that but even before this it was pretty much HW dependent. And, IIRC, this is likely not a big deal because if we can use CMA (or SG for that matter) we'll try to fallback to the old mmap based interface. And if all the high speed thing fails we should still have, as a last, resort file based IO. 

Ideally we should be able to detect HW capabilities and select accordingly but there's no why for that.... 

That said,  I do think we should have an helper API (basically the same function as before) that just uses the default implementation (basically just taking the buffer index as argument) and then have the real API named `iio_create_buffer_with_params()`. The reason is that using the parameters should be taken as **very advanced** usage and 99% of the time you should stick with the regular API.

nunojsa:
> @nunojsa I’ve only tested it with the local backend. 

I figured... well, things won't work with the other backends AFAICT...

> And honestly, I really think you should consider splitting the core parts of libiio from all the extra fancy stuff

Well, we don't really want to be doing any generic (and yet another) RPC library... This also goes way back to the beginning of libiio (I guess). Even if not ideal, sometimes you do need to stick we some "not so good" decisions just because you'll either break current users or because it will just take a very big effort to change things (risking lots of new issues)

> Why the hell, do I have to compile something what has DNS in its name, just to write an IIO application? 

You don't... just disable that in cmake. You still need to make sure that a new API works for all supported backends (at least on the upstream project). And, AFAIK, you should be able to strip everything down and only have the local backend enabled (if that is not the case today, it should be)

> Just look at the number of files I had to touch just to implement a single if — that’s honestly crazy and a clear sign of overdesign.

Well, I do agree there's a bit of overdesign and it is not easy to get things for new comers (it was the same for me when I looked more seriously into the code the last few months) but OTOH, this is really not intended to be just a "simple" library to run on your local board. If you wanna support different "backends" (and lot's of users just use USB or the network to stream data) you need to have abstractions in place. And for a fundamental option like this (creating a buffer), it needs to work on all supported backends.

And then, there's also the bindings which are useful to lot's of people... So yeah, I get the frustration but I don't see how it could be dramatically better if you wan't to support all the things libiio is supporting.

All that said, my 2 cents... either things get fixed or I would revert the change (but not my decision anyways)

xR3b0rn:
@nunojsa sry, my rant was rather unprofessionl.

I tested the feature using the local backend only. If you are interested in further investigations on how we use `dma-buf` in our driver, you can check out the driver [here](https://github.com/primes-gmbh/raspberrypi-linux/blob/b6c9c0b011f56bd71b82d736fcbb6bb3d1293c26/drivers/media/platform/primes-bcm2835/bcm2835-unicam-main.c#L189).

kind regards

dNechita:
@xR3b0rn your contribution is valuable, and I will make sure everything works for the other backends as well.

nunojsa:
> I tested the feature using the local backend only. If you are interested in further investigations on how we use dma-buf in our driver, you can check out the driver [here](https://github.com/primes-gmbh/raspberrypi-linux/blob/b6c9c0b011f56bd71b82d736fcbb6bb3d1293c26/drivers/media/platform/primes-bcm2835/bcm2835-unicam-main.c#L189).

The issue is not the kernel... Though I do see some serious abuse of the IIO subsystem in your driver :wink:.

The issue is here:

https://github.com/analogdevicesinc/libiio/blob/main/iiod/responder.c#L432

and also, of course, on the client side where we need to propagate all the buffer parameters. 

> @nunojsa sry, my rant was rather unprofessionl.

No worries... Though, I have to admit that I got a bit emotional as well :sweat_smile: 

The most straightforward solution is to introduce an helper struct to pack both the parameters and the channels mask and send that to the devices (likely we need to pack the elements or make sure there's no padding).

But maybe there are more elegant alternatives...

xR3b0rn:
> > I tested the feature using the local backend only. If you are interested in further investigations on how we use dma-buf in our driver, you can check out the driver [here](https://github.com/primes-gmbh/raspberrypi-linux/blob/b6c9c0b011f56bd71b82d736fcbb6bb3d1293c26/drivers/media/platform/primes-bcm2835/bcm2835-unicam-main.c#L189).
> 
> The issue is not the kernel... Though I do see some serious abuse of the IIO subsystem in your driver 😉.
> 
> The issue is here:
> 
> https://github.com/analogdevicesinc/libiio/blob/main/iiod/responder.c#L432
> 
> and also, of course, on the client side where we need to propagate all the buffer parameters.
> 
> > @nunojsa sry, my rant was rather unprofessionl.
> 
> No worries... Though, I have to admit that I got a bit emotional as well 😅
> 
> The most straightforward solution is to introduce an helper struct to pack both the parameters and the channels mask and send that to the devices (likely we need to pack the elements or make sure there's no padding).
> 
> But maybe there are more elegant alternatives...

I'm very interested in your feedback. If you can spare the time, please send it to j.weiss@primes.de

nunojsa:
> I'm very interested in your feedback. If you can spare the time, please send it to j.weiss@primes.de

Unfortunately not in the near future... Lot's of things ongoing

xR3b0rn:
> > I'm very interested in your feedback. If you can spare the time, please send it to [j.weiss@primes.de](mailto:j.weiss@primes.de)
> 
> Unfortunately not in the near future... Lot's of things ongoing

How rude — pointing out serious abuses without actually telling me what they are! 😄

xR3b0rn:
@nunojsa Thanks! I'm fully aware that this might look a bit odd at first glance. This driver isn’t meant to be merged into mainline, as there’s an FPGA on the other side of the MIPI connection pulling data from a high-speed ADC (pulling 5 channels a 16bit @10MHz)— so without the FPGA bitstream, the driver is essentially useless.

I did consider using the V4L2 subsystem, but since the FPGA is transmitting raw ADC values, the IIO subsystem felt like a more natural fit. That said, after seeing the kind of discussion this triggered, I’ve started reconsidering and might just stick with V4L2 after all.

What’s still holding me back, though, is the complexity involved in separating the control logic from the actual media driver using V4L2 pads and so on. I feel like IIO handles this in a much more straightforward and manageable way.

nunojsa:
> @nunojsa Thanks! I'm fully aware that this might look a bit odd at first glance. This driver isn’t meant to be merged into mainline, as there’s an FPGA on the other side of the MIPI connection pulling data from a high-speed ADC (pulling 5 channels a 16bit @10MHz)— so without the FPGA bitstream, the driver is essentially useless.

> I did consider using the V4L2 subsystem, but since the FPGA is transmitting raw ADC values,

There's a couple of options you could consider:

1) You can model your FPGA IP as a platform device and implement the IIO bits in there.
2)  Then, if you need to somehow control the channels of the ADC from your camera, you can use the IIO [inkern](https://elixir.bootlin.com/linux/v6.14-rc6/source/drivers/iio/inkern.c) API to get the channels.
3) Alternatively, You could see your device as an aggregate [multi function device](https://elixir.bootlin.com/linux/v6.14-rc6/source/drivers/mfd).

There's also an [IIO backend API](https://elixir.bootlin.com/linux/v6.14-rc6/source/drivers/iio/industrialio-backend.c) where we typically implement our FPGA cores that interface with another devices (but typically the frontends are also IIO devices so not sure if would much of an help for you).

The above might still not be the proper way of handling this... Just some stuff to consider :)

xR3b0rn:
Thanks a lot for these references!!! I will check them out once I find time.

rgetz:
> Ideally we should be able to detect HW capabilities and select accordingly but there's no why way for that....

Why not? Other hardware subsystems do it - we just need to define something in the kernel - that is similar to lib v4l2 inferface ...

```
v4l2-ctl --device=/dev/video0 --list-ctrls
v4l2-ctl --device=/dev/video0 --list-formats-ext
```
we could have an interface that exposes capabilities from the kernel if needed. Similar to:

https://www.kernel.org/doc/html/v4.9/media/uapi/v4l/vidioc-g-fmt.html

rgetz:
> Well, just saying right now there's nothing like that.

Today.

If no one starts working on it - it will never get done. That's all I'm saying.

PR #1273, opened 2025-04-14
Modify qemu

AAndrisa:
## PR Description

Remove qemu package and install qemu-user-static

## PR Type

## PR Checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -386,9 +386,6 @@ stages:
     - script: |
         set -e
         sudo apt-get update
-        sudo apt-get install -y gcc-arm-linux-gnueabihf libc6-dev-armhf-cross
-        sudo apt-get install -y g++-arm-linux-gnueabihf
-        sudo apt-get install -y g++-aarch64-linux-gnu
         sudo apt-get install -y qemu-system-ppc64
         sudo apt-get install qemu-user-static
         sudo docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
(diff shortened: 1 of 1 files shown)


rgetz:
> 2\. The GNU toolchain always crashes while building libiio for arm architecture ubuntu-ppc64le and crashes once in a while for other ARM architectures.

Is this on specific versions of gcc? for now - you should be able to go back a version to unblock things...

```
sudo apt-cache search ^gcc-9-
sudo apt-git install ...(whatever packages you need)
```

and report a problem/bug in gcc - https://gcc.gnu.org/bugs/

dNechita:
> > 2. The GNU toolchain always crashes while building libiio for arm architecture ubuntu-ppc64le and crashes once in a while for other ARM architectures.
> 
> Is this on specific versions of gcc? for now - you should be able to go back a version to unblock things...
> 
> ```
> sudo apt-cache search ^gcc-9-
> sudo apt-git install ...(whatever packages you need)
> ```
> 
> and report a problem/bug in gcc - https://gcc.gnu.org/bugs/

I have already tried with multiple of gcc versions and all have failed.
I have forced the build to use just one core and even forced the docker container to use 4GB of RAM but didn't work.
I have manually built lastest qemu from source but didn't work.

In the end, it worked by disabling Address Space Layout Randomization (ASLR) based on a discussion on this bug report:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1087822

rgetz:
Ok - thanks for the pointer - so it's not a gcc issue - its a qemu/host kernel issue. That helps.

PR #1280, opened 2025-04-27
cmake/doc: Ensure users see the default install path

### LINKED ISSUE #1276: Default installs into /usr and not /usr/local
I've had to manually tweezer out and delete all the files it has installed into the system areas.  How rude of it!

Maybe this project should at least output the install location that is configured, so the user is not seeing it after `sudo`, when it has configured by default to use /usr:
```
cmake ..
...
NOTICE: -DCMAKE_INSTALL_PREFIX=/usr is set by default, if you expect /usr/local please configure cmake
```

The alternative of not being a good citizen is to have to test build and test make install into a docker container and examine the snapshot before/after to confirm it only touches

rgetz:
Resolves #1276

document make uninstall in the README_BUILD.md

during configure - print out the install path, and a hint on how to change it if people want.


README_BUILD.md
```diff
@@ -134,6 +134,15 @@ analog@precision:~/libiio/build$ make -j$(nproc)
 analog@precision:~/libiio/build$ sudo make install
 ```
 
+Note: As specified above, the default installation path on Linux based systems is '\usr'.
```
dNechita, line 137: Looks good. Just a tiny remark: it should be '/usr' instead of '\usr'.

rgetz: fixed

dNechita (approved):

PR #1281, opened 2025-04-29
cmake: ensure cmake options are printed out

rgetz:
This adds a test to ensure that cmake options are printed out via the cmake toggle_iio_feature macro.

### DIFF
--- a/CI/azure/check_README_BUILD.sh
+++ b/CI/azure/check_README_BUILD.sh
@@ -29,6 +29,19 @@ do
 	fi
 done
 
+# check if any cmake options don't print out at the end of the cmake process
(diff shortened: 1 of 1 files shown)


dNechita (approved):
Looks good, thank you!

dNechita:
> Conduct pointed out some missing double quotes- i will fix soon.

All right.
I've seen that these kinds of issues have been ignored in the past, but it would be great if you could fix them.

rgetz:
> I've seen that these kinds of issues have been ignored in the past, but it would be great if you could fix them.

yeah - we should fix them/not let them go - I have seen them cause problems on bsd and bash/dash/shell differences before...

-Robin

rgetz:
all fixed now.

dNechita (approved):

PR #1282, opened 2025-04-30
local: Fix disabling of IIO buffers in local_enable_buffer

schwesinger:
Currently local_enable_buffer() can not be used to disable IIO buffers. First, instead of propagating the value of the enable parameter to the call to local_do_enable_buffer() a hard-coded value is used. And second, setting both the buffer length and watermark attributes is attempted not only in the enable case but also in the disable case for an already active buffer. This causes local_enable_buffer() to fail early.

To fix the issue, only set buffer attributes in the enable case and properly propagate the requested action to local_do_enable_buffer().

## PR Type

## PR Checklist

### DIFF
--- a/local.c
+++ b/local.c
@@ -331,7 +331,7 @@ static int local_enable_buffer(struct iio_buffer_pdata *pdata,
 	if ((pdata->dmabuf_supported | pdata->mmap_supported) != !nb_samples)
 		return -EINVAL;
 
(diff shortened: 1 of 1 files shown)


dNechita (approved):
The fix makes sense.
Thanks for finding and fixing it!

nunojsa (approved):

Issue #1256, opened 2025-03-09
build error -DCPP_BINDINGS=ON and -DWITH_EXTERNAL_BACKEND=ON


rgetz:
This builds fine:
```
cmake ../ -DBUILD_SHARED_LIBS=ON -DCOMPILE_WARNING_AS_ERROR=ON -DCPP_BINDINGS=OFF -DCSHARP_BINDINGS=OFF -DENABLE_IPV6=OFF -DHAVE_DNS_SD=OFF -DLIBIIO_COMPAT=OFF -DNO_THREADS=OFF -DOSX_FRAMEWORK=OFF -DPYTHON_BINDINGS=OFF -DWITH_AIO=OFF -DWITH_EXAMPLES=OFF -DWITH_EXTERNAL_BACKEND=ON -DWITH_GCOV=OFF -DWITH_HWMON=OFF -DWITH_IIOD=OFF -DWITH_LIBTINYIIOD=OFF -DWITH_LOCAL_BACKEND=OFF -DWITH_LOCAL_CONFIG=OFF -DWITH_LOCAL_DMABUF_API=OFF -DWITH_LOCAL_MMAP_API=OFF -DWITH_MAN=OFF -DWITH_MODULES=OFF -DWITH_NETWORK_BACKEND=OFF -DWITH_NETWORK_BACKEND_DYNAMIC=OFF -DWITH_SERIAL_BACKEND=OFF -DWITH_SERIAL_BACKEND_DYNAMIC=OFF -DWITH_USB_BACKEND=OFF -DWITH_USB_BACKEND_DYNAMIC=OFF -DWITH_UTILS=OFF -DWITH_XML_BACKEND=OFF -DWITH_ZSTD=OFF
-- cmake version: 3.18.4
-- The C compiler identification is GNU 10.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test HAS_WPEDANTIC
-- Performing Test HAS_WPEDANTIC - Success
-- Performing Test HAS_WSHADOW
-- Performing Test HAS_WSHADOW - Success
-- Looking for strdup
-- Looking for strdup - found
-- Looking for strndup
-- Looking for strndup - found
-- Looking for strerror_r
-- Looking for strerror_r - found
-- Looking for strtok_r
-- Looking for strtok_r - found
[...]

PR #1283, opened 2025-05-05
Fix build when with external backend

dNechita:
## PR Description

This PR addresses issue: https://github.com/analogdevicesinc/libiio/issues/1256

## PR Type

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,9 +54,9 @@ if (CMAKE_C_COMPILER_ID MATCHES "GNU|Clang" AND NOT ${CMAKE_SYSTEM_NAME} STREQUA
 	# within Libiio itself first. This makes sure that the library won't
 	# call the symbols from the compat layer if the compat layer is used.
 
-	# TODO: Use cmake_link_options() in CMake 3.13+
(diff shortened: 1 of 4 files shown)


rgetz:
aren't you actually missing the weak definition - right now - it's not erroring - since the default is off, and you aren't turning it on in the tests... 

```
-- Features disabled: external 
```

dNechita:
> aren't you actually missing the weak definition - right now - it's not erroring - since the default is off, and you aren't turning it on in the tests...
> 
> ```
> -- Features disabled: external 
> ```

You are right, I am missing the weak definition. I am currently adding it and doing some tests.
What's interesting is that compilation doesn't fail. Locally, I am enabling the 'external' as well as examples and utils and build passes. But at runtime things crash badly when attempting to access struct members of iio_external_backend.

rgetz:
> What's interesting is that compilation doesn't fail. Locally, I am enabling the 'external' as well as examples and utils and build passes. But at runtime things crash badly when attempting to access struct members of iio_external_backend.

that's expected. You are telling the linker that it's external. So it assumes that - well - it's external. You can `dlopen/dlsym` it at run time to provide it. When you don't - it crashes.

-Robin

rgetz:
I think this can be closed now, in favour of #1288 ?

dNechita:
> I think this can be closed now, in favour of #1288 ?

Yes, it should be closed.

PR #1287, opened 2025-05-14
Relicense specific files under MIT license #1210

mhennerich:
## PR Description

As part of our ongoing efforts to enhance the flexibility and usability of libiio, we are relicensing (#1210) the following files from LGPL-2.1 to the MIT license:

 - attr.c, attr.h
 - backend.c, block.c, buffer.c, channel.c, context.c, device.c, events.c
 - iiod/ops.h, iiod/responder.c, iiod-responder.h
 - iio-private.h
 - include/iio/iio-backend.h, include/iio/iio-debug.h, include/iio/iio.h, include/iio/iio-lock.h
 - library.c, lock-dummy.c, mask.c, scan.c, sort.c, sort.h, stream.c, task.c
 - tinyiiod/tinyiiod.c, tinyiiod/tinyiiod.h
 - utilities.c, xml.c

This change aims to better support environments where static linking is required, such as microcontroller and baremetal/no-OS embedded systems, and to facilitate integration with Zephyr.

The MIT license provides greater permissiveness, allowing for broader use and integration of libiio.

All contributors to those files have provided their written consent for this relicensing.

### DIFF
--- a/COPYING_MIT.txt
+++ b/COPYING_MIT.txt
@@ -1,0 +1,53 @@
+MIT License
+
+Copyright (c) 2025 Analog Devices, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+Files covered by the MIT License:
+
+attr.c
+attr.h
+backend.c
+block.c
+buffer.c
+channel.c
+context.c
+device.c
+events.c
+iiod/ops.h
+iiod/responder.c
+iiod-responder.h
+iio-private.h
+include/iio/iio-backend.h
+include/iio/iio-debug.h
+include/iio/iio.h
+include/iio/iio-lock.h
+library.c
+lock-dummy.c
+mask.c
+scan.c
+sort.c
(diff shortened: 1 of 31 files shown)


dNechita:
Should an MIT license file be included as well, or would that be confusing since the MIT license only applies to certain files?

mhennerich:
> Should an MIT license file be included as well, or would that be confusing since the MIT license only applies to certain files?

Updated this PR with a MIT License file

dNechita (approved):
Looks good!

rgetz:
I know this is merged - but it would have been nice to do it with #1289 - so we could check things, and make sure it's all good (rather than checking by hand)...

rgetz:
I'm not sure I'm a fan of changing the main iio.h file to a MIT license without some other wording - it will give people the perception they are free to use the entire library under an MIT license anytime - which they are not.

For example - while some licenses have been moved to MIT, many files are not:
```
-- === License Check for : iio ==========
--     MIT: attr.c, backend.c, block.c, buffer.c, channel.c, context.c, device.c, events.c, library.c, mask.c, scan.c, sort.c, stream.c, task.c, utilities.c, xml.c, iiod-responder.h, iio-private.h, sort.h, attr.h
--     LGPL-2.1-or-later: usb.c, local.c, local-dmabuf.c, local-mmap.c, network.c, network-unix.c, dns_sd.c, dns_sd_avahi.c, lock.c, iiod-client.c, iiod-responder.c, dynamic.h, network.h, local.h, dns_sd.h
--     Missing SPDX: deps/libini/libini.c, deps/libini/ini.h
--     External libraries: /usr/lib/x86_64-linux-gnu/libusb-1.0.so, /usr/lib/x86_64-linux-gnu/librt.so, /usr/lib/x86_64-linux-gnu/libatomic.so.1, /usr/lib/x86_64-linux-gnu/libzstd.so, /usr/lib/x86_64-linux-gnu/libavahi-common.so, /usr/lib/x86_64-linux-gnu/libavahi-client.so, /usr/lib/x86_64-linux-gnu/libxml2.so, /usr/lib/x86_64-linux-gnu/libpthread.so
```

meaning that the "work" is under LGPL when compiled like I have set up - and the header file should say that - shouldn't it?

mhennerich:
Regarding the main iio.h file, I acknowledge your point about the potential misperception of the entire library being under the MIT license. It's quite common for the resulting license to be based on the licenses of each component.

I’m open to add appropriate wording to the header file to clarify that while some files are under the MIT license, others remain under the LGPL-2.1-or-later license. 

Something like this ?: 
“Please BE AWARE that the resulting license depends on how libiio is configured during build time. In almost all cases, the resulting license is LGPL-2.1-or-later !”

rgetz:
I think in this case, it's better to use more words.

>libiio includes source components that are released under different open-source licenses, including MIT, GPL, and LGPL. As a result, the license of the final binary is determined by how libiio is configured at build time. The specific build options selected can affect which licensing terms apply to the resulting binary. In most cases—particularly when building for Linux or other POSIX-compliant systems—the default configuration produces a libiio shared library licensed under the GNU Lesser General Public License version 2.1 or later (LGPL-2.1-or-later). To ensure that the final license aligns with your project's compliance and distribution requirements, we strongly recommend reviewing your build configuration. You can enable a license compatibility check during configuration by adding the option -DLICENCE_CHECK=ON.

This assumes the merge of #1289 

-Robin

rgetz:
If that makes more sense - I can do it as part of that. As long as we are committed to doing that. I know some don't like all my cmake complexity...

PR #1290, opened 2025-05-18
mingw64 - fix CI issues

rgetz:
update path / binary name


dNechita (approved):
This is really helpful. Thank you for fixing it!

Issue #1304, opened 2025-06-21
iio_info should not read debug registers


dlech:
We've recently got some pushback in the Linux kernel about how debug attributes that read on-chip registers are implemented.

On chips that have separate data capture and register read modes, for example, reading a debug register can take the chip out of data capture mode and mess up a buffered read. So if someone runs `iio_info` while a buffered read is enabled, it can break things.

I.e. `iio_info` prints something like this:

```console
		1 debug attributes found:
				debug attr  0: direct_reg_access value: 0xE6
```

It was suggested that we really [should not be doing that](https://lore.kernel.org/linux-iio/20250621170824.249c6b0c@jic23-huawei/). So let's remove reading the attribute and just print the attribute name.

And if there are any other tools that read a debug attribute without the user explicitly requesting it, we should fix that too.

dNechita:
Hi @dlech,

The `iio_info` tool provides the entire structure of an iio context (devices, triggers, channels, buffers and attributes) and at the same time it captures a snapshot of the state of that context by reading the attributes values. Forcing `iio_info` not to read the debug attributes would conflict with one of its purposes.
Skipping reading debug attributes might not be a complete fix as reading a non-debug attribute might trigger some chip register read (a register read call somewhere in the linux driver implementation).

I suggest adding a command line option that would instruct `iio_info` to skip reading debug attributes or all types of attributes. But the default behavior should remain the same as of today.

dlech:
> Skipping reading debug attributes might not be a complete fix as reading a non-debug attribute might trigger some chip register read (a register read call somewhere in the linux driver implementation).

This may be true, but "normal" attributes are much less likely to cause a problem compared since they have more protections in the IIO drivers. Whereas debug attributes tend to be intentionally left "dangerous" by design.

I would be OK with adding options to control this, but I would still make the default only read regular attributes and not debug attributes because of this fundamental difference in the design choice of the attribute types.

dlech:
Also, the debug attribute read currently only works is there is a register at 0x00, so it isn't really fully enumerating all possible registers so I don't think it is actually particularly useful to print that value. One would have to look at the datasheet to even know what the register is compared to other attributes that have names that are self-describing.

dlech:
Again, this isn't about sysfs attribute. We expect kernel drivers to make sure they are always safe to read and write at any time. I'm not suggesting to change anything in that regard.

[...]

PR #1305, opened 2025-06-21
utils: iio_info: do not print debug attribute values

dlech:
## PR Description

Do not print the values of debug attributes in iio_info.

It has come up multiple times that reading debug attributes can cause unintended side effects. For example if the IIO driver is currently doing a buffered read, reading the debug attribute could cause the chip to exit data capture mode and break the buffered read. Also, some registers could clear or reset something, like an interrupt, when a status register is read.

So general-purpose tools like iio_info should not read debug attributes that the user did not explicitly request.

Closes: https://github.com/analogdevicesinc/libiio/issues/1304
Link: https://lore.kernel.org/linux-iio/20250621170824.249c6b0c@jic23-huawei/

## PR Type

## PR Checklist

### DIFF
--- a/utils/iio_info.c
+++ b/utils/iio_info.c
@@ -9,6 +9,7 @@
 #include <errno.h>
 #include <getopt.h>
 #include <iio/iio.h>
+#include <iio/iio-backend.h>
 #include <iio/iio-debug.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -33,12 +34,16 @@
 static struct iio_context *ctx;
 
 static const struct option options[] = {
-	{0, 0, 0, 0},
+	{ "read-debug-attr", no_argument, NULL, 'd' },
+	{ "no-read-attr", no_argument, NULL, 'n' },
+	{ }
 };
 
 static const char *options_descriptions[] = {
 	("[-x <xml_file>]\n"
 		"\t\t\t\t[-u <uri>]"),
+	("Read and print the value of debug attributes."),
+	("Do not read or print the value of regular attributes."),
 };
 
 static int dev_is_buffer_capable(const struct iio_device *dev)
@@ -56,7 +61,7 @@ static int dev_is_buffer_capable(const struct iio_device *dev)
 	return false;
 }
 
-#define MY_OPTS ""
+#define MY_OPTS "dn"
 
 static bool colors;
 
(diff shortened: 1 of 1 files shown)


dlech:
Umm... I think the static checker needs work. This is clearly not true.

image

gastmaier (approved):

gastmaier:
Note: merging to main is merging to >= v1.
@dNechita if this is approved by you, should a similar pr be created for v0.x?

rgetz:
I would rather add a '-d' option, which enables/suppresses reading them - so people have a way to do what they have been doing for 10+ years.

nunojsa:
> I would rather add a '-d' option, which enables/suppresses reading them - so people have a way to do what they have been doing for 10+ years.

I tend to agree with the above. Arguably this should never have been added but it was and users have been using it. Even wonder what the default should be? Meaning, should we keep behavior and add a switch to shut it down? But given that this can lead to unexpected bugs, I would go for an explicit "enable reading these attributes".

dlech:
I updated this to add to new command line options.

I still think it is better to have reading debugfs attributes disabled by default to avoid the issues it can cause, so I've implemented the options this way.

There is an option to enable reading debugfs attributes just in case anyone was relying on that (although since it typically only reads register 0x00 if one even exists, I don't expect that it was actually very useful).

And there is a separate option to disable reading sysfs attributes just in case anyone wants to avoid poking hardware at all, but the default here is read and print the sysfs attribute values just like we have always done.

In all cases, the attribute names are still printed, it is just the values that may or may not be printed.

dNechita:
> Umm... I think the static checker needs work. The code compiles, so there clearly isn't a problem with the code itself.

You can ignore that finding.

utils/iio_info.c
```diff
@@ -33,12 +34,16 @@
 static struct iio_context *ctx;
 
 static const struct option options[] = {
-	{0, 0, 0, 0},
+	{ "read-debugfs", no_argument, NULL, 'd' },
```
dNechita: When `iio_info` generates its output, it categorizes the attributes into context attributes, device attributes, buffer attributes, channel attributes, and debug attributes. Therefore, it makes sense to use the same naming. I suggest:
`{ "read-debug-attr", no_argument, NULL, 'd' }`

utils/iio_info.c
```diff
@@ -33,12 +34,16 @@
...
+	{ "no-read-attr", no_argument, NULL, 'n' },
+	{ }
 };
 
 static const char *options_descriptions[] = {
 	("[-x <xml_file>]\n"
 		"\t\t\t\t[-u <uri>]"),
+	("Read and print the value of debugfs attributes."),
+	("Do not read sysfs attributes, only print the names."),
```
dNechita: I would rephrase these in order to:
 - use  'debug' instead of 'debugfs'
 - create more consistency between the 2 descriptions

I suggest:

`("Read and print the value of debug attributes."),`
`("Do not read the value of regular attributes."),`

dNechita:
When running with option `-n`, iio_info also hides the context attributes that could contain useful information such as the kernel version, the URI used to create the context, hardware version, hardware firmware version, etc. Additionally, these are not sysfs attributes. Instead, libiio reads them from an .ini file.
Therefore, it makes sense to always print their values. They can be easily detected by checking if they are of type:`IIO_ATTR_TYPE_CONTEXT`.

dlech:
Thanks for the suggestions, I have updated the changes accordingly.

PR #1309, opened 2025-06-24
Add helpers to extract range and items from _available attributes

### LINKED ISSUE #1235: add functions to decode _available attributes
Many IIO channel attributes have an _available modifier which is [described ](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio)as:

```
/sys/bus/iio/devices/iio:deviceX/sampling_frequency_available

                - a small discrete set of values like "0 2 4 6 8"
		- a range with minimum, step and maximum frequencies like
		  "[min step max]"

/sys/bus/iio/devices/iio:deviceX/in_activity_calibgender_available
		Lists all available gender values (e.g.: male, female).

/sys/bus/iio/devices/iio:deviceX/in_intensity_hardwaregain_available
		Lists all available hardware applied gain factors. Shared across all
		channels.

/sys/.../iio:deviceX/out_voltageY_powerdown_mode_available
		Lists all available output power down modes.
		If Y is not present the mode is shared across all outputs.

/sys/bus/iio/devices/iio:deviceX/calibration_forced_value_available
		Available range for the forced calibration value, expressed as:
		- a range specified as "[min step max]"

/sys/.../events/in_accel_gesture_tap_value_available
		Lists all available threshold values which can be used to
		modify the sensitivity of the tap detection.
```

I would suggest we need a few fuctions for this:

* `ssize_t iio_channel_get_min(const struct iio_channel *chn)`   (returns SSIZE_MIN if doesn't exist)
* `ssize_t iio_channel_get_step(const struct iio_channel *chn)`   (returns 0 if doesn't exist)
* `ssize_t iio_channel_get_max(const struct iio_channel *chn)`   (returns SSIZE_MAX if doesn't exist)
* `ssize_t iio_channel_get_available(const struct iio_channel *chn, char * buf, size_t len)` (returns zero if get_min/get_max should be used, or if doesn't exist, and positive number of elements with all elements in buf.).

It would make things much easier for people, and get rid of all the string parsing in applications for users.

dNechita:
## PR Description
For those attributes ending in '_available' provide helpers to:
 - extract range values: minimum value, step size and maximum value
 - split the string containing multiple word into an array of strings

Closes #1235.

## PR Type

## PR Checklist


utilities.c
```diff
@@ -435,3 +435,76 @@ uint64_t iio_read_counter_us(void)
...
+	if (!input)
+		return -EINVAL;
+
+	// Expect format: [min step max]
+	#if defined(_MSC_VER)
+    n = iio_sscanf(input, " [ %lf %lf %lf %c", &lmin, &lstep, &lmax, &extra, (unsigned int)sizeof(extra));
+#else
+    n = iio_sscanf(input, " [ %lf %lf %lf %c", &lmin, &lstep, &lmax, &extra);
+#endif
+
```
rgetz: this looks like white space issues? The #if should be alighned with the #else/#endif?

include/iio/iio.h
```diff
@@ -445,6 +445,27 @@ iio_attr_get_filename(const struct iio_attr *attr);
...
+ * @param input A string that is expected to contain multiple words separated by spaces
+ * or newlines
+ * @param count Will be set with the number of words found if the string was successfully
+ * parsed. On fail, it will be set to 0
+ * @return On success, a dynamically allocated array of strings. It is the caller’s
+ * responsibility to free these newly allocated resources
+ * @return On error, NULL is returned */
+__api __pure
+char **iio_get_list_from_available(const char *input, size_t *count);
+
```
rgetz, line 506: I don't understand how his can be used in an application?

include/iio/iio.h
```diff
@@ -445,6 +445,27 @@ iio_attr_get_filename(const struct iio_attr *attr);
...
+/** @brief Extract the range from a string originating from an '_available" attribute
+ * @param input A string that is expected to have the [min step max] format
+ * @param min Will be set with the range minimum value if string was successfully parsed
+ * @param step Will be set with the range step value if string was successfully parsed
+ * @param max Will be set with the range maximum value if string was successfully parsed
+ * @return On success, 0 is returned
+ * @return On error, a negative errno code is returned */
+__api __pure
+int iio_get_range_from_available(const char *input, double *min, double *step, double *max);
+
```
rgetz, line 459: I'm not sure why the application needs to read the channel _attribute_available attribute, and then pass it here, rather than just passing the channel_attribute to this in the first place? (like the way that the rest of libiio works?)

dNechita: Not all attributes have the "_available" which made the think of avoiding misusage. One would need to actually read the attribute and then pass the data further to this helper. But I haven't thought about inconsistency with the rest of the API). Also, this approach causes the buffer storing the attribute value to be create twice: once when attribute gets read and once inside the helper.
It makes sense to change the signature so that it takes an `iio_attr` as parameter.

dNechita:
I gave it another go at this:
V1:

- the 2 functions that get the range and the list of available items, now take a `struct iio_attr *`
- the function that gets the list of available items will allocate space for each item as well as for the array containing all those items. This looks a bit complex. I added a helper that can be used to easily free the array of arrays. Would it make sense to just expect the caller to provide an already allocated space where the raw value of the attribute is going to be copied? Just like `iio_attr_read_raw()` does?

@rgetz @nunojsa

rgetz:
This is still missing hiding `_available` in `local.c` like we do with `_label` ...

and then using that in `iio_info` and `iio_attr` - to print and check bounds before writing things...

I think the more we see how it will be used in applications - the easier it will be to provide feedback on your other questions...

rgetz:
I like keeping it simple, and aligned with other widely used C libraries and POSIX APIs that follow the pattern of:
 - The library allocates space for variable-length lists or strings.
- The user is responsible for freeing them, often with a dedicated helper.
  - POSIX: getaddrinfo() and freeaddrinfo()
  - strdup() and free()
  - glib: g_strsplit(), g_strfreev()
  - readdir()/opendir()/closedir() 
  - getline() and free()
  - etc

These examples show that allocating variable-length lists or structures within the library and providing a specific free function (or requiring free()) is a standard, user-friendly pattern in C APIs, especially when the size/count is not known in advance. However, in tight memory constrainted systems, offering a variant that allows the caller to supply the buffer (or, at least, to precompute the needed size) on the stack rather than the heap may be preferred. To do that, we could offer an additional “raw” version where the caller supplies a buffer.

```
/**
 * iio_attr_get_available_buf
 * @attr:      The attribute to query.
 * @buf:       Buffer provided by the caller, will be filled with raw attribute value.
 * @buflen:    Size of the buffer.
 * @list:      (Optional) Array of char* pointers to set to each item.
 * @count:     (Optional) On input: number of slots in list[]. On output: number of items found.
 *
 * Returns: 0 on success, -errno on failure, or -ENOSPC if not enough space in list[].
 */
int iio_attr_get_available_buf(const struct iio_attr *attr,
                               char *buf, size_t buflen,
                               char **list, size_t *count)
{
    ssize_t ret;
    size_t n = 0, max = (size_t)-1;
    char *p, *delim;
    size_t remaining;

    if (!attr || !buf || buflen == 0)
        return -EINVAL;

    if (list && count)
        max = *count;                  // Limit to user-supplied count
    else if (list && !count)
        max = buflen / 2 + 1;          // Heuristic: max tokens if all single-char with spaces
    // else if !list: just count, so max = SIZE_MAX

    ret = iio_attr_read_raw(attr, buf, buflen);
    if (ret < 0)
        return (int)ret;
    if (buf[0] == '[') // skip if it's a range
        return -ENODATA;

    // Guarantee: buffer is null-terminated within buflen

    p = buf;
    remaining = buflen - (size_t)(p - buf);

    while (n < max && remaining && *p) {
        if (list)
            list[n] = p;
        n++;

        // Find next space, but do not exceed buflen
        delim = memchr(p, ' ', remaining);
        if (delim) {
            *delim = '\0';
            p = delim + 1;
            remaining = buflen - (size_t)(p - buf);
        } else {
            break;
        }
    }

    // If the list[] is not large enough
    if (list && count && *p && n == max)
        return -ENOSPC;

    if (count)
        *count = n;

    return 0;
}
```

rgetz:
with the above, we should be able to do things in Qt/GTK like;

```
void populateComboBox(QComboBox *combo, const struct iio_attr *attr)
{
    char buf[1024];
    char *list[32];
    size_t count = 32;

    int ret = iio_attr_get_available_buf(attr, buf, sizeof(buf), list, &count);
    if (ret != 0)
        return; // handle error as needed

    combo->clear();
    for (size_t i = 0; i < count; ++i) {
        combo->addItem(QString::fromUtf8(list[i]));
    }
}
```

or

```
void populate_combo_box_text(GtkComboBoxText *combo, const struct iio_attr *attr)
{
    char buf[256];
    char *list[32];
    size_t count = 32;

    int ret = iio_attr_get_available_buf(attr, buf, sizeof(buf), list, &count);
    if (ret != 0)
        return; // handle error as needed

    gtk_combo_box_text_remove_all(combo);
    for (size_t i = 0; i < count; ++i) {
        gtk_combo_box_text_append_text(combo, list[i]);
    }
}
```

which I think keeps things pretty clean...

utils/iio_info.c
```diff
@@ -117,19 +120,36 @@ static void print_attr(const struct iio_attr *attr,
...
+	if (value) {
 		printf(" value: %s\n", value);
-	else if (colors)
-		print_fmt(" value: " FMT_ERR "\n", value);
-	else
-		printf(" value: ERROR: %s\n", value);
+	} else {
+		ret = iio_attr_get_available(attr, &available_list, &available_count);
+		if (ret == 0) {
+			printf(" value:");
```
rgetz: Instead of "value:", a better label might be "options:". This clarifies to the user that the attribute has multiple possible values, not just a single one.

rgetz:
> This is still missing hiding `_available` in `local.c` like we do with `_label` ...

this.

> and then using that in `iio_info` and `iio_attr` - to print and check bounds before writing things...

and `iio_attr`...

- Enhanced Attribute Display
Provides a list or a range, and provide that after it prints out the value. Display these options/ranges in a human-readable way when listing attributes, instead of just printing raw strings.
If the attribute describes a range (e.g., scale_available contains: min=0.5 step=0.1 max=2.0):
Print as:
attr  0: sampling_frequency
 value: 2000000
 Available options: 1000000, 2000000, 4000000
attr  1: scale
 value: 1.0
 Available range: min=0.5, step=0.1, max=2.0


- Validation of Write Operations
When the user tries to set a value for an attribute, iio_attr can:
Use the _available functions to check if the requested value is valid (i.e., part of the available list or within the allowed range).
Reject invalid values with a helpful human understandable error message, even before actually writing the hardware.

dNechita:
We won't be hiding the "_available" attributes because it will require too many changes in the clients code. These newly added functions work great as helpers for the attributes of type "_available".
I have updated the iio_attr which now can provide an enhanced display.
I won't add the validation in this PR. It's a nice enhancement but could be added at a later time.

rgetz:
@dNechita 
>We won't be hiding the "_available" attributes because it will require too many changes in the clients code. 

This defeats the entire purpose, and then you should revert this.

We did the same for _label (it was hidden in libiio much after it was added to the kernel). This is the same.

few people are using 1.0 in their applications - so the impact is small now. Do this now, rather than later...

>I won't add the validation in this PR. It's a nice enhancement but could be added at a later time.

if you don't do it now - you will never do it. That's just the way things work. If you don't want to - just say that - but later means never.

rgetz:
@dNechita  - it looks like you broke (removed useful information) with this commit.

tip of main:
```
./utils/iio_attr -v -u usb:1.73.5 -c ad9361-phy TX_LO frequency
ERROR: iio_attr: Could not find channel (TX_LO)
```
How are you supposted to know what anything is like this?
```
./utils/iio_attr  -u usb:1.73.5 -c ad9361-phy altvoltage1 
value '0'
value '0'

ERROR: iio_attr: Unable to read attribute: Invalid argument (22)
value '0 245,245,245,245,245,245,245,245,245,245,245,245,245,245,245,245'
value '0'
value '2500000000'
range '[min: 3.25e+08 step: 1 max: 3.8e+09]'
value '0'
```
And verbose is gone?
```
getz@brain:~/github/libiio-mainline/build$ ./utils/iio_attr  -u usb:1.73.5 -c ad9361-phy altvoltage1 frequency
2500000000
rgetz@brain:~/github/libiio-mainline/build$ ./utils/iio_attr -v  -u usb:1.73.5 -c ad9361-phy altvoltage1 frequency
value '2500000000'
```


Previous versions:
```
./tests/iio_attr -u usb:1.73.5 -c ad9361-phy TX_LO
dev 'ad9361-phy', channel 'altvoltage1' (output), id 'TX_LO', attr 'external', value '0'
dev 'ad9361-phy', channel 'altvoltage1' (output), id 'TX_LO', attr 'fastlock_load', value '0'
dev 'ad9361-phy', channel 'altvoltage1' (output), id 'TX_LO', attr 'fastlock_recall', ERROR: Invalid argument (22)
dev 'ad9361-phy', channel 'altvoltage1' (output), id 'TX_LO', attr 'fastlock_save', value '0 245,245,245,245,245,245,245,245,245,245,245,245,245,245,245,245'
dev 'ad9361-phy', channel 'altvoltage1' (output), id 'TX_LO', attr 'fastlock_store', value '0'
dev 'ad9361-phy', channel 'altvoltage1' (output), id 'TX_LO', attr 'frequency', value '2500000000'
dev 'ad9361-phy', channel 'altvoltage1' (output), id 'TX_LO', attr 'frequency_available', value '[325000000 1 3800000000]'
dev 'ad9361-phy', channel 'altvoltage1' (output), id 'TX_LO', attr 'powerdown', value '0'
```
and verbose worked:
```
./tests/iio_attr -u usb:1.73.5 -c ad9361-phy TX_LO frequency
2500000000
./tests/iio_attr -v -u usb:1.73.5 -c ad9361-phy TX_LO frequency
dev 'ad9361-phy', channel 'altvoltage1' (output), id 'TX_LO', attr 'frequency', value '2500000000'
```

PR #1319, opened 2025-07-04
attr: Fix context attributes getting out of sync with their values

dNechita:
## PR Description

Each time a context attribute is added, its value gets appended to a list of values while its name gets appended to a list of names.  But since the names list gets sorted and the values list doesn't, the two lists can become misaligned.

To fix this, after sorting, find where the newly added attribute was inserted. Because the list gets sorted each time a new attribute is added we know attributes positioned after the newly inserted attributes just get shifted towards the end of the list. So we do the same for the values.

## PR Type

## PR Checklist

### DIFF
--- a/attr.c
+++ b/attr.c
@@ -291,6 +291,25 @@ int iio_context_add_attr(struct iio_context *ctx,
 		return ret;
 	}
 
+	/* Keep attr values in sync with the names which just got sorted in iio_add_attr() */
+	unsigned int j, new_idx = ctx->attrlist.num - 1;
+
+	/* Find the new index of the added attr */
+	for (j = 0; j < ctx->attrlist.num - 1; j++) {
+		if (!strcmp(ctx->attrlist.attrs[j].name, key)) {
+			new_idx = j;
+			break;
+		}
+	}
+
+	/* If the new position is not at the end, it was inserted at a position
+	causing subsequent attributes to shift forward as they were already sorted */
+	if (new_idx != ctx->attrlist.num - 1) {
+		memmove(&ctx->values[new_idx + 1], &ctx->values[new_idx],
+			(ctx->attrlist.num - new_idx - 1) * sizeof(*ctx->values));
+		ctx->values[new_idx] = new_val;
+	}
+
 	return 0;
 }
 


attr.c
```diff
@@ -291,6 +291,25 @@ int iio_context_add_attr(struct iio_context *ctx,
 		return ret;
 	}
 
+	/* Keep attr values in sync with the names which just got sorted in iio_add_attr() */
+	unsigned int j, new_idx = ctx->attrlist.num - 1;
```
nunojsa, line 295: Couple of things. First I would keep the style mostly used in libiio. We might have instances where mixed code declarations is being used but never really saw it in core code. Arguably you might not agree but it important to keep coding style consistent.

Second point is regarding defaulting `new_idx` to `ctx->attrlist.num - 1`. Fine to do it but then I would say there's no point for the loop to go til `ctx->attrlist.num`?

dNechita: Alright, I will place the declaration at the beginning of the function along the others. 

> I would say there's no point for the loop to go til ctx->attrlist.num?

The idea here was that if the new attribute isn't found, `new_idx` would point to the last position, so the shifting the values won't occur (since the code wouldn’t enter the if statement). However, I agree the logic can be a bit confusing, and in practice, it’s almost impossible for the attribute not to be found, since it was just added.

rgetz:
>But since the names list gets sorted and the values list doesn't, the two lists can become misaligned.

I guess I don't understand why we just don't refactor things:
 - sort both at the same time
 - read/create the value list after the sort

wouldn't that be cleaner?

attr.c
nunojsa: > The idea here was that if the new attribute isn't found, new_idx would point to the last position, so the

Yes, but as the code stands you can still go over the loop. and set `new_idx = ctx->attrlist.num - 1` in the last iteration (which seems useless). My point is that the loop should likely go only until `ctx->attrlist.num - 1`

attr.c
```diff
@@ -215,7 +227,30 @@ int iio_add_attr(union iio_pointer p, struct iio_attr_list *attrs,
 
 	attrs->num++;
 
-	iio_sort_attrs(attrs);
+	if (lvalues) {
+		attrs_and_vals.attr_val_pairs = malloc(attrs->num * sizeof(struct iio_attr_val));
```
rgetz: why not just pass `lvalues` to `iio_sort_attrs`  and then if `lvalues` isn't null, combine/separate them there - it will keep everything a little cleaner (I think).

dNechita: It might help a bit, but I would go back to the first implementation where I was inserting the value of the attribute.
I kept the first commit (that's why I reverted it so it can still be seen here in this PR) to compare its complexity with this approach. The 1st is simpler and more efficient. In the 2nd I am doing quite a few of `mallocs` and `memcpys`.
The thing is that these context attributes are the exception as they are the only ones for which we need to keep the list of their corresponding values. For all the other attributes (device, channel, buffer, debug), the function `iio_add_attr()` fits perfectly.

rgetz: maybe the "right" answer is not to use qsort, and just a simple sort algorithm that you do yourself, that can manage both structures... (to keep them sync'ed) That might be the simple answer...

rgetz: plus - I think we need this when iio_emu shows up (doesn't it?) - so this would be useful.

dNechita: I am keeping the initial implementation for now as I can't anticipate the exact form in which the attributes will be kept with iio_emu. But I would use your idea to have a way to sort both attributes and values when the time comes.

PR #1321, opened 2025-07-15
fix: Use bytes_used instead of buffer_size for benchmark throughput c…

xR3b0rn:
It's entirely possible that the underlying driver does not completely fill the buffer, especially when using `dma_buf`. Therefore, throughput calculations should not assume the buffer is full and should instead use `bytes_used` to determine the actual throughput.


xR3b0rn:
I haven't looked too deeply into what `libiio` does or doesn't do. However, in my case, the driver always uses the same number of bytes, regardless of the size of the DMA buffer it receives from the userspace application. Since the driver sets the correct value for `bytes_used`, I assumed that `libiio` receives this value through the attribute of the same name.

nunojsa:
While for the low speed file io implementation, that value does matter and you only get (on your read() write()) the number of bytes sent/received. For the high speed mode does not work like that and you would need some custom way of doing that...

nunojsa:
> Is this a restriction of dma_buf or of libiio?

AFAIU, the kernel does not give this information back to you when your block is mmaped or you have dmabufs.

PR #1322, opened 2025-07-15
Small fixes improvements

dNechita:
## PR Description

- Found that iiod was trying to handle an error but using a wrong variable.

- Log a message even for exceptional use cases (having a really large sample size) instead of silently returning earlier from a functton.

## PR Type

## PR Checklist

### DIFF
--- a/iiod/iiod.c
+++ b/iiod/iiod.c
@@ -106,6 +106,8 @@ static void *get_xml_zstd_data(const struct iio_context *ctx, size_t *out_len)
 	void *buf;
 #if WITH_ZSTD
 	size_t ret;
+	char *bytes;
+	size_t i;
 
 	len = ZSTD_compressBound(xml_len);
 	buf = malloc(len);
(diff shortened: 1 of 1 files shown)


nunojsa (approved):
LG.

nit: git subject on the second patch: 's/channel.c/channel/'

iiod/iiod.c
```diff
@@ -119,7 +119,7 @@ static void *get_xml_zstd_data(const struct iio_context *ctx, size_t *out_len)
 
 	if (ZSTD_isError(ret)) {
 		IIO_WARNING("Unable to compress XML string: %s\n",
-			    ZSTD_getErrorName(xml_len));
+			    ZSTD_getErrorName(ret));
```
rgetz, line 124: while you are here - do you actually want to see `buf` to know what didn't compress?

dNechita: I could print it out. At the moment, I am not sure what I could do with this kind of information but it's better than not knowing.

PR #1335, opened 2025-08-12
network: add flag to convert SIGPIPE to EPIPE

### LINKED ISSUE #1333: Missing MSG_NOSIGNAL flag on network_write_data for libiio v0
Hello, I noticed that when using an iio-emu device if the device is no longer available and we try to run any functions we get the following error
```
Unable to create IIO context ip:127.0.0.1: connection refused (111)
Segmentation fault (core dumped) 
```

When a client app is using libiio and receives "Segmentation fault" it will crash.
This issue seems to be on both linux and windows systems.

This issue does not appear for when using a usb or ip connected device, for those we just get 
```
Unable to create IIO context ip:127.0.0.1: connection refused (111)
```

The main difference is that using the emu and having that error would cause a crash 
From my research the emu connection is retuning SIGPIPE  and the device connection is retuning EPIPE  

This can be fixed by adding the MSG_NOSIGNAL to the return of [this function ](https://github.com/analogdevicesinc/libiio/blob/ba74e6c53007925ff714ed28a258b5aebf6a3ad4/network.c#L954)

resulting in the new return being: 

`return network_send(io_ctx, src, len, MSG_NOSIGNAL);`

Adding MSG_NOSIGNAL  will convert SIGPIPE  to EPIPE. 

I encountered this issue using libiiov0 but I noticed the same fix could be applied for libiiov1. 

To be able to reproduce the issue mentioned do the following 
1) start an iio-emu 
2) do a read for that emu (for example /iio_reg -u "ip:127.0.0.1" IIO_DEVICE 0x0 )
3) stop the emu or kill the emu process 
4) do step 2) again

IonutMuthi:
This pr would close #1333 . 
This issue was initially found on libiio v0.26 but it persists to this version.
Adding this flag to the network would convert SIGPIPIE to EPIPE preventing app crashes on client side.

### DIFF
--- a/network.c
+++ b/network.c
@@ -21,6 +21,7 @@
 #include <string.h>
 #include <sys/types.h>
 
+
 #ifdef _WIN32
 #include <netioapi.h>
 #include <winsock2.h>
@@ -49,6 +50,10 @@
 
 #ifdef _WIN32
 #define close(s) closesocket(s)
+
+#ifndef MSG_NOSIGNAL
+#define MSG_NOSIGNAL 0
+#endif
 #endif
 
 #define NETWORK_TIMEOUT_MS 5000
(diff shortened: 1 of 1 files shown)


network.c
```diff
@@ -21,6 +21,7 @@
 #include <string.h>
 #include <sys/types.h>
 
+
```
nunojsa, line 24: unrelated change

dNechita (approved):

Issue #1329, opened 2025-08-03
Fix file chosing for CMA


xR3b0rn:
This issue can be seen as a follow-up to [libiio issue #1259](https://github.com/analogdevicesinc/libiio/issues/1259).

According to this [kernel issue](https://bugzilla.kernel.org/show_bug.cgi?id=220382), `linux,cma` is **not** a universally defined name - that is, it's not guaranteed to point to the CMA memory on every system. The actual name of the device file depends on the name of the CMA node in the device tree or may be `reserved` if `cma=256M` is set on the  linux cmdline.

So, technically speaking, the hardcoded path `/dev/dma_heap/linux,cma` may be correct on many systems, but it is not guaranteed to work everywhere. If `cma=<size>` is specified or a differently named device tree node is used, the corresponding file name will change. As such, the hardcoded path referenced [here](https://github.com/analogdevicesinc/libiio/blob/main/local-dmabuf.c#L92) may not actually refer to the CMA memory on all systems.

A potential solution would be to introduce a configurable option, e.g., `dma_dev_file`, with a default value of `/dev/dma_heap/system`. If the user intends to use CMA memory, they would need to explicitly set `dma_dev_file` to `/dev/dma_heap/linux,cma` (or whatever path corresponds to the CMA memory on their particular system).

PR #1339, opened 2025-08-18
[draft] feat(local): Make DMA allocator freely configurable

xR3b0rn:
Fixing #1329

### DIFF
--- a/iiod-client.c
+++ b/iiod-client.c
@@ -1461,8 +1461,7 @@ iiod_client_create_buffer(struct iiod_client *client,
 			err = -ENOMEM;
 			goto err_free_pdata;
 		}
-
-		buf_params->dma_allocator = htonl(params->dma_allocator);
+		buf_params->dma_allocator = params->dma_allocator;
 		buf_params->nb_mask = (unsigned int) mask->words;
 		for (unsigned int i = 0; i < mask->words; i++)
 			buf_params->mask[i] = htonl(mask->mask[i]);
--- a/include/iio/iio.h
+++ b/include/iio/iio.h
@@ -157,18 +157,6 @@ struct iio_context_params {
 };
 
 /**
- * @enum iio_buffer_dma_allocator
- * @brief Provides low-level control over DMA buffer allocation.
- */
-enum iio_buffer_dma_allocator {
-	/** @brief May use scattered DMA memory (default). */
-	IIO_DMA_ALLOCATOR_SYSTEM = 0,
-
-	/** @brief Enforces contiguous DMA memory (for expert use only). */
-	IIO_DMA_ALLOCATOR_CMA_LINUX = 1,
-};
-
-/**
  * @struct iio_buffer_params
  * @brief Used to pass parameters to the iio_device_create_buffer function
  */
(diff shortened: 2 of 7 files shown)


include/iio/iio.h
```diff
@@ -178,10 +166,11 @@ struct iio_buffer_params {
 	unsigned int idx;
 
 	/** @brief Use the allocator to allocate a DMA buffer.
-	 * In most cases, this should be set to IIO_DMA_ALLOCATOR_SYSTEM.
+	 * In most cases, this should be set to NULL, what defaults to system.
 	 * ONLY change this to a different option if you fully understand the
-	 * implications. (default: IIO_DMA_ALLOCATOR_SYSTEM) */
-	enum iio_buffer_dma_allocator dma_allocator;
+	 * implications. (default: system) */
+	const char *dma_allocator;
```
nunojsa, line 172: If libiio v1 was already out the door, this would be ABI breakage. There has been some discussions around all of this implementation about the allocator and this seems to be one more example of this being very fragile while adding fairly unnecessary complexity (at this point at least).

Ultimately, this is up to @dNechita, but if it was to me, I would be very tempted to drop this whole implementation (I know is not that nice for you) and just add an IIOD CMake build option with the allocator. I know I was the one pushing for more flexibility in the beginning but looking at this I feel is better to have something very simple and minimal and add/implement stuff when needed. 

The above is based on the assumption that for a given system we'll be using the same allocator (which I believe is fairly reasonable).

schwesinger (commented):
I recently ran into #1329, so thanks for working on this.

I had a look at the core changes and tested them without issues.

Maybe the commit could be split into core API changes and tool changes to make reviews (especially partial reviews) a bit easier.

So for `include/iio/iio.h` and `local-dmabuf.c`:

Reviewed-by: Folker Schwesinger <dev@folker-schwesinger.de>
Tested-by: Folker Schwesinger <dev@folker-schwesinger.de>

include/iio/iio.h
schwesinger: > Ultimately, this is up to @dNechita, but if it was to me, I would be very tempted to drop this whole implementation (I know is not that nice for you) and just add an IIOD CMake build option with the allocator.

I ran into this in a non-IIOD context, so the build option would be needed for the core library as well. Though, using a build option would be fine in my use-case.

nunojsa: > I ran into this in a non-IIOD context, so the build option would be needed for the core library as well. Though, using a build option would be fine in my use-case.

Hmm likely right! This only has effect on the local backend that's why I tend to link it with iiod

iiod-client.c
```diff
@@ -1461,8 +1461,7 @@ iiod_client_create_buffer(struct iiod_client *client,
 			err = -ENOMEM;
 			goto err_free_pdata;
 		}
-
-		buf_params->dma_allocator = htonl(params->dma_allocator);
+		buf_params->dma_allocator = params->dma_allocator;
```
nunojsa, line 1464: Also this does not look right... Have you tried it? This is just a pointer so the actual string size is not being taken into account which means `buf_params` won't be fully copied

include/iio/iio.h
xR3b0rn: The only thing I’d like to avoid is having to maintain a fork or manually integrate a patch into my build process, because I really need the CMA memory allocator. I’d therefore be perfectly happy with a CMake option.

iiod-client.c
xR3b0rn: Untested.

include/iio/iio.h
nunojsa: Sure, I fully understand and can relate to that

dNechita: @xR3b0rn would the environment variable approach work for you? This way, you won't have to reconfigure & recompile each time you want to change the allocator.

xR3b0rn: @dNechita yes

dNechita:
@xR3b0rn I am proposing the environment variable approach that can be seen here: #1344.
It should do the job. Your work will remain in the commit history and could be reused in the future if needed, but for now this level of configurability seems premature for libiio.

PR #1341, opened 2025-09-03
network: Translate WSA error codes to POSIX codes

dNechita:
## PR Description

It has been observed that creating a network iio context with a wrong IP (that doesn't point to a system with iio devices) will cause a crash when executed in a Windows environment.
This happens because:
 - the windows implementation of do_create_socket() returns WSA error
 - the WSA error has value -10060 which is outside of [-4095, -1]
 - WSA error code is returned by network_create_context()
 - iio_create_context() treats -10060 as a valid context address
 - iio_create_context() calls iio_context_update_scale_offset() causing the crash (access violation).

Fix this by translating a few common WSA error codes to POSIX codes that are know to be in the range of [1, 4095].

## PR Type

## PR Checklist

### DIFF
--- a/network-windows.c
+++ b/network-windows.c
@@ -103,6 +103,11 @@ bool network_connect_in_progress(int err)
 	return err == -EAGAIN;
 }
 
+int network_translate_so_error(int so_error)
+{
+	return translate_wsa_error_to_posix(so_error);
+}
+
 /* Use it if available */
 #ifndef WSA_FLAG_NO_HANDLE_INHERIT
 #define WSA_FLAG_NO_HANDLE_INHERIT 0
--- a/network.c
+++ b/network.c
@@ -206,8 +206,9 @@ static int do_connect(int fd, const struct addrinfo *addrinfo,
 	if(ret < 0)
 		return network_get_error();
 
-	if (error)
-		return -error;
+	if (error) {
+		return network_translate_so_error(error);
+	}
 
 	ret = set_blocking_mode(fd, true);
 	if (ret < 0)
--- a/network-unix.c
+++ b/network-unix.c
@@ -161,6 +161,11 @@ bool network_connect_in_progress(int err)
 	return err == -EINPROGRESS;
 }
 
+int network_translate_so_error(int so_error)
+{
+	return -so_error;
+}
+
 
 /* Use it if available */
 #ifndef SOCK_CLOEXEC
--- a/network.h
+++ b/network.h
@@ -45,6 +45,7 @@ int network_get_error(void);
 bool network_should_retry(int err);
 bool network_is_interrupted(int err);
 bool network_connect_in_progress(int err);
+int network_translate_so_error(int so_error);
 
 int network_platform_init(const struct iio_context_params *params);
 


network.c
```diff
@@ -210,6 +210,28 @@ static int do_connect(int fd, const struct addrinfo *addrinfo,
 
 	return 0;
 }
+#ifdef _WIN32
+static inline int translate_wsa_error_to_posix(const struct iio_context_params *params, int wsa_error)
+{
+	// Translate Windows-specific errors to standard errno values
+	if (wsa_error == -WSAETIMEDOUT)
+		return -ETIMEDOUT;
+	else if (wsa_error == -WSAECONNREFUSED)
```
nunojsa: redundant else if()

network.c
```diff
@@ -707,7 +732,11 @@ static struct iio_context * network_create_context(const struct iio_context_para
 
 	fd = create_socket(res, params->timeout_ms);
 	if (fd < 0) {
+#ifdef _WIN32
```
nunojsa: I would rather have a stub implementation of `translate_wsa_error_to_posix()` for when __WIN32 is not defined

network.c
```diff
@@ -54,8 +54,59 @@
...
+	case WSAESHUTDOWN:       return -ESHUTDOWN;
+#else
+	case WSAESHUTDOWN:       return -ECONNABORTED;
+#endif
+	case WSAENOTCONN:        return -ENOTCONN;
+	case WSAEOPNOTSUPP:      return -EOPNOTSUPP;
+	default: {
+		if (wsa_err > -4096 && wsa_err < 0) // pass through for POSIX errors
+			return wsa_err;
+		else
```
nunojsa: redundant else

network.c
```diff
@@ -54,8 +54,59 @@
...
+			return -EIO;   // generic fallback
+	}
```
nunojsa: DOo we need {} in here?

network.c
```diff
@@ -586,7 +644,8 @@ static struct iio_context * network_create_context(const struct iio_context_para
 	WSADATA wsaData;
 
 	ret = WSAStartup(MAKEWORD(2, 0), &wsaData);
-	if (ret < 0) {
+	if (ret != 0) {
```
nunojsa: Any reason to change to != 0?

network.c
```diff
@@ -553,13 +609,15 @@ static ssize_t network_write_data(struct iiod_client_pdata *io_ctx,
 				  const char *src, size_t len,
 				  unsigned int timeout_ms)
 {
-	return network_send(io_ctx, src, len, MSG_NOSIGNAL, timeout_ms);
+	return (ssize_t)translate_error_for_api(
+		(int)network_send(io_ctx, src, len, MSG_NOSIGNAL, timeout_ms));
```
nunojsa: Do we need the cast at all? In linux for sure we don't but maybe windows complains about it

network.c
```diff
@@ -54,8 +54,59 @@
 #ifndef MSG_NOSIGNAL
 #define MSG_NOSIGNAL 0
 #endif
+
+static int translate_wsa_error_to_posix(int wsa_err)
+{
+	switch (wsa_err) {
+	case WSAEACCES:          return -EACCES;
```
nunojsa: I would prefer to have the normal style and have the return in a new line

network.c
```diff
@@ -380,7 +433,8 @@ static ssize_t network_read_attr(const struct iio_attr *attr,
 	const struct iio_context *ctx = iio_device_get_context(dev);
 	struct iio_context_pdata *pdata = iio_context_get_pdata(ctx);
 
-	return iiod_client_attr_read(pdata->iiod_client, attr, dst, len);
+	return (ssize_t)translate_error_for_api(
+		(int)iiod_client_attr_read(pdata->iiod_client, attr, dst, len));
```
nunojsa: I'm not really excited about the above and how easy it might be to forget about it. Not sure it works but here it goes an idea... Being this windows specific, what about trying to put/hide this inside network-windows.c? For instance in:

```
int network_get_error(void)
{
	return -WSAGetLastError();
}
```
We could already take care of translation and so we would make things consistent and more cleaner in the core network file. Ideally, in that file we should not have to care about being on windows or unix. We could the same  in functions like `set_blocking_mode()`, `do_create_socket()` and `do_create_socket()`.

And if the above works, I think we could have a generic (with posix handling) for:

`network_is_interrupted()` and `network_connect_in_progress()`

Thoughts?

network.c
dNechita: Yes, WSAStartup() returns 0 on success and a positive error on failure. You can see more on the commit description. Now I see that I messed up the commits. I used to have 2 commits, 1st was about this change, and the 2nd was about the hole translation. The message of the 2nd commit got lost but I will restore it.

network.c
dNechita: It would make more sense to have the translation inside the Windows specific files (network-windows.c and dns_sd_windows.c). I will give it another go.

network.c
dNechita: Windows complains by issuing warnings and CI builds treat them as errors causing the jobs to fail.

network.c
dNechita: not really. No local variable is being declared here so {} are actually not needed.

network-windows.c
```diff
@@ -24,8 +24,7 @@ int set_blocking_mode(int s, bool blocking)
 
 	ret = ioctlsocket(s, FIONBIO, &nonblock);
 	if (ret == SOCKET_ERROR) {
-		ret = -WSAGetLastError();
-		return ret;
+		return network_get_error();
```
nunojsa, line 27: nit: given that you are already in windows specific code, I would directly call translate_wsa_error_to_posix(). Not a big deal though...

network-windows.c
```diff
@@ -101,7 +100,7 @@ bool network_is_interrupted(int err)
 
 bool network_connect_in_progress(int err)
 {
-	return err == -WSAEWOULDBLOCK;
+	return err == -EAGAIN;
```
nunojsa, line 103: Couldn't we actually get rid of the above two functions now that we have that error code transalation?

nunojsa (approved):
Just some minor stuff from me. The real reason to spin another version is if we can get rid of those two functions in unix and windows code.

network-windows.c
dNechita: That was my intention as well, but I couldn't do it without slightly changing the logic:
The current logic:
Windows: network_connect_in_progress returns true when EAGAIN
Unix: network_connect_in_progress returns true when EINPROGRESS

Windows: network_should_retry returns true when EAGAIN || ETIMEDOUT
Unix: network_should_retry returns true when EAGAIN

I could remove both network_connect_in_progress() and network_should_retry() but then I would need to somehow merge the conditions coming from Windows and Unix. For example, network_connect_in_progress() could become expression: `EAGAIN || EINPROGRESS` but I don't know what the implication would then be. Do you think it would be harmless to do so?

nunojsa: > Windows: network_should_retry returns true when EAGAIN || ETIMEDOUT

I wonder about why ETIMEDOUT is treated as should retry on windows...

> For example, network_connect_in_progress() could become expression: EAGAIN || EINPROGRESS but I don't know what the implication would then be. Do you think it would be harmless to do so?

Yeah, maybe not a good idea... I don't really now why the windows conditions have to be different so better to keep them. Looks good then!

rgetz:
I think from looking at:
https://learn.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2

you are still missing some  - depends on how complete you want to be - I don't think you would see many (other than the DNS issues)?

System initialization and shutdown errors
- WSASYSNOTREADY (10091) – Network subsystem unavailable
- WSAVERNOTSUPPORTED (10092) – Winsock version out of range
- WSANOTINITIALISED (10093) – WSAStartup not called

Overlapped I/O / asynchronous operations
- WSA_IO_INCOMPLETE (996)
- WSA_IO_PENDING (997)
- WSA_OPERATION_ABORTED (995)

Name resolution / DNS errors
- WSAHOST_NOT_FOUND (11001)
- WSATRY_AGAIN (11002)
- WSANO_RECOVERY (11003)
- WSANO_DATA (11004)

Quality of Service (QoS) errors
- WSA_QOS_* series (11005–11031)

Other service/provider errors
- WSAENOMORE, WSAECANCELLED, WSAEINVALIDPROCTABLE, WSAEINVALIDPROVIDER, WSAEPROVIDERFAILEDINIT, WSASYSCALLFAILURE, WSASERVICE_NOT_FOUND, WSATYPE_NOT_FOUND, WSA_E_NO_MORE, WSA_E_CANCELLED, WSAEREFUSED

PR #1344, opened 2025-09-05
Revert changing the DMA allocator via public API and use environment variables

dNechita:
## PR Description

"This reverts the functionality that allowed users to change the DMA allocator type through the public API. The use case is niche and rarely needed at present. The existing functionality supported two options: the default 'system' and 'cma,linux'—the latter being non-standard and system-dependent (see #1329), often requiring a custom string. Supporting this through the public API adds unnecessary complexity and risks future breakage. Instead, I am introducing an alternative approach for changing the DMA allocator that avoids impacting the API.

## PR Type

## PR Checklist

### DIFF
--- a/README_BUILD.md
+++ b/README_BUILD.md
@@ -164,6 +164,46 @@ It will try to link against the `call_function` symbol even though it's clearly 
 
 For this reason, when building with MSVC, please build in `RelWithDebInfo` mode. If you try to build in `Debug` mode, it will error.
 
+## Environment Variable Configuration
+
+### DMA Heap Path Configuration
+
+libiio supports configuring the DMA heap path globally through the `LIBIIO_DMA_HEAP_PATH` environment variable. This overrides the default `/dev/dma_heap/system` path for all IIO devices.
+
+#### Supported Format (Global Only)
+```bash
+export LIBIIO_DMA_HEAP_PATH=heap_name
+```
+This will use `/dev/dma_heap/<heap_name>` for every device.
+
+#### Accepted Values
+
+The environment variable accepts only the following predefined heap names:
+- `system` (default when unset or empty)
+- `default_cma_region`
+- `reserved`
+- `linux,cma`
+- `default-pool`
+
+**Examples:**
+```bash
+export LIBIIO_DMA_HEAP_PATH=default_cma_region
+./an_iio_application
+
+export LIBIIO_DMA_HEAP_PATH=reserved
+./an_iio_application
+```
+
+#### Behavior and Error Handling
+
+- **Unset or empty**: Defaults to `system` heap (`/dev/dma_heap/system`)
+- **Valid value**: Uses the specified heap (`/dev/dma_heap/<heap_name>`)
+- **Invalid value**: **Operation fails with error** - no fallback occurs
+
+Setting an invalid heap name will cause DMABUF operations to fail immediately with an error message listing the accepted values. This ensures users are aware when their configuration is incorrect rather than silently using a fallback.
+
+This feature is intended for users who need to select an alternative DMA heap present under `/dev/dma_heap/` (for example a reserved or CMA heap).
+
 ## Instructions applicable to Microcontroller configurations
 
 ### Install Prerequisites/Dependencies


README_BUILD.md
```diff
@@ -161,6 +161,65 @@ It will try to link against the `call_function` symbol even though it's clearly
...
+#### Constraints and Error Handling
+
+- Heap names must not exceed 64 characters
+- Device names are matched exactly (case-sensitive)
+- Whitespace around device names is automatically trimmed
+- Invalid configurations fall back to the default "system" heap
+- If the environment variable is not set, the default `/dev/dma_heap/system` is used
+
+This feature is intended for advanced users who understand the implications of customizing DMA heap selection. It is especially useful on systems with multiple DMA heap types, allowing experienced users to exercise fine-grained control over memory allocation for specific IIO devices.
+
```
nunojsa, line 206: I would separate docs from the code in terms of commits...

local-dmabuf.c
```diff
@@ -55,6 +59,87 @@ struct dma_buf_sync {
 	uint64_t flags;
 };
 
+/* Parse environment variable to get DMA heap path for all or a specific device.
+ * Environment variable format:
+ * - LIBIIO_DMA_HEAP_PATH=heap_name (applies to all devices)
+ * - LIBIIO_DMA_HEAP_PATH=heap_name:device1 (applies only to device1)
+ * - LIBIIO_DMA_HEAP_PATH=heap_name:device1,device2 (applies to device1 and device2)
```
nunojsa: Hmm, I would make this way more simple for now. I guess the ENV var is ok (even though I would likely go for the compile time option) but make it simple. Meaning, just give in the allocator name. For now this is a system thing and we don't really need to have a per device setting. Let's deal with this when we really have an use case for that.

The above is one of the reasons why I would likely go with a build option. ATM, this is kind of a system level decision and the build option would be the simpler approach for that. But as I said, I guess I can live with the env var (even though that brings a runtime penalty - but not when streaming at least).

local-dmabuf.c
```diff
@@ -55,6 +58,33 @@ struct dma_buf_sync {
...
+        return "system";
+
+    size_t env_len = strnlen(env_value, 65); /* detect overflow beyond 64 */
+	if (env_len == 0) {
+		return "system";
+	}
+	if (env_len > 64) {
+		prm_warn(NULL, "LIBIIO_DMA_HEAP_PATH value too long (>64); falling back to system heap\n");
+		return "system";
+	}
```
nunojsa, line 86: Not sure if it is github but it seems you somehow screwed indentation?

local-dmabuf.c
```diff
@@ -55,6 +58,33 @@ struct dma_buf_sync {
...
+    const char *env_value = getenv(LIBIIO_DMA_HEAP_ENV_VAR);
+    static char heap_name[65]; /* 64 chars + null terminator */
+
+    if (!env_value)
+        return "system";
+
+    size_t env_len = strnlen(env_value, 65); /* detect overflow beyond 64 */
+	if (env_len == 0) {
+		return "system";
+	}
```
nunojsa: nit: no need for {}

local-dmabuf.c
```diff
@@ -55,6 +58,33 @@ struct dma_buf_sync {
...
+static const char *get_dma_heap_name(void)
+{
+    const char *env_value = getenv(LIBIIO_DMA_HEAP_ENV_VAR);
+    static char heap_name[65]; /* 64 chars + null terminator */
+
+    if (!env_value)
+        return "system";
+
+    size_t env_len = strnlen(env_value, 65); /* detect overflow beyond 64 */
+	if (env_len == 0) {
```
nunojsa: nit: slight preference for `if (!env_len)`

local-dmabuf.c
```diff
@@ -55,6 +58,33 @@ struct dma_buf_sync {
...
+ */
+static const char *get_dma_heap_name(void)
+{
+    const char *env_value = getenv(LIBIIO_DMA_HEAP_ENV_VAR);
+    static char heap_name[65]; /* 64 chars + null terminator */
+
+    if (!env_value)
+        return "system";
+
+    size_t env_len = strnlen(env_value, 65); /* detect overflow beyond 64 */
```
nunojsa: since we mostly follow kernel coding style, I would not mix code and declarations... Also, `sizeof(heap_name)`

local-dmabuf.c
```diff
@@ -55,6 +58,33 @@ struct dma_buf_sync {
...
+	if (env_len == 0) {
+		return "system";
+	}
+	if (env_len > 64) {
```
nunojsa: I think this one could be `env_len == sizeof(heap_name)`. From strnlen() man:

" The strnlen() function returns strlen(s), if that is less than maxlen, or maxlen if there is no null terminating ('\0')  among  the  first  maxlen characters pointed to by s."

local-dmabuf.c
```diff
@@ -23,6 +23,9 @@
 #include <time.h>
 #include <unistd.h>
 
+#define LIBIIO_DMA_HEAP_ENV_VAR "LIBIIO_DMA_HEAP_PATH"
+#define MAX_DMA_HEAP_PATH 256
```
nunojsa: seems way to big given than we only allow 65 (including NULL character)

local-dmabuf.c
```diff
@@ -55,6 +58,33 @@ struct dma_buf_sync {
...
+
+    size_t env_len = strnlen(env_value, 65); /* detect overflow beyond 64 */
+	if (env_len == 0) {
+		return "system";
+	}
+	if (env_len > 64) {
+		prm_warn(NULL, "LIBIIO_DMA_HEAP_PATH value too long (>64); falling back to system heap\n");
+		return "system";
+	}
+
```
nunojsa, line 87: Actually rather than the above (checking for the size) I wonder if we should properly validate it? We should only match for:

```
"linux,cma" 
"reserved"
"system"
```

If we get something different, no point in trying to open(). I would just print a meaningful message and return error

dNechita: I thought it is possible to have other names that would come from the device tree CMA node name.

nunojsa: Ok, It actually seems that the name  depends on the DT node name but it seems we have standard names for it. And we can actually do better than libcamera (and also things changed in recent kernels). So, the below should be what we follow:

https://elixir.bootlin.com/linux/v6.18-rc6/source/Documentation/userspace-api/dma-buf-heaps.rst#L19

So in newer kernel, `default_cma_region` should be the only possible name. For legacy, the ones I mentioned and it also looks like `default-pool` is a possibility.

dNechita: Thanks @nunojsa, this is really useful information.
I updated things to include your suggestion.

local-dmabuf.c
dNechita: Makes sense. Changed to 64 since "/dev/dma_heap/" is 14 characters long. The remaining difference is enough to fit one of the values from the accepted names list.

local-dmabuf.c
dNechita: I haven't seen any misalignments or anything. But this has been removed anyway.

local-dmabuf.c
```diff
@@ -55,6 +58,41 @@ struct dma_buf_sync {
...
+{
+	const char *env_value = getenv(LIBIIO_DMA_HEAP_ENV_VAR);
+	static const char *accepted_heaps[] = {
+		"system",
+		"default_cma_region",
+		"reserved",
+		"linux,cma",
+		"default-pool"
+	};
+	static const size_t num_accepted_heaps = sizeof(accepted_heaps) / sizeof(accepted_heaps[0]);
```
nunojsa: I would imagine libiio defines somewhere `ARRAY_SIZE()`. If we have it, use it. Also there's no point in making it const and  more important `static`. Why keeping that memory around after the call?

local-dmabuf.c
```diff
@@ -55,6 +58,41 @@ struct dma_buf_sync {
...
+			return accepted_heaps[i];
+	}
+
+	/* Environment variable value is not in the accepted pool */
+	prm_warn(NULL, "LIBIIO_DMA_HEAP_PATH value '%s' is not recognized. "
+		"Accepted values are: \"system\", \"default_cma_region\", "
+		"\"reserved\", \"linux,cma\", \"default-pool\". "
+		"Defaulting to \"system\"\n", env_value);
+
+	return "system";
```
nunojsa: Not sure if this should not be an error given that the user indeed made an error and given that this happens locally one cannot easily realize what's the issue.

That said, up to you

local-dmabuf.c
```diff
@@ -78,13 +116,24 @@ local_create_dmabuf(struct iio_buffer_pdata *pdata, size_t size, void **data)
...
 
 	priv = zalloc(sizeof(*priv));
 	if (!priv)
 		return iio_ptr(-ENOMEM);
 
-	devfd = open("/dev/dma_heap/system", O_RDONLY | O_CLOEXEC | O_NOFOLLOW); /* Flawfinder: ignore */
+	/* Determine DMA heap path from global environment variable */
+	heap_name = get_dma_heap_name();
+
+	ret = snprintf(dma_heap_path, sizeof(dma_heap_path), "/dev/dma_heap/%s", heap_name);
```
nunojsa, line 134: If we go the path of defaulting to "system" in case of error, you can just call `get_dma_heap_name()` inline in `snprintf()`

nunojsa (commented):
Just some minor stuff from me

local-dmabuf.c
dNechita: I found ARRAY_SIZE() definition. Also, I removed the const and static.

local-dmabuf.c
dNechita: If I were the user, I would like to get an error instead of the library doing some else silently.
Updated.

local-dmabuf.c
dNechita: Not applicable anymore.

PR #1345, opened 2025-09-15
Several small fixes

dNechita:
## PR Description

Several small fixes came up along the way and added up to a noticeable change.

## PR Type

## PR Checklist

### DIFF
--- a/README_BUILD.md
+++ b/README_BUILD.md
@@ -102,7 +102,7 @@ Cmake Options     | Default | Description                                    |
 |                 |   Info  | Info    : Print info, warnings and errors      |
 |                 |         | Debug   : Print debug/info/warnings/errors (very verbose)  |
 
-Options which effect iiod only. These are only available on Linux.
+Options which affect iiod only. These are only available on Linux.
 
 Cmake Options       | Default | Description                                    |
 ------------------- | ------- | ---------------------------------------------- |
--- a/examples/ad9371-iiostream.c
+++ b/examples/ad9371-iiostream.c
@@ -188,7 +188,14 @@ bool cfg_ad9371_streaming_ch(struct stream_cfg *cfg, enum iodev type, int chid)
 }
 
 /* simple configuration and streaming */
-int main (__notused int argc, __notused char **argv)
+/* usage:
+ * Default context, assuming local IIO devices, i.e., this script is run on a platform to
+ * which ad9371 is connected.
+ $./a.out
+ * URI context, find out the uri by typing `iio_info -s` at the command line of the host PC
+ $./a.out usb:x.x.x
+ */
+int main (int argc, char **argv)
 {
 	// Streaming devices
 	struct iio_device *tx;
@@ -213,10 +220,15 @@ int main (__notused int argc, __notused char **argv)
 	txcfg.lo_hz = GHZ(2.5); // 2.5 GHz rf frequency
 
 	printf("* Acquiring IIO context\n");
-
-	ctx = iio_create_context(NULL, NULL);
-	err = iio_err(ctx);
-	IIO_ENSURE(!err && "No context");
+	if (argc == 1) {
+		ctx = iio_create_context(NULL, NULL);
+		err = iio_err(ctx);
+		IIO_ENSURE(!err && "No context");
+	} else if (argc == 2) {
+		ctx = iio_create_context(NULL, argv[1]);
+		err = iio_err(ctx);
+		IIO_ENSURE(!err && "No context");
+	}
 	IIO_ENSURE(iio_context_get_devices_count(ctx) > 0 && "No devices");
 
 	printf("* Acquiring AD9371 streaming devices\n");
--- a/examples/ad9371-iiostream.c
+++ b/examples/ad9371-iiostream.c
@@ -55,8 +55,8 @@ static struct iio_stream  *txstream = NULL;
 static struct iio_channels_mask *rxmask = NULL;
 static struct iio_channels_mask *txmask = NULL;
 
-/* cleanup and exit */
-static void shutdown(void)
+/* cleanup resources */
+static void cleanup(void)
 {
 	printf("* Destroying streams\n");
 	if (rxstream) {iio_stream_destroy(rxstream); }
@@ -72,7 +72,13 @@ static void shutdown(void)
 
 	printf("* Destroying context\n");
 	if (ctx) { iio_context_destroy(ctx); }
-	exit(0);
+}
+
+/* cleanup and exit with error */
+static void shutdown(void)
+{
+	cleanup();
+	exit(1);
 }
 
 static void handle_sig(int sig)
@@ -301,7 +307,7 @@ int main (int argc, char **argv)
 	printf("* Starting IO streaming (press CTRL+C to cancel)\n");
 	stream(rx_sample_sz, tx_sample_sz, BLOCK_SIZE,
 	       rxstream, txstream, rx0_i, tx0_i);
-	shutdown();
+	cleanup();
 
 	return 0;
 }
--- a/examples/adrv9002-iiostream.c
+++ b/examples/adrv9002-iiostream.c
@@ -222,7 +222,14 @@ stream_channels_get_mask(const struct iio_device *dev, struct iio_channel **chan
 	return mask;
 }
 
-int main(void)
+/* usage:
+ * Default context, assuming local IIO devices, i.e., this script is run on a platform to
+ * which adrv9002 is connected.
+ $./a.out
+ * URI context, find out the uri by typing `iio_info -s` at the command line of the host PC
+ $./a.out usb:x.x.x
+ */
+int main(int argc, char **argv)
 {
 	struct iio_device *tx;
 	struct iio_device *rx;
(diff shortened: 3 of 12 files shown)


README_BUILD.md
```diff
@@ -102,7 +102,7 @@ Cmake Options     | Default | Description                                    |
 |                 |   Info  | Info    : Print info, warnings and errors      |
 |                 |         | Debug   : Print debug/info/warnings/errors (very verbose)  |
 
-Options which effect iiod only. These are only available on Linux.
+Options which affect iiod only. These are only available on Linux.
```
nunojsa, line 105: git subject a bit odd. The "Fix typo:" part is not that usual :)

README_BUILD.md
```diff
@@ -48,7 +48,7 @@ Cmake Options          | Default | Target | Description
 `CPP_EXAMPLES`         | OFF |        All | Build C++ examples (C++17 required) |
 `PYTHON_BINDINGS`      | OFF |        All | Install PYTHON bindings                            |
 `WITH_UTILS`           |  ON |        All | Build the utility programs (iio-utils)           |
-`WITH_EXAMPLES`        | OFF |        All | Build the example programs                         |
+`WITH_EXAMPLES`        | OFF |        Linux, MaC | Build the example programs                         |
```
nunojsa: Not on windows?

dNechita: Actually, they work on Windows, but one example needs a bit of tweaking. I will remove this change.
Thanks!

dNechita:
V1:

- Undo the statement that examples work only on Linux and Mac and not Windows
- Fix adrv9002 not building on windows
- Update the list of examples in examples/README by adding the adrv9002-iiostream
- Update examples: adrv9002-iiostream, adrv9009-iiostream and ad9371-iiostream to take an URI via command line

README_BUILD.md
dNechita: I went with "fix grammar". Seems more accurate :)

nunojsa: Oh, not my point :). The `:` separators is for kind of directory walk. So having `README_BUILD.md: Fix grammar: ` is odd. What I meant was something like `README_BUILD.md: Fix grammar (use  'affect' instead of 'effect')`. I would also strip the files suffix from the git subject. So, just `README_BUILD: ...`

examples/adrv9002-iiostream.c
```diff
@@ -18,13 +18,11 @@
...
-#define FMT(fmt, ...)	fmt
-#define error(...) \
-	printf("%s, %d: ERROR: " FMT(__VA_ARGS__, 0)"%s", __func__, __LINE__, ARGS(__VA_ARGS__, ""))
+#define error(fmt, ...) \
+	printf("%s, %d: ERROR: " fmt, __func__, __LINE__, ##__VA_ARGS__)
 
-#define info(...) \
-	printf("%s, %d: INFO: " FMT(__VA_ARGS__, 0)"%s", __func__, __LINE__, ARGS(__VA_ARGS__, ""))
+#define info(fmt, ...) \
+	printf("%s, %d: INFO: " fmt, __func__, __LINE__, ##__VA_ARGS__)
```
nunojsa: Be careful with these changes... Did you tested it on some older windows compilers? IIRC, I had to do all of that gymnastics with the macros exactly because of windows.

AFAIK, ##__VA_ARGS__ is a gnu extension and I thought we were not using it? So I'm a bit confused how the above compiles on window. With what compiler did you tested it? Maybe newer compilers support it...

examples/adrv9009-iiostream.c
```diff
@@ -206,9 +213,16 @@ int main (__notused int argc, __notused char **argv)
...
 	printf("* Acquiring IIO context\n");
-	ctx = iio_create_context(NULL, NULL);
-	err = iio_err(ctx);
-	IIO_ENSURE(!err && "No context");
+	if (argc == 1) {
+		ctx = iio_create_context(NULL, NULL);
+		err = iio_err(ctx);
+		IIO_ENSURE(!err && "No context");
+	}
+	else if (argc == 2) {
```
nunojsa: } else if (...)?

examples/adrv9002-iiostream.c
```diff
@@ -226,11 +233,22 @@ int main(void)
...
+	printf("* Acquiring IIO context\n");
+	if (argc == 1) {
+		ctx = iio_create_context(NULL, NULL);
+		ret = iio_err(ctx);
+		if (ret) {
+			error("Could not create IIO context\n");
+			return EXIT_FAILURE;
+		}
+	}
+	else if (argc == 2) {
```
nunojsa: ditto

examples/ad9371-iiostream.c
```diff
@@ -188,7 +188,14 @@ bool cfg_ad9371_streaming_ch(struct stream_cfg *cfg, enum iodev type, int chid)
...
 /* simple configuration and streaming */
-int main (__notused int argc, __notused char **argv)
+/* usage:
+ * Default context, assuming local IIO devices, i.e., this script is run on a platform to
+ * which ad9371 is connected.
+ $./a.out
+ * URI context, find out the uri by typing `iio_info -s` at the command line of the host PC
+ $./a.out usb:x.x.x
+ */
+int main (int argc, char **argv)
```
nunojsa, line 204: Not sure if we should not have this a patch per file...

nunojsa (commented):
My main concern is with the macro change in the adrv9002 iio stream example.

examples/adrv9002-iiostream.c
dNechita: I compiled it with MSVC version 16.11 (that came with Visual Studio 2019).
The ##_VA_ARGS is a GNU extension indeed but has been standardized in C2x. I wouldn't worry about this not working with older windows compilers, as long as it's not part of the library itself.

nunojsa:
> It would make sure the examples stay buildable in the future. I will update CI so that at least one job builds them.

I would say that we should enable them at least:

 * one Linux distro;
 * all macOs jobs;
 * all windows jobs;

On linux we can be a bit more confident that if the examples build on one distro, they should be build in the others as well. For the other OSs I just don't know them enough so I would be careful. OTHO, building the examples should be fast enough so we could just enabled it everywhere.

examples/adrv9002-iiostream.c
dNechita: Is there a reason for why we need the examples to be built without GNU extensions?

nunojsa:
> I enabled the examples on all windows jobs where a few warnings are being thrown making the jobs fail. I need to figure out whether to fix them or not treat them as errors.

yes, I suspected that would happen

examples/adrv9002-iiostream.c
nunojsa: Note that I'm aware. I guess it's just a matter of portability since the examples should run in multiple platforms.

> The ##_VA_ARGS is a GNU extension indeed but has been standardized in C2x.

Yes, I assumed you're compiling using a recent standard.

> I wouldn't worry about this not working with older windows compilers, as long as it's not part of the library itself.

I'm not going to argue much here as I'm not aware about what is our policy (commitment) in terms of compilers (and C standards) support. But this always opens the door to get someone opening issues or complaining about a given piece of code not compiling in the environment.

dNechita: From the initial commit of the CMake file for these examples, C99 was specified and C extensions were explicitly disabled. I would keep this policy for now. I have found a solution that doesn’t rely on extensions.

dNechita:
V2:

- Removed ##__VA_ARGS as it depends on GNU extensions when using older compilers (that don't fully use the C23 standard)
- Fixed warnings (that are treated as errors) about code `return 0;` that never gets reached
- Fixed warnings about implicit conversions between `ssize_t` and `int`
- Changed cmake such that wingetopt gets compiled into a separate library and each example link to it. This helps control how wingetopt gets compiled allowing flexibility when needing to silence warnings, etc.
- The `shutdown()` that is used by many examples will now exit with code 1 (instead of 0) to indicate an error occurred.
- Enabled the build of examples on CI, for Ubuntu 24.04 and for all MSVC builds. Observed that examples were already enabled for all MacOS builds and for all ARM builds.

nunojsa (commented):
LGTM. Only complain is with the lack of consistency with git subjects. Like we have:

`examples: ...`
`example/ad9361: ...`

And If I'm not mistaken I think the most used style in libiio is similar to linux. So it would be eg: `example: ad9361-iiostream: ...`. 

Another mismatch is `cmake: ...`  vs `CMake:...`. I would just go with the most used style.

I know the above might be a pain in the ass given the amount of patches in the PR and that I might be annoying but I do really believe that consistency matters in your git log.

dNechita:
> LGTM. Only complain is with the lack of consistency with git subjects. Like we have:
> 
> `examples: ...` `example/ad9361: ...`
> 
> And If I'm not mistaken I think the most used style in libiio is similar to linux. So it would be eg: `example: ad9361-iiostream: ...`.
> 
> Another mismatch is `cmake: ...` vs `CMake:...`. I would just go with the most used style.
> 
> I know the above might be a pain in the ass given the amount of patches in the PR and that I might be annoying but I do really believe that consistency matters in your git log.

I appreciate the feedback. The subjects of some of my commits could use a bit more consistency.
Fixed!

PR #1348, opened 2025-09-30
libiio: backend: Add atomic register access support for local backend

mhennerich:
## PR Description

This patch introduces thread-safe register read/write operations for the local backend using `direct_reg_access` via debugfs. It adds two new backend ops:
- `reg_read()`
- `reg_write()`

These functions use file locking (`flock`) to ensure atomic access and prevent race conditions when multiple threads interact with the same register interface.

The fallback logic in `iio_device_reg_read()` and `iio_device_reg_write()` is preserved for backends that do not implement these atomic operations.

This improves reliability for multi-threaded applications using libiio with the local backend.

## PR Type
- [X] Bug fix (a change that fixes an issue)

## PR Checklist
- [X] I have conducted a self-review of my own code changes
- [X] I have commented new code, particularly complex or unclear areas
- [X] I have checked that I did not introduce new warnings or errors (CI output)
- [X] I have checked that components that use libiio did not get broken


local.c
```diff
@@ -714,7 +714,6 @@ static ssize_t local_read_dev_attr(const struct iio_device *dev,
 	else
 		dst[0] = '\0';
 
-	fflush(f);
```
nunojsa, line 715: The commit message has some undesired characters :)

local.c
```diff
@@ -1151,6 +1153,113 @@ static int local_set_trigger(const struct iio_device *dev,
...
+		uint32_t address, uint32_t *value)
+{
+	char buf[1024];
+	char addr_str[32];
+	char val_str[32];
+	FILE *f;
+	int fd;
+	ssize_t ret;
+
+	/* Get the path to direct_reg_access */
```
nunojsa: I would get rid of obvious comments. IMO, they don't really add much

local.c
```diff
@@ -1151,6 +1153,113 @@ static int local_set_trigger(const struct iio_device *dev,
...
+	if ((size_t)ret != strlen(addr_str) || ferror(f)) {
+		ret = -errno;
+		goto unlock;
+	}
+
+	/* Read the register value */
+	rewind(f);
+	ret = fread(val_str, 1, sizeof(val_str) - 1, f);
+	if (!feof(f))
+		ret = -EFBIG;
```
nunojsa, line 1196: Shouldn't we `goto unlock`? Furthermore, I find the below also a bit confusing. I would group errors together. Something like:

```
if (!feof(f)) {
		ret = -EFBIG;
        goto unlock;
}

if (ferror(f)) {
		ret = -errno;
		goto unlock;
}

// then the other if() ele for val_str
....
```

local.c
```diff
@@ -1151,6 +1153,113 @@ static int local_set_trigger(const struct iio_device *dev,
...
+		fclose(f);
+		return ret;
+	}
+
+	/* Format and write the register address and value */
+	iio_snprintf(write_str, sizeof(write_str), "0x%" PRIx32 " 0x%" PRIx32, address, value);
+	ret = fwrite(write_str, 1, strlen(write_str), f);
+	fflush(f);
+	if ((size_t)ret != strlen(write_str) || ferror(f)) {
+		ret = -errno;
```
nunojsa: I think errno is only set in case of `ferror()` right? In case we get here because of the first condition we might have some unexpected behavior.

nunojsa (commented):
Some minor stuff from me. IMO, I would also split it in two patches:

`local: Add atomic debug register access`
`device: Make use of the new atomic register access`

local.c
mhennerich: Fixed

local.c
mhennerich: ACK

local.c
mhennerich: makes sense

local.c
mhennerich: if ((size_t)ret < strlen(write_str)) {
		if (ferror(f))
			ret = -errno;
		else
			ret = -EIO;
	} else {
		ret = 0;
	}

dNechita (approved):

mhennerich:
pushed a new version.

PR #1352, opened 2025-10-07
Handle buffer creation concurrency

dNechita:
## PR Description

IIOD is crashing when 2 or more clients are trying to create the same buffer (for the same iio device and with the same index).
With these changes, the buffer can be created only once, and later clients will be informed that it’s already been created.

Testing this revealed a missing discard() function in the iiod responder which has been included in this PR.

## PR Type

## PR Checklist


iiod/responder.c
```diff
@@ -1250,10 +1250,36 @@ static ssize_t iiod_write(void *d, const struct iiod_buf *buf, size_t nb)
 	return write_all(d, buf->ptr, buf->size);
 }
 
+static ssize_t iiod_discard(void *d, size_t bytes)
+{
+	struct parser_pdata *pdata = d;
+	char buf[0x1000];
```
nunojsa: I would rather prefer to have it decimal. I know it's the same style as in the client but I would rather change it in there as well (in another PR for example).

Also feels that with a bit of rework we could unify this with the client code and share most of the code. But let's leave that for another day

iiod/responder.c
```diff
@@ -367,23 +367,53 @@ static void handle_create_buffer(struct parser_pdata *pdata,
...
+	iio_mutex_lock(buflist_lock);
+
+	SLIST_FOREACH(existing_entry, &bufferlist, entry) {
+		if (existing_entry->dev == dev && existing_entry->idx == buffer_idx) {
+			iio_mutex_unlock(buflist_lock);
+			free(placeholder_entry);
+			ret = -EEXIST;
+			goto err_send_response;
+		}
+	}
```
nunojsa: I'm not sure all this dance is worth it in comparison with locking the whole thing. I mean, I think that until after `iio_create_channels_mask()` should be thread safe. After that point we start already configuring the device so I would say  to lock the `buflist_lock` after that call. Then, you could use `get_iio_block_unlocked()` to check for the buffer existence and act accordingly. I think it should work and IMO, it's much simpler. I also dunno for any contention issue (and this should not be a fast path anyways so I would go for simplicity)

nunojsa (commented):
Here it goes my 2cents.

iiod/responder.c
dNechita: I modified to use decimal.

iiod/responder.c
dNechita: I went with your approach. I needed `get_iio_buffer_unlocked()` which wasn't implemented so I added in this PR.

nunojsa: > I went with your approach. I needed get_iio_buffer_unlocked() which wasn't implemented so I added in this PR.

Should be here:

https://github.com/analogdevicesinc/libiio/blob/main/iiod/responder.c#L552

nunojsa: Oh forget about it. That's to get a block

iiod/responder.c
```diff
@@ -32,6 +32,10 @@ static inline uint32_t ntohl(uint32_t n) {
 #endif
 
 /* Forward declaration */
+static struct iio_buffer * get_iio_buffer_unblocked(struct parser_pdata *pdata,
+					  const struct iiod_command *cmd,
+					  struct buffer_entry **entry_ptr);
+
```
nunojsa: nit: I would avoid the above and just define the function in proper place

iiod/responder.c
```diff
@@ -434,6 +438,19 @@ static void handle_create_buffer(struct parser_pdata *pdata,
...
+	entry->dev = dev;
+	entry->idx = (uint16_t) cmd->code;
+
+	buf = get_iio_buffer_unblocked(pdata, cmd, &entry);
+	ret = iio_err(buf);
+	if (ret != -EBADF) {
+		ret = -EEXIST;
+		iio_mutex_unlock(buflist_lock);
+		goto err_free_mask;
```
nunojsa, line 474: For consistency you could do the same as when creating a block:

https://github.com/analogdevicesinc/libiio/blob/main/iiod/responder.c#L703

A log message could also be useful but arguably -EEXIST is already self explanatory so up to you.

iiod/responder.c
dNechita: makes sense. Less code, simpler to read.

PR #1355, opened 2025-10-09
utils-windows: Add more WSA error codes

dNechita:
Add some more WSA error codes to the translation function what are likely to occur.

## PR Description

Add more WSA errors to the translation function that are likely to be encountered.

## PR Type

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,6 +248,10 @@ check_symbol_exists(strtok_r "string.h" HAS_STRTOK_R)
 check_symbol_exists(newlocale "locale.h" HAS_NEWLOCALE)
 check_symbol_exists(clock_gettime "time.h" HAS_CLOCK_GETTIME)
 
+# For EAI_* codes
+include(CheckIncludeFile)
+check_include_file("netdb.h" HAVE_NETDB_H)
+
 if (HAS_CLOCK_GETTIME)
 	target_compile_definitions(iio PRIVATE CLOCK_GETTIME_AVAILABLE=1)
 endif()
--- a/iio-config.h.cmakein
+++ b/iio-config.h.cmakein
@@ -52,6 +52,7 @@
 #cmakedefine HAS_NEWLOCALE
 #cmakedefine HAS_PTHREAD_SETNAME_NP
 #cmakedefine HAVE_IPV6
+#cmakedefine01 HAVE_NETDB_H
 
 #define IF_ENABLED(cfg, ptr) ((cfg) ? (ptr) : NULL)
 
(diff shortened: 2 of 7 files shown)


dNechita:
> I don't know if you want to fix the rest of the `network.c` code at the same time?

It would make sense to do so.

> ret = setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, ...);
> ptr2 = inet_ntop(AF_INET6, &in->sin6_addr, description, INET6_ADDRSTRLEN);

I have fixed both of the above.

> and in `network-windows.c` there is alot of missing error checking in these WinSock functions:
 > * [getnameinfo()](https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getnameinfo) Any nonzero return value indicates failure and a specific error code can be retrieved by calling [WSAGetLastError]

The errors from `getnameinfo()` are ignored so no WSA error codes can leak from here.

> * [ioctlsocket()](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-ioctlsocket) the ioctlsocket returns zero. Otherwise, a value of SOCKET_ERROR is returned, and a specific error code can be retrieved by calling [WSAGetLastError]

In dns_sd_windows the return value is ignored and in network-windows.c the return value is already handled.
> * [select()](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-select) SOCKET_ERROR if an error occurred. If the return value is SOCKET_ERROR, [WSAGetLastError]

Fixed.

> And this Win32 function: [GetAdaptersAddresses()](https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses) If the function fails, the return value is one of the following error codes (which are outside the POSIX range : https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--1000-1299-)

The return values are never returned as the function calling GetAdaptersAdresses() is only interested in returning the number of sockets.

Additionaly, I found that [getaddrinfo()](https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfo) returns errors with naming convention EAI_. While these have the same values as WSA errors (they are simply aliases) on Windows, on Linux the EAI_* don't always map to the `errno` values. So, I had to create another mapping function that converts EAI to `errno` codes.
While doing so, I also made sure the return values of `getaddrinfo` are not fed directly into `iio_ptr()`.

With these changes, no errors outside of [-4095, -1] should emerge.

rgetz (approved):
looks good to me - thanks

cristina-suteu (approved):

PR #1360, opened 2025-10-14
CMake: Bring back lost NO_THREADS definition

dNechita:
During the latest cmake options refactoring the NO_THREADS cmake option should have been downgraded from option to a simple definition but instead it was removed entirely.
Bring it back and add NO_THREADS definition when building TINYIIOD.

## PR Type

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -626,6 +626,12 @@ else ()
 	target_sources(iio PRIVATE lock.c)
 endif()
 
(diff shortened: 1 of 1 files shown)


dNechita:
> doesn't this point out there are some missing CI tests?

I would like to avoid things like this in the future. I am considering some tests that check what is being set in the CMakeCache.txt against what is actually expected. That's one idea.

PR #1362, opened 2025-10-17
Don't exist when grep command returns non zero

dNechita:
## PR Description
It doesn't make sense to exit script when grep fails (returns 2) or when no lines were selected (returns 1).

## PR Type

## PR Checklist


rgetz:
best way to do this is just with a `|| true`

then you don't need to add a set +/- e

PR #1365, opened 2025-10-28
Added iio_buffer_release() which enqueues the last dequeued buffer back to the kernel

pfeatherstone:
## PR Description

Added the `iio_buffer_release()` function which enqueues the last dequeued buffer back to the kernel.
This is only applicable to the local backend and when using the MMAP API.
The implementation of this function in the local backend is essentially copy-pasted from the first part of `local_get_buffer()`.
The use of this function is totally optional and indeed only valid for local and MMAP.
The motivation for this function is when manually calling `poll()` or `select()` on the file descriptor returned by `iio_buffer_get_poll_fd()`. If there is only 1 kernel buffer and it hasn't yet been enqueued back to the kernel then a call to poll() will hang. This is where this function comes in handy. If you first call `iio_buffer_release()` then call `poll()` it will work.

## PR Type

## PR Checklist

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -120,6 +120,16 @@ int iio_buffer_set_blocking_mode(struct iio_buffer *buffer, bool blocking)
 	return iio_device_set_blocking_mode(buffer->dev, blocking);
 }
 
+ssize_t iio_buffer_release(struct iio_buffer *buffer)
+{
+	const struct iio_device *dev = buffer->dev;
+	ssize_t ret = 0;
+	if (dev->ctx->ops->release_buffer) {
+		ret = dev->ctx->ops->release_buffer(dev, &buffer->buffer, buffer->length);
+	}
+	return ret;
+}
+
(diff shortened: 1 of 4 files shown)


buffer.c
```diff
@@ -120,6 +120,16 @@ int iio_buffer_set_blocking_mode(struct iio_buffer *buffer, bool blocking)
 	return iio_device_set_blocking_mode(buffer->dev, blocking);
 }
 
+ssize_t iio_buffer_release(struct iio_buffer *buffer)
+{
+	const struct iio_device *dev = buffer->dev;
+	ssize_t ret = 0;
```
nunojsa, line 126: new line

nunojsa (commented):
Ok, I provided some feedback but I'm not really sure we want to have this. libiiov0 is pretty much in maintenance mode as we move to libiio v1 so I think we would need a very compelling reason to merge this.

So I'm afraid you might have to workaround it in your usecase. You could directly do the enqueue `ioctl()` or just have more than one buffer?

PR #1370, opened 2025-11-21
Add atomic register access for all backends

### LINKED ISSUE #1351: Enable atomic register access support across all backends
PR #1348 achieves this but only for local backend and targets libiio v0.

dNechita:
### Local Backend
First patch introduces thread-safe register read/write operations for the local backend using `direct_reg_access` via debugfs. It adds two new backend ops:
- `reg_read()`
- `reg_write()`

These functions use file locking (`flock`) to ensure atomic access and prevent race conditions when multiple threads interact with the same register interface.

The fallback logic in `iio_device_reg_read()` and `iio_device_reg_write()` is preserved for backends that do not implement these atomic operations.

This improves reliability for multi-threaded applications using libiio with the local backend.

### Network, USB and Serial Backends
Second patch implement atomic register operations which ensures that a register read or a register write cannot be interrupted by other clients of IIOD.
Advantages:
 - half round trips: single command vs write-address + read-value
 - binary addresses/values vs hex string conversion

On client-server compatibility:
v1.0 clients can connect to v1.0+ servers (atomic ops used)
v1.0 clients connecting to v0.x servers fall back to attribute method

Closes #1351.

## PR Type

## PR Checklist


rgetz:
Don't you want to fix this at the same time?
```
     * NOTE: This has a race condition when multiple clients access
     * the same IIOD server, as the two-step process (write address,
     * then read value) is not atomic.
```

dNechita:
> Don't you want to fix this at the same time?
> 
> ```
>      * NOTE: This has a race condition when multiple clients access
>      * the same IIOD server, as the two-step process (write address,
>      * then read value) is not atomic.
> ```

Yes, I think this can be removed. The race condition won't happen on IIOD server anymore.
Thanks!

PR #1372, opened 2025-11-26
Enable no timeout via params

### LINKED ISSUE #1371: Logical conflict between libiio's "no timeout" handling and default timeout
[Here](https://github.com/analogdevicesinc/libiio/blob/44bb23abf90c91144c37d95a06193fa80a5a4b46/local.c#L225), the libiio meaning of "no timeout" (0) is translated into the Unix meaning of "no timeout" (-1), so that "no timeout" works correctly on Unix systems. Therefore, to tell libiio that I do not want a timeout (i.e., I want it to wait indefinitely), I must pass **0** as the `timeout_ms` parameter in the context parameters.

However, whenever I pass **0** as the `timeout_ms` parameter, libiio instead uses the [default value](https://github.com/analogdevicesinc/libiio/blob/44bb23abf90c91144c37d95a06193fa80a5a4b46/context.c#L488), which is **1000 ms**. Passing **-1** does not work either, because it gets zeroed out by the [get-timeout](https://github.com/analogdevicesinc/libiio/blob/44bb23abf90c91144c37d95a06193fa80a5a4b46/local.c#L234) logic.

=> As a result, it is logically impossible to tell libiio not to use a timeout.

dNechita:
## PR Description
Closes #1371.

## PR Type

## PR Checklist

### DIFF
--- a/network-unix.c
+++ b/network-unix.c
@@ -194,13 +194,19 @@ int do_select(int fd, unsigned int timeout)
 {
 	struct pollfd pfd;
 	int ret;
+	int poll_timeout;
 
 	pfd.fd = fd;
 	pfd.events = POLLOUT | POLLERR;
 	pfd.revents = 0;
 
+	if (!timeout)
+		poll_timeout = -1;
+	else
+		poll_timeout = (int)timeout;
+
 	do {
-		ret = poll(&pfd, 1, timeout);
+		ret = poll(&pfd, 1, poll_timeout);
 	} while (ret == -1 && errno == EINTR);
 
 	if (ret < 0)


utils/iio_common.c
```diff
@@ -331,6 +331,9 @@ struct iio_context * handle_common_opts(char * name, int argc,
 				goto err_fail;
 			}
 			params.timeout_ms = sanitize_clamp("timeout", optarg, 0, INT_MAX);
+			/* Wait forever if timeout is explicitly set to 0 */
+			if (!params.timeout_ms)
+				params.timeout_ms = -1;
```
nunojsa, line 336: But what happens if no `--timeout` option is given? If that defaullts to the backend default, then it makes sense. However, if it defaults to "wait forever", I think it would make sense to update the meaning of the argument to match the `params.timeout_ms` meaning.

network-unix.c
```diff
@@ -194,13 +194,19 @@ int do_select(int fd, unsigned int timeout)
 {
 	struct pollfd pfd;
 	int ret;
+	int poll_timeout;
 
 	pfd.fd = fd;
 	pfd.events = POLLOUT | POLLERR;
 	pfd.revents = 0;
 
+	if (!timeout)
```
nunojsa, line 203: But can we get a 0 value though? Looking at the code it seems we always call this with `NETWORK_TIMEOUT_MS`?

utils/iio_common.c
dNechita: If no timeout is given, then it defaults to backend default.

network-unix.c
dNechita: Yes, there is one case (that I know of) where we can get the 0 value and it's when creating a network context.
`do_select()` can be called by `do_connect()` (network.c) which is called by `create_socket()` (network.c) which can also be called by `network_create_context()` (network.c) which is called by `iio_create_context()` (context.c) via `ops->create()`.

rgetz:
@dNechita :
You changed the front-facing API without updating the backends to match. Maintaining consistency throughout the model would make the API much easier to use and understand. Moving to a standard convention, like Unix/POSIX APIs (`poll()`, `select()`) where negative values mean “wait indefinitely,” would be more intuitive and safer.

As it stands, this mismatch between front and back ends introduces inconsistencies and forces fragile casts whenever a signed int is treated as an unsigned value, increasing the risk of subtle bugs. Aligning the types end-to-end would prevent these issues and make the code safer and more maintainable.

Currently, the front end might expect negative values to indicate “no timeout” (wait indefinitely), but with an unsigned backend, this gets interpreted as UINT_MAX, resulting in a 24-day timeout — a behavior that’s easy to miss and extremely difficult to reproduce.

```
/include/iio/iio-backend.h:
struct iio_backend {
        unsigned int                    api_version;
        const char                      *name;
        const char                      *uri_prefix;
        const struct iio_backend_ops    *ops;
        unsigned int                    default_timeout_ms;
};

./include/iio/iiod-client.h:
struct iiod_client_ops {
        ssize_t (*write)(struct iiod_client_pdata *desc,
                         const char *src, size_t len, unsigned int timeout_ms);
        ssize_t (*read)(struct iiod_client_pdata *desc,
                        char *dst, size_t len, unsigned int timeout_ms);
        ssize_t (*read_line)(struct iiod_client_pdata *desc,
                             char *dst, size_t len, unsigned int timeout_ms);
        void (*cancel)(struct iiod_client_pdata *desc);
};


./include/iio/iio.h:
/** @brief Set a timeout for I/O operations
 * @param ctx A pointer to an iio_context structure
 * @param timeout_ms A positive integer representing the time in milliseconds
 * after which a timeout occurs. A value of 0 is used to specify that no
 * timeout should occur.
 * @return On success, 0 is returned
 * @return On error, a negative errno code is returned */
__api __check_ret int iio_context_set_timeout(
                struct iio_context *ctx, unsigned int timeout_ms);

./include/iio/iio-lock.h:
__api int iio_cond_wait(struct iio_cond *cond, struct iio_mutex *lock,
                        unsigned int timeout_ms);
__api int iio_task_sync(struct iio_task_token *token, unsigned int timeout_ms);
__api int iio_task_cancel_sync(struct iio_task_token *token, unsigned int timeout_ms);
```

rgetz:
And just to provide some specifics, these are all the places that need fixes/reviews/casts...


```
[  9%] Building C object CMakeFiles/iio.dir/context.c.o
/home/rgetz/github/libiio/context.c: In function ‘iio_context_set_timeout’:
/home/rgetz/github/libiio/context.c:391:34: warning: conversion to ‘int’ from ‘unsigned int’ may change the sign of the result [-Wsign-conversion]
  391 |         ctx->params.timeout_ms = timeout;
      |                                  ^~~~~~~
/home/rgetz/github/libiio/context.c: In function ‘iio_create_context’:
/home/rgetz/github/libiio/context.c:492:46: warning: conversion to ‘int’ from ‘unsigned int’ may change the sign of the result [-Wsign-conversion]
  492 |                         params2.timeout_ms = backend->default_timeout_ms;
      |                                              ^~~~~~~
[ 15%] Building C object CMakeFiles/iio.dir/scan.c.o
/home/rgetz/github/libiio/scan.c: In function ‘iio_scan’:
/home/rgetz/github/libiio/scan.c:121:46: warning: conversion to ‘int’ from ‘unsigned int’ may change the sign of the result [-Wsign-conversion]
  121 |                         params2.timeout_ms = backend->default_timeout_ms;
      |                                              ^~~~~~~
[ 23%] Building C object CMakeFiles/iio.dir/local.c.o
/home/rgetz/github/libiio/local.c: In function ‘buffer_check_ready’:
/home/rgetz/github/libiio/local.c:256:38: warning: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Wsign-conversion]
  256 |         unsigned int rw_timeout_ms = pdata->dev->ctx->params.timeout_ms;
      |                                      ^~~~~
[ 29%] Building C object CMakeFiles/iio.dir/network.c.o
/home/rgetz/github/libiio/network.c: In function ‘network_create_context’:
/home/rgetz/github/libiio/network.c:712:39: warning: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Wsign-conversion]
  712 |         fd = create_socket(res, params->timeout_ms);
      |                                 ~~~~~~^~~~~~~~~~~~
[ 31%] Building C object CMakeFiles/iio.dir/dns_sd.c.o
/home/rgetz/github/libiio/dns_sd.c: In function ‘port_knock_discovery_data’:
/home/rgetz/github/libiio/dns_sd.c:170:30: warning: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Wsign-conversion]
  170 |                 timeout_ms = params->timeout_ms;
      |                              ^~~~~~
[ 36%] Building C object CMakeFiles/iio.dir/iiod-client.c.o
/home/rgetz/github/libiio/iiod-client.c: In function ‘iiod_client_read_integer’:
/home/rgetz/github/libiio/iiod-client.c:99:53: warning: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Wsign-conversion]
   99 |         unsigned int i, nb, first = 0, timeout_ms = client->params->timeout_ms;
      |                                                     ^~~~~~
/home/rgetz/github/libiio/iiod-client.c: In function ‘iiod_client_write_all’:
/home/rgetz/github/libiio/iiod-client.c:162:50: warning: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Wsign-conversion]
  162 |         unsigned int remaining = 0, timeout_ms = client->params->timeout_ms;
      |                                                  ^~~~~~
/home/rgetz/github/libiio/iiod-client.c: In function ‘iiod_client_read_all’:
/home/rgetz/github/libiio/iiod-client.c:218:50: warning: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Wsign-conversion]
  218 |         unsigned int remaining = 0, timeout_ms = client->params->timeout_ms;
      |                                                  ^~~~~~
/home/rgetz/github/libiio/iiod-client.c: In function ‘iiod_client_new’:
/home/rgetz/github/libiio/iiod-client.c:288:53: warning: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Wsign-conversion]
  288 |         err = iiod_client_set_timeout(client, params->timeout_ms);
      |                                               ~~~~~~^~~~~~~~~~~~
/home/rgetz/github/libiio/iiod-client.c: In function ‘iiod_client_set_timeout’:
/home/rgetz/github/libiio/iiod-client.c:470:28: warning: conversion to ‘int32_t’ {aka ‘int’} from ‘unsigned int’ may change the sign of the result [-Wsign-conversion]
  470 |                 cmd.code = remote_timeout;
      |                            ^~~~~~
 [ 42%] Building C object utils/CMakeFiles/iio_tests_helper.dir/iio_common.c.o
/home/rgetz/github/libiio/utils/iio_common.c: In function ‘handle_common_opts’:
/home/rgetz/github/libiio/utils/iio_common.c:333:45: warning: conversion from ‘long unsigned int’ to ‘int’ may change value [-Wconversion]
  333 |                         params.timeout_ms = sanitize_clamp("timeout", optarg, 0, INT_MAX);
      |                                             ^~~~~~~~~~~~~~
[ 87%] Building C object iiod/CMakeFiles/iiod.dir/responder.c.o
/home/rgetz/github/libiio/iiod/responder.c: In function ‘handle_timeout’:
/home/rgetz/github/libiio/iiod/responder.c:117:54: warning: conversion to ‘unsigned int’ from ‘int32_t’ {aka ‘int’} may change the sign of the result [-Wsign-conversion]
  117 |         ret = iio_context_set_timeout(pdata->ctx, cmd->code);
      |                                                   ~~~^~~~~~
```

rgetz:
Dan: the way to "fix" it - is to "finish" it - to go through the entire tree (based on the warnings I provided) and make everything signed, and make sure negative is handled properly...

I don't think there is any other possibility that will be correct.

PR #1377, opened 2025-12-18
Device event attributes

dNechita:
## PR Description
See discussion https://github.com/analogdevicesinc/libiio/pull/1368#issuecomment-3538607695.

Things left to do in this PR:

- [x] Update C#, C++ and Python bindings

## PR Type

## PR Checklist


dNechita:
> I’m not convinced that the right abstraction here is something that simply exposes an interface where the user must poll for events. (it's better than what exists today, but still doesn't match what the normal use case is).
> 
> In the Linux IIO subsystem, events are conceptually hardware-driven notifications (threshold crossings, data ready, buffer watermarks, etc.), not just another value to periodically check. For example, the kernel exposes event channels that deliver timestamped events through a character device once they are enabled via sysfs.
> 
> Typical workflow (per the sysfs-bus-iio ABI):
> 
> * Enable the event through sysfs `/sys/bus/iio/devices/iio:deviceX/events/<event>_en`
> * Wait for the event on the device node (`/dev/iio:deviceX`)
> * Read the event structure once the file descriptor signals readiness.
> 
> This is inherently event-driven, not just attribute polling.
> 
> IMHO - what this means is we need a push model (`iiod` needs to push something to the connected clients) - not a pull model (where clients continually poll).

The push mode you are suggesting here would be a new feature and I agree that it would be a useful one. But I would create a new issue that requests such a feature, as this PR is not about the event propagation but about parsing iio attributes that happen to relate to events. To me, the below attributes indicate that a user could write values to them in order to configure how the events behave. E.g. how often some conditions to be checked (sampling_frequency) or what's the threshold above which an event should be issued (thresh_rising_value), etc.
├── events
│   ├── sampling_frequency
│   ├── sampling_frequency_available
│   ├── thresh_either_en
│   ├── thresh_falling_hysteresis
│   ├── thresh_falling_value
│   ├── thresh_rising_hysteresis
│   └── thresh_rising_value

Without the changes in this PR, the above attributes will be wrongly parsed.
These new attributes that will be exposed ar not for polling.

nunojsa:
So, conceptually I'm fine with the approach! We need to be able to enable events, hysteresis and so on! For the push stuff, conceptually it should have it's similarities with buffering! FWIW, I would assume we already support through:

https://github.com/analogdevicesinc/libiio/blob/main/events.c

Have you looked into the above? They need to properly integrate with each other

dNechita:
> So, conceptually I'm fine with the approach! We need to be able to enable events, hysteresis and so on! For the push stuff, conceptually it should have it's similarities with buffering! FWIW, I would assume we already support through:

It does have it's similarities with buffering but it's a polling strategy. Client code requests buffers at a time of its choosing. Same goes for the current implementation of events: client code creates a stream-event object and reads events whenever it sees fits. Events aren't lost as they pile up until they are read.

I see the push strategy as a way for client code to register a callback that would be executed at the time the event occurs. This is not implemented. Implementing this, would add quite a bit of complexity in the client - server architecture. As things are right now, IIOD always listens for client requests and only responds when asked. It never initiates anything. If we want the IIOD to take action when an event occurs, we need to setup a communication channel where the client code always listen for server notifications. This channel would exist simultaneously with the original communication channel.
> 
> https://github.com/analogdevicesinc/libiio/blob/main/events.c
> 
> Have you looked into the above? They need to properly integrate with each other

nunojsa:
> It does have it's similarities with buffering but it's a polling strategy. Client code requests buffers at a time of its choosing. Same goes for the current implementation of events: client code creates a stream-event object and reads events whenever it sees fits. Events aren't lost as they pile up until they are read.

And the above makes sense to me. If I'm not mistaken, events are very similar to buffers in the way they are implemented. Events are just pushed into a kernel fifo that you can read on /dev/iio....

> I see the push strategy as a way for client code to register a callback that would be executed at the time the event occurs. This is not implemented

Agreed this would be too much for now. You would need more thread handling (also on the client side) for taking care of the async nature of the above. For now I wouldn't complicate things. An app can always build that on top of libbio.

My 2 cents

dNechita:
V1:

- Amend the commit with the utils where I updated the code generation that is part of iio_attr to also handle the events attributes. In the same commit I fixed the iio_attr not showing event attributes in verbose mode.

emu.c
```diff
@@ -732,6 +739,9 @@ static int check_available(const struct iio_attr *attr, const char *src)
 	case IIO_ATTR_TYPE_DEBUG:
 		avail_attr = iio_device_find_debug_attr(attr->iio.dev, avail_name);
 		break;
+	case IIO_ATTR_TYPE_EVENT:
+		avail_attr = iio_device_find_event_attr(attr->iio.dev, avail_name);
+		break;
```
nunojsa, line 1060: Is the emu change really needed for this patch?

local.c
```diff
@@ -1253,8 +1283,13 @@ static int add_attr_or_channel(void *d, const char *path)
...
 static int add_event(void *d, const char *path)
 {
-	return add_attr_or_channel_helper((struct iio_device *) d,
-				path, "events/", false);
+	struct iio_device *dev = (struct iio_device *) d;
+	const char *name = strrchr(path, '/') + 1;
+
+	if (is_event_attr(dev, name))
+		return iio_device_add_attr(dev, name, IIO_ATTR_TYPE_EVENT);
+	else
```
nunojsa: redundant else. 

But more important, this also looks a bit weird and misleading. Aren't all events attributes under events/? So `is_event_attr()` is a bit misleading given that the else has "events/".

I'm also not sure how we should handle this! As far as I can tell, static events attrs go into the device list and dynamic ones (which are defined as part of the channel definition) are part of the channel? I honestly think this is complicating things a bit but I see we already have this logic very deepen in that direction (same happens with scan_elements). But maybe there's room to improve things a bit. How will things look like if we do iio_info? We'll only see the "event attributes" for "device" kind of attributes right? The channel ones will come as "normal" channel attributes? 

So maybe we could introduce  the notion of event attributes for channels given that a channel can in  fact have an array of events:

https://elixir.bootlin.com/linux/v7.0.1/source/include/linux/iio/iio.h#L235

It could be even be a list. Then we could dump (and mark) channel event attributes when dumping channels. I think it would make things a bit more clear. But being objective, like this should work the same.

But up to you. Now that I think about it, it would also make sense to have APIs like iio_channel_read/write_event_attr(). Again, it would make things more clear about what is going on (even more given the fact these attrs can be enabled to be latter on buffered) but it would also complicate things a bit I guess.

local.c
```diff
@@ -1245,6 +1255,26 @@ static int add_attr_or_channel_helper(struct iio_device *dev,
...
+
+		/* Match patterns like "in_voltage0_" or "in_temp0_" */
+		if (strncmp(name, "in_", 3) == 0 || strncmp(name, "out_", 4) == 0) {
+			if (strstr(name, id) == name + 3 || strstr(name, id) == name + 4)
+				return false; /* It's a channel event attribute */
+		}
+	}
+
+	/* If no channel prefix found, it's a device-level event attribute */
+	return true;
```
nunojsa, line 1286: I think the API is a bit over commented

iiod/lexer.l
```diff
@@ -121,6 +121,11 @@ WORD (([[:alpha:]]+,)|(iio:))?(-|_|\.|[[:alnum:]])+
 	return BUFFER_ATTR;
 }
 
+<WANT_CHN_OR_ATTR>EVENT|event {
+	BEGIN(WANT_ATTR);
+	return EVENT_ATTR;
+}
```
nunojsa: I wonder if we need to care about this? This will be a libiio v1 only feature so do we care about the legacy protocol 😅?

nunojsa (commented):
Here it goes my 2 cents

gastmaier:
for the future new feature of not pooling, we would need, for different backends, having something similar to a (web)socket, where the connection is open and data is pushed to the host. the benefits are unclear.

I will test this pr tomorrow on hw

emu.c
dNechita: It can be in a different patch series to make it easier to review the core stuff first.

local.c
dNechita: agree

local.c
dNechita: You are right, channels will have attributes of type event and libiio should handle them.
I am thinking that adding `iio_channel_read/write_event_attr()` to the API is a good approach.

dNechita:
> I will test this pr tomorrow on hw

Thanks, your test results would be of great help.

Note that I will make additional changes to this PR to address the channel event attributes.

iiod/lexer.l
dNechita: In a realistic scenario this won't be needed since:
1. v1 client + v1 server use the binary protocol
2. v0 client + v1 server use ASCII protocol and v0 doesn't use have 'event attributes'

but
1. There is a very small chance that, during client–server negotiation, the client’s request to https://github.com/analogdevicesinc/libiio/blob/87f1497246ca1c573336fe8bce9cff5ac3c732da/iiod-client.c#L937 fails (for example, due to a sudden network drop causing a timeout). In that case, the connection falls back to using ASCII.
2. Alternatively, someone may be manually debugging via telnet or nc and sending READ device EVENT attr\r\n. This works because IIOD starts in ASCII mode by default and switches to binary mode only upon a client request.

I would keep it for now.

nunojsa:
> for the future new feature of not pooling, we would need, for different backends, having something similar to a (web)socket, where the connection is open and data is pushed to the host. the benefits are unclear.

I don't think we have pooling now in the "real" sense of the word. We just do `pool()` waiting for events. So it's kind of synchronous mechanism. An app will sleep/block waiting for events. The very same thing we have with buffers (unless I'm missing something). Of course that, one sees events and maybe the expectation is to get notified in an async way. But, honestly, I would defer that to the app rather than building such complexity in the library. At least for now...

emu.c
dNechita: was moved into a separate commit

iiod/lexer.l
dNechita: I took another look at it, and my arguments didn’t seem very compelling. I removed it.

dNechita:
V2:

- Introduced event-type channel attributes, previously only regular types existed.
- removed event attributes from the old ASCII parser
- updated emu backend in a separate commit 
- rebased on main

context.c
```diff
@@ -21,13 +21,14 @@ static const char xml_header[] = "<?xml version=\"1.0\" encoding=\"utf-8\"?>"
 "<!DOCTYPE context ["
 "<!ELEMENT context (device | context-attribute)*>"
 "<!ELEMENT context-attribute EMPTY>"
-"<!ELEMENT device (channel | buffer | attribute | debug-attribute | buffer-attribute)*>"
+"<!ELEMENT device (channel | buffer | attribute | debug-attribute | buffer-attribute | event-attribute)*>"
 "<!ELEMENT channel (scan-element?, attribute*)>"
```
999Marius: Channel element should also include `event-attribute*` if not any XML with channel event attributes will fail when being parsed

context.c
```diff
@@ -38,6 +39,7 @@ static const char xml_header[] = "<?xml version=\"1.0\" encoding=\"utf-8\"?>"
 "<!ATTLIST attribute name CDATA #REQUIRED filename CDATA #IMPLIED value CDATA #IMPLIED>"
 "<!ATTLIST debug-attribute name CDATA #REQUIRED value CDATA #IMPLIED>"
 "<!ATTLIST buffer-attribute name CDATA #REQUIRED value CDATA #IMPLIED>"
+"<!ATTLIST event-attribute name CDATA #REQUIRED value CDATA #IMPLIED>"
```
999Marius: Here It seems we miss filename attribute `filename CDATA #IMPLIED` since `iio_snprintf_chan_event_attr_xml` adds a filename attribute when present

device.c
```diff
@@ -20,6 +20,7 @@ static const char * const xml_attr_prefix[] = {
 	"",
 	"debug-",
 	"buffer-",
+	"event-",
```
999Marius, line 22: `"event-"`  should be between `buffer` and `debug`  since  `xml_attr_prefix[attr->type]` here would index by attr->type wich 
```
enum iio_attr_type {
	IIO_ATTR_TYPE_DEVICE = 0,
	IIO_ATTR_TYPE_DEBUG,
	IIO_ATTR_TYPE_EVENT,
	IIO_ATTR_TYPE_BUFFER,
	IIO_ATTR_TYPE_CHANNEL,
	IIO_ATTR_TYPE_CONTEXT,
};
```
and would put the prefix `buffer-` for `event-` and reverse by indexing with this enum

gastmaier (commented):
iio_info grouping works properly

```
root@analog-vck190:/sys/bus/iio/devices/iio:device0# echo 1000000 > events/sampling_frequency
root@analog-vck190:/sys/bus/iio/devices/iio:device0# echo 10 > events/thresh_falling_hysteresis
root@analog-vck190:/sys/bus/iio/devices/iio:device0# echo 5 > events/thresh_rising_value
root@analog-vck190:/sys/bus/iio/devices/iio:device0# iio_info
WARNING: Unable to read INI file: -5
iio_info version: 1.0 (git tag:af968f1)
Libiio version: 1.0 (git tag: af968f1) backends: local ip usb xml
IIO context created with local backend.
Backend version: 1.0 (git tag: af968f1)
Backend description string: Linux analog-vck190 6.12.0+ #9 SMP PREEMPT Mon May 11 19:23:09 CEST 2026 armv7l
IIO context has 2 attributes:
        attr  0: local,kernel value: 6.12.0+
        attr  1: uri value: local:
IIO context has 1 devices:
        iio:device0: ad4050 (buffer capable)
                1 channels found:
                        voltage0:  (input, index: 0, format: le:s12/32>>0)
                                attr  0: calibscale (in_voltage_calibscale) value: 1.000000000
                                attr  1: oversampling_ratio (in_voltage_oversampling_ratio) value: 1
                                attr  2: oversampling_ratio_available (in_voltage_oversampling_ratio_available) options: 1 2 4 8 16 32 64 128 256
                                attr  3: raw (in_voltage_raw) value: 0
                                attr  4: sampling_frequency (in_voltage_sampling_frequency) value: 2000000
                                attr  5: scale (in_voltage_scale) value: 2.441406250
                2 buffer attributes found:
                        attr  0: data_available value: 0
                        attr  1: length_align_bytes value: 8
                1 debug attributes found:
                        attr  0: direct_reg_access
                7 device-specific event attributes found:
                        attr  0: sampling_frequency value: 1000000
                        attr  1: sampling_frequency_available options: 2000000 1000000 300000 100000 33300 10000 3000 500 333 250 200 166 140 124 111
                        attr  2: thresh_either_en value: 0
                        attr  3: thresh_falling_hysteresis value: 10
                        attr  4: thresh_falling_value value: 0
                        attr  5: thresh_rising_hysteresis value: 0
                        attr  6: thresh_rising_value value: 5
```
iio_attr missing, with https://github.com/analogdevicesinc/libiio/tree/device-event-attributes-iio-attr

```
iio_attr -d ad4050
iio_attr: Found ad4050 device, but it has 0 device attributes
dev 'ad4050', device event attr 'sampling_frequency', value '1000000'
dev 'ad4050', device event attr 'sampling_frequency_available', options '2000000' '1000000' '300000' '100000' '33300' '10000' '3000' '500' '333' '250' '200' '166' '140' '124' '111'
dev 'ad4050', device event attr 'thresh_either_en', value '0'
dev 'ad4050', device event attr 'thresh_falling_hysteresis', value '10'
dev 'ad4050', device event attr 'thresh_falling_value', value '0'
dev 'ad4050', device event attr 'thresh_rising_hysteresis', value '0'
dev 'ad4050', device event attr 'thresh_rising_value', value '5'
```

device.c
gastmaier: Confirming they are indeed swapped
we can benefit from verbosity here
```
+ [IIO_ATTR_TYPE_DEVICE] = "",
+ [IIO_ATTR_TYPE_DEBUG]  = "debug-",
+ [IIO_ATTR_TYPE_EVENT]  = "event-",
+ [IIO_ATTR_TYPE_BUFFER] = "buffer-",
   ```

**Comment by dNechita** [comment]
V3:

- reworked the commits as if both device and channel event attributes were added from the start
- added `-E` and `-e` flags to the `iio_attr` utility to handle device and channel attributes consistently with regular attributes
- rebased onto main, which includes recent multi-buffer changes, and resolved conflicts.

**Comment by dNechita** [comment]
V4:

- The initial approach introduced a new attribute type, `IIO_ATTR_TYPE_EVENT`. However, this came with limitations, as there was no clear way to distinguish between channel event attributes and device event attributes. As a result, workarounds were introduced, such as relying on the attribute’s filename structure and checking for an "events" suffix. o address this, I introduced `IIO_ATTR_TYPE_DEVICE_EVENT` and `IIO_ATTR_TYPE_CHANNEL_EVENT`, which resolve the issue and provide a much cleaner separation.
- The above change impacted other parts of the implementation, but it did not introduce any functional differences. The utils and bindings remain unchanged.
- Updated the API tests to cover the newly introduced API functions

**Comment by dNechita** [comment]
@gastmaier @nunojsa @999Marius I am ready to merge this. It would be great if you could take another look at this.

**Comment by dNechita** [comment]
V5:

- fix in `iio_device_add_attr()`: buffer-type attributes are now rejected, preventing out-of-bounds access to `dev->attrlist[3]` when the maximum size is 2.

**Review by 999Marius — APPROVED** [review]
LGTM

**Review by gastmaier — COMMENTED** [review]
Tested per previous comment, values are correctly set:

```
$ iio_info
IIO context has 2 attributes:
        attr  0: local,kernel value: 6.12.0+
        attr  1: uri value: local:
IIO context has 1 devices:
        iio:device0: ad4050 (buffer capable)
                1 channels found:
                        voltage0:  (input, index: 0, format: le:s14/32>>0)
                                attr  0: calibscale (in_voltage_calibscale) value: 1.000000000
                                attr  1: oversampling_ratio (in_voltage_oversampling_ratio) value: 4
                                attr  2: oversampling_ratio_available (in_voltage_oversampling_ratio_available) options: 1 2 4 8 16 32 64 128 256
                                attr  3: raw (in_voltage_raw) value: -516
                                attr  4: sampling_frequency (in_voltage_sampling_frequency) value: 2000000
                                attr  5: scale (in_voltage_scale) value: 0.610351562
                1 buffer found:
                        buffer0:  (input)
                        1 scan elements found:
                                Scan element 0: voltage0 (input)
                        2 buffer attributes found:
                                attr  0: data_available value: 1600
                                attr  1: length_align_bytes value: 8
                7 event attributes found:
                        attr  0: sampling_frequency value: 1000000
                        attr  1: sampling_frequency_available options: 2000000 1000000 300000 100000 33300 10000 3000 500 333 250 200 166 140 124 111
                        attr  2: thresh_either_en value: 0
                        attr  3: thresh_falling_hysteresis value: 10
                        attr  4: thresh_falling_value value: 0
                        attr  5: thresh_rising_hysteresis value: 0
                        attr  6: thresh_rising_value value: 5
                1 debug attributes found:
                        attr  0: direct_reg_access
                No trigger on this device
```

dNechita:
Awesome. Thanks for testing this!

PR #1381, opened 2025-12-26
ci: ensure proper build type for releases and nightlies

rgetz:
Currently, Azure pipelines always built with the default CMake type, causing tagged releases to use RelWithDebInfo instead of Release. This could result in slower binaries or debug artifacts being shipped inadvertently.

- Pass CMAKE_BUILD_TYPE from azure-pipelines.yml to CI shell and PowerShell scripts based on build from tag or main
- Default to RelWithDebInfo for nightlies and Release for tagged releases
- Update Linux, Windows MinGW, and MSVC scripts to respect the build type

Tagged releases are built with Release, whereas nightlies default to RelWithDebInfo. Although both are optimized and define NDEBUG, some compilers apply -O3 for Release and -O2 for RelWithDebInfo. As a result, final performance characteristics may differ slightly from nightly artifacts. Pre-release validation on tagged builds is recommended for performance-critical or timing-sensitive deployments.

Note: Users can still override the build type manually by passing -DCMAKE_BUILD_TYPE=<type>
when running cmake, as usual.


CI/azure/ci-ubuntu.sh
```diff
@@ -3,7 +3,7 @@ set -x
 uname -a
 echo "$PWD"
 mkdir build && cd build
-cmake .. -Werror=dev -DCOMPILE_WARNING_AS_ERROR=ON -DWITH_SERIAL_BACKEND=ON -DWITH_EXAMPLES=ON -DPYTHON_BINDINGS=ON -DCPP_BINDINGS=ON -DENABLE_PACKAGING=ON -DCPACK_SYSTEM_NAME="${ARTIFACTNAME}"
+cmake .. -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE:-RelWithDebInfo}" -Werror=dev -DCOMPILE_WARNING_AS_ERROR=ON -DWITH_SERIAL_BACKEND=ON -DWITH_EXAMPLES=ON -DPYTHON_BINDINGS=ON -DCPP_BINDINGS=ON -DENABLE_PACKAGING=ON -DCPACK_SYSTEM_NAME="${ARTIFACTNAME}"
```
dNechita, line 6: Shouldn't it default to 'Release'?

rgetz: This is really only a backup - it is set by the top level azure file.

Release removes all the asserts, which IMHO - is a bad idea for a default build.

dNechita (approved):

PR #1391, opened 2026-01-29
Staging/consistent timeout convention

dNechita:
## PR Description

This PR makes the timeout values consistent across libiio.

Before, there were 2 conventions:
Using `iio_context_set_timeout()`:
 -  0 value meant no wait
 - Greater than 0 value meant wait that amount of milliseconds
 - no option to wait forever!

Using `struct_iio_context_params`:
 - Zero: use backend default timeout
 - Negative values: no timeout (wait indefinitely)
 - Positive values: timeout after specified milliseconds

With changes in this PR:
For both `iio_context_set_timeout()` and `struct_iio_context_params the following applies:
 - Zero: use backend default timeout
 - Negative values: no timeout (wait indefinitely)
 - Positive values: timeout after specified milliseconds

To do:
- [ ] Handle how this affects the backward compatibility with v0

## PR Type

## PR Checklist

### DIFF
--- a/include/iio/iio.h
+++ b/include/iio/iio.h
@@ -759,11 +759,11 @@ __api __check_ret __pure struct iio_device * iio_context_find_device(
 /** @brief Set a timeout for I/O operations
  * @param ctx A pointer to an iio_context structure
  * @param timeout_ms Timeout value in milliseconds:
- * - Negative values (typically -1): infinite timeout (wait forever)
+ * - Negative values: infinite timeout (wait forever)
  * - Zero: use backend default timeout
  * - Positive values: explicit timeout in milliseconds
  * @return On success, 0 is returned
- * @return On error, a negative errno code is returned (e.g., -EINVAL for timeout < -1) */
+ * @return On error, a negative errno code is returned */
 __api __check_ret int iio_context_set_timeout(
 		struct iio_context *ctx, int timeout_ms);
 
--- a/include/iio/iio.h
+++ b/include/iio/iio.h
@@ -758,13 +758,14 @@ __api __check_ret __pure struct iio_device * iio_context_find_device(
 
 /** @brief Set a timeout for I/O operations
  * @param ctx A pointer to an iio_context structure
- * @param timeout_ms A positive integer representing the time in milliseconds
- * after which a timeout occurs. A value of 0 is used to specify that no
- * timeout should occur.
+ * @param timeout_ms Timeout value in milliseconds:
+ * - Negative values (typically -1): infinite timeout (wait forever)
+ * - Zero: use backend default timeout
+ * - Positive values: explicit timeout in milliseconds
  * @return On success, 0 is returned
- * @return On error, a negative errno code is returned */
+ * @return On error, a negative errno code is returned (e.g., -EINVAL for timeout < -1) */
 __api __check_ret int iio_context_set_timeout(
-		struct iio_context *ctx, unsigned int timeout_ms);
+		struct iio_context *ctx, int timeout_ms);
 
 
 /** @brief Get a pointer to the params structure
--- a/local.c
+++ b/local.c
@@ -217,12 +217,12 @@ static int set_channel_name(struct iio_channel *chn)
  * before each poll() invocation the timeout is recalculated relative to the
  * start of refill() or push() operation.
  */
-static int get_rel_timeout_ms(struct timespec *start, unsigned int timeout_rel)
+static int get_rel_timeout_ms(struct timespec *start, int timeout_rel)
 {
 	struct timespec now;
 	int diff_ms;
 
-	if (timeout_rel == 0) /* No timeout */
+	if (timeout_rel <= 0) /* No timeout or infinite timeout */
 		return -1;
 
 	clock_gettime(CLOCK_MONOTONIC, &now);
@@ -230,14 +230,14 @@ static int get_rel_timeout_ms(struct timespec *start, unsigned int timeout_rel)
 	diff_ms = (now.tv_sec - start->tv_sec) * 1000;
 	diff_ms += (now.tv_nsec - start->tv_nsec) / 1000000;
 
-	if (diff_ms >= (int) timeout_rel) /* Expired */
+	if (diff_ms >= timeout_rel) /* Expired */
 		return 0;
 	if (diff_ms > 0) /* Should never be false, but lets be safe */
 		timeout_rel -= diff_ms;
 	if (timeout_rel > INT_MAX)
 		return INT_MAX;
 
-	return (int) timeout_rel;
+	return timeout_rel;
 }
 
 int buffer_check_ready(struct iio_buffer_pdata *pdata, int fd,
--- a/lock-windows.c
+++ b/lock-windows.c
@@ -76,14 +76,17 @@ void iio_cond_destroy(struct iio_cond *cond)
 }
 
 int iio_cond_wait(struct iio_cond *cond, struct iio_mutex *lock,
-		  unsigned int timeout_ms)
+		  int timeout_ms)
 {
 	BOOL ret;
+	DWORD win_timeout;
 
-	if (timeout_ms == 0)
-		timeout_ms = INFINITE;
+	if (timeout_ms <= 0)
+		win_timeout = INFINITE;
+	else
+		win_timeout = (DWORD)timeout_ms;
 
-	ret = SleepConditionVariableCS(&cond->cond, &lock->lock, timeout_ms);
+	ret = SleepConditionVariableCS(&cond->cond, &lock->lock, win_timeout);
 
 	return ret ? 0 : -ETIMEDOUT;
 }
--- a/lock.c
+++ b/lock.c
@@ -90,13 +90,13 @@ void iio_cond_destroy(struct iio_cond *cond)
 }
 
 int iio_cond_wait(struct iio_cond *cond, struct iio_mutex *lock,
-		  unsigned int timeout_ms)
+		  int timeout_ms)
 {
 	struct timespec ts;
 	uint64_t usec;
 	int ret = 0;
 
-	if (timeout_ms == 0) {
+	if (timeout_ms <= 0) {
 		pthread_cond_wait(&cond->cond, &lock->lock);
 	} else {
 		clock_gettime(CLOCK_REALTIME, &ts);
--- a/context.c
+++ b/context.c
@@ -375,7 +375,7 @@ const char * iio_context_get_version_tag(const struct iio_context *ctx)
 	return LIBIIO_VERSION_GIT;
 }
 
-int iio_context_set_timeout(struct iio_context *ctx, unsigned int timeout)
+int iio_context_set_timeout(struct iio_context *ctx, int timeout)
 {
 	int ret = 0;
 
@@ -487,10 +487,7 @@ struct iio_context * iio_create_context(const struct iio_context_params *params,
 		if (!params2.timeout_ms) {
 			/* Zero means use backend default */
 			params2.timeout_ms = backend->default_timeout_ms;
-		} else if (params2.timeout_ms < 0) {
-			/* Negative means infinite - translate to 0 for backends */
-			params2.timeout_ms = 0;
-		} /* Positive values pass through unchanged */
+		}
 
 		ctx = backend->ops->create(&params2,
 					   uri + strlen(backend->uri_prefix));
--- a/dns_sd.c
+++ b/dns_sd.c
@@ -161,10 +161,10 @@ void port_knock_discovery_data(const struct iio_context_params *params,
 			       struct dns_sd_discovery_data **ddata)
 {
 	struct dns_sd_discovery_data *d, *ndata;
-	unsigned int timeout_ms;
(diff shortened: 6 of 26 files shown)


local.c
```diff
@@ -217,27 +217,27 @@ static int set_channel_name(struct iio_channel *chn)
...
  * start of refill() or push() operation.
  */
-static int get_rel_timeout_ms(struct timespec *start, unsigned int timeout_rel)
+static int get_rel_timeout_ms(struct timespec *start, int timeout_rel)
 {
 	struct timespec now;
 	int diff_ms;
 
-	if (timeout_rel == 0) /* No timeout */
+	if (timeout_rel <= 0) /* No timeout or infinite timeout */
```
rgetz, line 225: is there a difference between "No timeout" and/or "infinite timeouts"? If not - just pick one.

Man pages for `poll` use:
>Specifying a negative value in timeout means an infinite timeout.

I would just stick with that.

include/iio/iio.h
```diff
@@ -759,11 +759,11 @@ __api __check_ret __pure struct iio_device * iio_context_find_device(
 /** @brief Set a timeout for I/O operations
  * @param ctx A pointer to an iio_context structure
  * @param timeout_ms Timeout value in milliseconds:
- * - Negative values (typically -1): infinite timeout (wait forever)
+ * - Negative values: infinite timeout (wait forever)
```
rgetz, line 762: does this mean you can only have a time out of >1 ms? (since 0 = backend default?, and all negatives are infinite timeouts? (1 -> 2,147,483,648ms or 24 days, 20 hours, 31 minutes, 23.648 seconds).

There is no way to have no timeout?

Again, `poll` indicates:
>Specifying a timeout of zero causes poll() to  return immediately, even if no file descriptors are ready.

lock-windows.c
```diff
@@ -76,14 +76,17 @@ void iio_cond_destroy(struct iio_cond *cond)
...
 int iio_cond_wait(struct iio_cond *cond, struct iio_mutex *lock,
-		  unsigned int timeout_ms)
+		  int timeout_ms)
 {
 	BOOL ret;
+	DWORD win_timeout;
 
-	if (timeout_ms == 0)
-		timeout_ms = INFINITE;
+	if (timeout_ms <= 0)
```
rgetz, line 84: it's no longer <=, it's only < isn't it?

lock.c
```diff
@@ -90,13 +90,13 @@ void iio_cond_destroy(struct iio_cond *cond)
...
 int iio_cond_wait(struct iio_cond *cond, struct iio_mutex *lock,
-		  unsigned int timeout_ms)
+		  int timeout_ms)
 {
 	struct timespec ts;
 	uint64_t usec;
 	int ret = 0;
 
-	if (timeout_ms == 0) {
+	if (timeout_ms <= 0) {
```
rgetz, line 99: same here - in theory - it should never be zero, since zero will always get replaced by the backend?

rgetz:
And I think you missed this?
```
./utils/iio_common.c:333:24: warning: implicit conversion loses integer precision: 'unsigned long' to 'int' [-Wshorten-64-to-32]
  333 |                         params.timeout_ms = sanitize_clamp("timeout", optarg, 0, INT_MAX);
      |                                           ~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

timeouts should no longer be clamped to zero....

include/iio/iio.h
dNechita: Yes, there is no way to have no timeout. This is a downside that I shouldn't have treated so lightly. I prioritized keeping the timeout semantics consistent between `iio_context_set_timeout()` and `struct iio_context_params`.

Maybe it's better that for `iio_context_set_timeout`, a timeout value of:
- negative: infinite timeout (wait forever)
- zero: no timeout (return immediately)
- positive: a timeout between 1 and 2,147,483,648ms

And keep `struct iio_context_params` having the same timeout meaning:
- negative: infinite timeout
- zero: use backend default timeout
- positive: a timeout between 1 and 2,147,483,648ms

There won't be a way to set no timeout using `struct iio_context_params` (unless writing a custom backend where default timeout is explicitly set to 0) but I don't think there is any real use case, at least when it comes to creating an iio context (local and serial have it set to 1000ms, netwok and usb have it set to 5000ms). Once contxt is created, one can easily call `iio_context_set_timeout()` and set no timeout to be applied to future calls.

dNechita:
> And I think you missed this?
> 
> ```
> ./utils/iio_common.c:333:24: warning: implicit conversion loses integer precision: 'unsigned long' to 'int' [-Wshorten-64-to-32]
>   333 |                         params.timeout_ms = sanitize_clamp("timeout", optarg, 0, INT_MAX);
>       |                                           ~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> timeouts should no longer be clamped to zero....

Indeed, this needs to be updated. Thanks!

include/iio/iio.h
dNechita: Just did some more digging and it looks like even prior to this PR, there wasn't a way to have a timeout of 0 ms (return immediately). The following commit description states that the 0 value was used to wait forever:
https://github.com/analogdevicesinc/libiio/commit/4d0b516278e838874bad7ee9d0af3fb860b3d7b8
And it is being backed up by the actual implementation where for a timeout of 0, `get_rel_timeout_ms()` is returning -1 that is then passed to poll() which waits forever. See latest main branch: https://github.com/analogdevicesinc/libiio/blob/c971edc8545d6e4f9845bf4325d75e259936dd22/local.c#L225

local.c
dNechita: By "No timeout" I meant "no wait" so yes, it's different. I guess it will be better to user more clearer terms such as: "immediate timeout" or "do not wait".

lock-windows.c
dNechita: yes, you're right

include/iio/iio.h
rgetz: My main concern is just being explicit about the API contract and semantics.

Right now the behavior is a bit subtle because 0 and negative values mean different things depending on whether you’re using `iio_context_set_timeout()` vs `struct iio_context_params`. If we intentionally diverge, that increases the cognitive load for users, and I’m not sure what the concrete benefit is. On the other hand, diverging from the usual poll() semantics that many people are already familiar with is also a bit of a question mark to me.

It feels like the complexity comes from trying to encode four distinct semantics:
- infinite timeout (wait forever)
- no timeout (return immediately)
- a finite timeout (1 … 2^31 ms)
- use backend default timeout

into only three value classes:
- positive
- zero
- negative

Do we need a specific IIO_RETURN_IMMEDIATELY as valid timeout (mapped to `INT_MIN` (−2,147,483,648)) to cover this? and then translate that in the specific poll calls (to zero).

A practical IIO case where “no timeout / return immediately” can be useful is event-loop or GUI integrations. For example, an application that is already using `select()`/`poll()`/`epoll()` on multiple FDs may want to periodically call into libiio to check for data availability without blocking. In that scenario, an immediate return lets libiio act as a non-blocking probe rather than a wait primitive, and avoids needing a dedicated thread just for blocking IIO calls.

My goal isn’t to add surface area, just to make sure the semantics are unambiguous for someone reading the header without needing to inspect backend implementations.

local.c
rgetz: using the same terminology as others (poll man page) may help ensure everyone understands the purpose...

> timeout interval will be rounded up to the system clock
granularity, and kernel scheduling delays mean that the blocking  interval may overrun by a small amount.  Specifying a negative value in timeout  means  an  infinite  timeout.   Specifying a timeout of zero causes `poll()` to return immediately, even if no file descriptors are ready.

>Some  implementations  define  the  nonstandard constant `INFTIM` with the value -1 for use as a timeout for `poll()`.  This constant is not provided in glibc.

defining a few constants - and getting people to use those - might be the least worst idea.

include/iio/iio.h
rgetz: > Just did some more digging and it looks like even prior to this PR, there wasn't a way to have a timeout of 0 ms (return immediately).

Understand - but I think you would agree - there are use cases  - where you don't want things to block, having that would be useful...

If we are going to become more like standard functions that people are already familar with, it's going to be easier for people to understand how to use things.
 - [`poll()`](https://man7.org/linux/man-pages/man2/poll.2.html) takes an `int timeout`
- [`ppoll()`](https://man7.org/linux/man-pages/man2/poll.2.html) takes an `struct timespec *_Nullable tmo_p`
- [`select()`](https://man7.org/linux/man-pages/man2/select.2.html) takes `struct timeval *_Nullable  timeout`
- [`epoll_wait()`](https://man7.org/linux/man-pages/man2/epoll_wait.2.html) takes `int timeout`
- [`epoll_pwait2`](https://man7.org/linux/man-pages/man2/epoll_wait.2.html) takes `struct timespec *_Nullable timeout`

Most Linux I/O primitives support three core semantics natively:
- Infinite
- Immediate
- Finite

“Use backend default” is the extra fourth semantic that typical syscalls do not try to encode in the same parameter. That’s why “4 meanings in 3 values” feels wrong. Maybe it makes more sense to move to `struct timeval` and differentiate between `NULL` (default backed) and `{0, 0}` (return immediately) - something we can't do today with `int`. 

to me - moving to `struct timeval` would be the cleanest but would take a little more effort to be backwards compatible with v0.

The alternative (to be as clear to end users) is include a `IIO_TIMEOUT_INFINITE` (similar to the historic System V `INFTIM`) and `IIO_TIMEOUT_NONBLOCK` and `IIO_TIMEOUT_BACKEND` to distinguish, and set them to the numbers we want/know.

dNechita: > It feels like the complexity comes from trying to encode four distinct semantics:
> 
> * infinite timeout (wait forever)
> * no timeout (return immediately)
> * a finite timeout (1 … 2^31 ms)
> * use backend default timeout
> 
> into only three value classes:
> 
> * positive
> * zero
> * negative

Well said!
But there's another source of complexity and that is the value pattern in `struct iio_context_params`.
Looking at `struct iio_context_params`:
```
FILE *out;              // NULL = defaults to stdout
FILE *err;              // NULL = defaults to stderr
enum iio_log_level stderr_level;    // 0 = defaults to LEVEL_WARNING
enum iio_log_level timestamp_level; // 0 = defaults to LEVEL_DEBUG
int timeout_ms;         // 0 = ??? (the conflict)
```
The established pattern is: zero/NULL = "use default". But poll() semantics say: 0 = "return immediately". These are incompatible.

Treating timeout_ms = 0 as 'return immediately' could become a trap for users that don't care about filling up `struct iio_context_params` and might do this:
``
struct iio_context_params params = {0};  // timeout becomes non-blocking!
``

If timeout becomes non-blocking, backends like network, usb and serial won't work because it won't be enough time for context to be created.
I was trying not to break that pattern when I first implemented the changes in this PR.

rgetz: >The established pattern is: zero/NULL = "use default". But poll() semantics say: 0 = "return immediately". These are incompatible.

I think you mean - _the **previous** (v0) pattern is zero/NULL = "use default"_ We can fix things in v1.

And I get the backwards compatibility argument...

To me the questions are separate:
1. Do we want to have no timeout (return immediately)
2. If the answer to 1 is yes - how to encode it.

If and only if the answer to 1 is yes - what I suggested above:
 - `IIO_TIMEOUT_BACKEND` = `0`;
 - `IIO_TIMEOUT_INFINITE` = `-1` (similar to the historic System V `INFTIM`)
 - `IIO_TIMEOUT_NONBLOCK` = `INT_MIN` (−2,147,483,648)
to distinguish, and set them to the numbers we want/know. This keeps things mostly backwards compatible, and is most useful on local backends.

As for what the default timeouts (currently) are:
 - `usb.c` : `.default_timeout_ms = 5000,`
 - `network.c` : `#define NETWORK_TIMEOUT_MS 5000`
 - `serial.c` : `.default_timeout_ms = 1000,`
 - `local.c` : `.default_timeout_ms = 1000,`

All seems sort of arbitrary - and weird that the slowest (serial) and fastest (local) share the same value... ?

dNechita: > 1. Do we want to have no timeout (return immediately)

It isn't helpful in certain scenarios like creating usb/network/serial contexts but controlling the timeout would allow different libiio blocking operations to become non-blocking. I think having this kind of flexibility adds value. I can't find a certain function that would benefit from `return immediately` because the ones that needed it have been designed with a parameter. See `iio_event_stream_read()` and `iio_block_dequeue()`.

> If and only if the answer to 1 is yes - what I suggested above:
> 
> * `IIO_TIMEOUT_BACKEND` = `0`;
> * `IIO_TIMEOUT_INFINITE` = `-1` (similar to the historic System V `INFTIM`)
> * `IIO_TIMEOUT_NONBLOCK` = `INT_MIN` (−2,147,483,648)
>   to distinguish, and set them to the numbers we want/know. This keeps things mostly backwards compatible, and is most useful on local backends.

I think this would work.

PR #1392, opened 2026-02-04
[RFC] Buffer Refactor

nunojsa:
## PR Description

This is a RFC (request for comments) given that it's a bit disruptive and I want to make sure we're all in the same page before going all the way with this (converting all internal users and so on). The biggest motivation for this came from adding support for multibuffer where I realized buffers are not handled as I would expect. Given that now is the time to break API, I propose the following main two changes:

1) Creating buffers during context creation which naturally obsolutes  `iio_device_create_buffer()` ;
2) Add a new buffer_stream interface for buffers that are open and ready to stream.

For 1), it really does not make sense for one to have to call `iio_device_create_buffer()` in order to be able to access buffer attributes. Buffers are fundamentally the same as channels or any other device attribute. They are always there as soon as the IIO device is registered. Hence there's no real reason why buffers are not added the same way as channels or other attributes during device creation. Not to speak about all of the overhead of opening the IIO chardev when we only care about checking the buffer attributes (like in iio_info). Having buffers attributes in the device list and then copy them in `iio_device_create_buffer()` is also far from ideal. Moreover  `iio_device_create_buffer()` kind of takes a meaningless index as if users have any say about this index.

Having said the above, naturally from a libiio standpoint there are some attributes that we should not allow to handle if the buffer is not opened (like enable).

For 2), I was really not 100% sure how to proceed. The way I have it in the PR is what makes more sense to me but it is way more disruptive in terms of API. We have yet another handle (iio_buffer_stream) to represent a buffer ready for streaming which leads to way more changes and updates throughout the API.  An alternative I had in mind was to have a boolean flag directly in iio_buffer to track the open status. But to me, that is way more prone to errors since we have to add checks all over the place to ensure proper state (not to speak about concurrency). With the new API, the new handle ensures the state (the same way as iio_buffer does now).

I'm also going with lots of trouble to keep each commit compilable and bisectable. That is the reason why I'm keeping both APIs and making sure we still use "legacy" fields when switching to the new API. That way we can have incremental changes and change one user at a time (instead of needing a huge, harder to review, commit). But naturally the final patch in a proper PR should be to drop the legacy code.

Also note this PR is far from being complete

* Lots of users to convert (including compat.c)
* Update bindings
* Docs
* Some other stuff I might be missing...

So, don't go into much detail about reviewing this. The intent is more to give an overview on how the changes look like and if if we should  go down this route.

FWIW, we can implement multi buffer without all of the above [1](https://github.com/analogdevicesinc/libiio/commits/staging/multibuffer-support/) but, IMO, it is not the best way to handle and implement it.

## PR Type

## PR Checklist

### DIFF
--- a/iiod/responder.c
+++ b/iiod/responder.c
@@ -20,9 +20,8 @@
 #define ARRAY_SIZE(x) (sizeof(x) ? sizeof(x) / sizeof((x)[0]) : 0)
 
 /* Forward declaration */
-static struct iio_buffer * get_iio_buffer(struct parser_pdata *pdata,
-					  const struct iiod_command *cmd,
-					  struct buffer_entry **entry_ptr);
+static struct buffer_entry * get_iio_buffer_entry(struct parser_pdata *pdata,
+						  const struct iiod_command *cmd);
 
 static SLIST_HEAD(BufferList, buffer_entry) bufferlist;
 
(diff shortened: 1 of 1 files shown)


dNechita:
I think that both points make sense and that it's ok to continue on this direction.

PR #1393, opened 2026-02-10
Staging/iiod fix and simplify buffer entry

nunojsa:
## PR Description

The first commit fixes a possible leak and use after free (see commit for more details) and the second patch simplifies get_iio_buffer(). Note that this could just be one patch but I do think it makes for a better history to clearly separate the fix from the refactor (even tough there's no need for any backport right now).

## PR Type

## PR Checklist


mhennerich (approved):
LGTM

PR #1396, opened 2026-02-18
Create buffers during device discovery

nunojsa:
## PR Description

This is the first half of what was purposed at #1392 . For now, we are adding support to create buffers during device discovery and making sure we're still compatible with libiio (iiod) v0.26. To test this, I did:

1. iiod v0.26 with v1.0 client - Run iio_info and make sure I can read buffers attrs;
2. iiod v1.0 with v0.26 client - Same thing.

And of course both iiod and client on the same level. I also tested this on linux so it would very welcome if others can run mores tests on some other platforms.

The last two commits are already a preview of what is coming next. I'm putting them in this PR because they are simple enough and it allows to make the following PR a bit smaller.

## PR Type

## PR Checklist


utils/iio_info.c
```diff
@@ -367,18 +367,15 @@ int main(int argc, char **argv)
 			}
 		}
 
-		if (mask)
-			buffer = iio_device_create_buffer(dev, 0, mask);
-		if (mask && !iio_err(buffer)) {
+		buffer = iio_device_get_buffer(dev, 0);
```
dNechita, line 370: It is possible that multiple pre-allocated buffers exist?

nunojsa: Not for now. Or at least we do not support anyways. But yes, that's the end goal of all of this work

nunojsa:
> CI is detecting some build errors coming from serial.c.

Yeah, I messed up somewhere or might just be that I did not had serial enabled locally! Will re-spin next week

PR #1398, opened 2026-02-19
network-windows: Fix race condition in wait_cancellable

RubenNatanael:
## PR Description

Multiple threads calling wait_cancellable (e.g., one reading, one writing)  were overwriting the socket's WSAEventSelect configuration.

This commit adds a CriticalSection and waiter-counting logic to:
- Ensure FD_READ and FD_WRITE are merged rather than overwritten.
- Protect the shared event object reset and selection logic.
- Defer decrementing waiter counts until after the wait to prevent prematurely clearing the socket event mask.

Edit:
Issue: https://github.com/analogdevicesinc/libiio/issues/1402

## PR Type

## PR Checklist

### DIFF
--- a/network-windows.c
+++ b/network-windows.c
@@ -12,8 +12,15 @@
 #include <errno.h>
 #include <ws2tcpip.h>
 #include <iio/iio-debug.h>
+#include <iio/iio-lock.h>
 
 #define close(s) closesocket(s)
+
+struct iiod_client_pdata_os {
+	struct iio_mutex *event_lock;
+	int read_waiters;
+	int write_waiters;
+};
 
 int set_blocking_mode(int s, bool blocking)
 {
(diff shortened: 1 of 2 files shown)


rgetz:
it would likely be better if you re-used the primitives in `lock-windows.c`.

network.h
```diff
@@ -21,6 +21,11 @@ struct addrinfo;
 struct iiod_client_pdata {
 	int fd;
 
+#ifdef _WIN32
```
dNechita: I'd maintain platform independence for this file. The existing architecture consists of a single header file (network.h), a core implementation file (network.c) containing cross-platform code (with some exceptions), and separate platform-specific implementation files (network-windows.c and network-unix.c).

RubenNatanael: Modification done, let me know if it's ok now!

network-windows.c
```diff
@@ -42,11 +49,26 @@ int setup_cancel(struct iiod_client_pdata *io_ctx)
 		return -ENOMEM;
 	}
 
+	struct iio_event_sync *sync = (struct iio_event_sync*)malloc(sizeof(struct iio_event_sync));
```
rgetz: don't you need something here in case the malloc fails?

RubenNatanael: yep, thanks.

network-windows.c
```diff
@@ -42,11 +49,35 @@ int setup_cancel(struct iiod_client_pdata *io_ctx)
 		return -ENOMEM;
 	}
 
-	return 0;
+    struct iiod_client_pdata_os *os_priv = malloc(sizeof(*os_priv));
+    if (!os_priv) {
+        goto err_close_events;
+    }
+
```
rgetz, line 55: I think normal libiio coding style is that you don't need `{}` around one line statements.

RubenNatanael: Yes, thank you so much for the advices!

network-windows.c
```diff
@@ -42,11 +49,34 @@ int setup_cancel(struct iiod_client_pdata *io_ctx)
...
+    struct iiod_client_pdata_os *os_priv = malloc(sizeof(*os_priv));
+    if (!os_priv)
+        goto err_close_events;
+
+    os_priv->event_lock = iio_mutex_create();
+    if (!os_priv->event_lock) {
+        free(os_priv);
+        goto err_close_events;
+    }
+
```
rgetz, line 59: also similar here - it's more like other functions to move the `free(os_priv);` into the error/return path (at a new `err_free_os_priv` (or whatever your prefer), right before the existing `err_close_events:` label, and then in the above - you have a single `goto` without `{}`.

RubenNatanael: Right, thanks!

Issue #1401, opened 2026-02-20
Incorrect mapping from hwmon sensor type into iio sensor type


damijans:
iio library defines two bunch of sensor type, one for hwmon and the other for iio sensors.

hwmon sensor type are defined with 
enum hwmon_chan_type {
	HWMON_VOLTAGE,
	HWMON_FAN,
	HWMON_PWM,
	HWMON_TEMP,
	HWMON_CURRENT,
	HWMON_POWER,
	HWMON_ENERGY,
	HWMON_HUMIDITY,
	HWMON_INTRUSION,
	HWMON_CHAN_TYPE_UNKNOWN = IIO_CHAN_TYPE_UNKNOWN,
};
while iio sensor types are define with 
enum iio_chan_type {
	IIO_VOLTAGE,
	IIO_CURRENT,
	IIO_POWER,
	IIO_ACCEL,
	IIO_ANGL_VEL,
	IIO_MAGN,
	IIO_LIGHT,
	IIO_INTENSITY,
	IIO_PROXIMITY,
	IIO_TEMP,
	IIO_INCLI,
	IIO_ROT,
	IIO_ANGL,
	IIO_TIMESTAMP,
	IIO_CAPACITANCE,
	IIO_ALTVOLTAGE,
	IIO_CCT,
	IIO_PRESSURE,
	IIO_HUMIDITYRELATIVE,
	IIO_ACTIVITY,
	IIO_STEPS,
	IIO_ENERGY,
	IIO_DISTANCE,
	IIO_VELOCITY,
	IIO_CONCENTRATION,
	IIO_RESISTANCE,
	IIO_PH,
	IIO_UVINDEX,
	IIO_ELECTRICALCONDUCTIVITY,
	IIO_COUNT,
	IIO_INDEX,
	IIO_GRAVITY,
	IIO_POSITIONRELATIVE,
	IIO_PHASE,
	IIO_MASSCONCENTRATION,
	IIO_DELTA_ANGL,
	IIO_DELTA_VELOCITY,
	IIO_COLORTEMP,
	IIO_CHROMATICITY,
	IIO_ATTENTION,
	IIO_ALTCURRENT,
	IIO_CHAN_TYPE_UNKNOWN = INT_MAX
};

[...]

PR #1404, opened 2026-02-26
channel type: Improve code readability for channel type

RubenNatanael:
## PR Description

Currently, the channel type is stored as enum iio_chan_type, even when it represents a hwmon_chan_type. By changing this field to int, we improve code readability and flexibility. This change also enables the implementation of validation checks, such as in hwmon_channel_get_type(), which now issues a warning if a hwmon function is incorrectly called for a non-hwmon channel and vice versa.

- Issue: https://github.com/analogdevicesinc/libiio/issues/1401

## PR Type

## PR Checklist

### DIFF
--- a/channel.c
+++ b/channel.c
@@ -353,6 +353,9 @@ enum iio_modifier iio_channel_get_modifier(const struct iio_channel *chn)
 
 enum iio_chan_type iio_channel_get_type(const struct iio_channel *chn)
 {
+	if (iio_device_is_hwmon(iio_channel_get_device(chn))) {
+		chn_warn(chn, "iio_channel_get_type() called on non-hwmon channel\n");
+	}
 	return chn->type;
 }
 
@@ -412,6 +415,15 @@ const struct iio_data_format * iio_channel_get_data_format(
 		const struct iio_channel *chn)
 {
 	return &chn->format;
+}
+
+enum hwmon_chan_type hwmon_channel_get_type(
+		const struct iio_channel* chn)
+{
+	if (!iio_device_is_hwmon(iio_channel_get_device(chn))) {
+		chn_warn(chn, "hwmon_channel_get_type() called on non-hwmon channel\n");
+	}
+	return (enum hwmon_chan_type)chn->type;
 }
 
 bool iio_channel_is_enabled(const struct iio_channel *chn,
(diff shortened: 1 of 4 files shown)


channel.c
```diff
@@ -414,6 +417,15 @@ const struct iio_data_format * iio_channel_get_data_format(
 	return &chn->format;
 }
 
+enum hwmon_chan_type hwmon_channel_get_type(
+		const struct iio_channel* chn)
+{
+	if (!iio_device_is_hwmon(iio_channel_get_device(chn))) {
+		chn_warn(chn, "hwmon_channel_get_type() called on non-hwmon channel\n");
+	}
+	return (enum hwmon_chan_type)chn->type;
```
nunojsa, line 426: well, we give a warning but prevent nothing... IOW, we continue to do as before.

nunojsa (commented):
I'm not sure I agree with this change, It seems like unneeded churn to me. It is fair to assume that users will call hwmon_channel_get_type() on an hwmon device. If not, they need to fix their app anyways! So I would either:

1. Leave as-is;
2. Or be strict about it. Meaning, return an error in case we try to get an IIO channel from an hwmon device or vice-versa.

Honestly I would probably go with 1. For 2, we might need to change the API a bit...

Having said the above, we should have introduced different types for hwmon devices. Yeah, it would have been more complicated but we would not have these discussions.

channel.c
RubenNatanael: I agree. Like I said in the issue, this doesn't stop the error from happening, it's just a 'hint' for developers who assume  that the member `enum iio_chan_type type` in `struct iio_channel ` can't be an `hwmon_chan_type`. That said, I'm not entirely sure about this change either, so I'd like to hear other opinions.

dNechita: > well, we give a warning but prevent nothing... IOW, we continue to do as before.

Yes, we continue as before which aligns with your suggestion: "1. Leave as-is;"

Myself and @RubenNatanael looked at this from multiple angles and also came to the same conclusion as you and @rgetz that this isn't an issue, that there isn't an incorrect mapping taking place anywhere. Well, that not entirely true, because the only actor doing this mapping the debugger. So, we said, since we got to this point let's just fix that. Additionally, we thought to warn when `hwmon_channel_get_type()` and `iio_channel_get_type()` aren't used properly. But I see that at the same time we've updated the documentation asking users to make sure they are using the proper iio device by calling `iio_device_is_hwmon()`. But we're also calling it inside `hwmon_channel_get_type()` and `iio_channel_get_type()`. Maybe we just stick with updating the documentation and not issue the warnings.
But I don't see any harm in fixing the storage type for channel information.

PR #1424, opened 2026-04-13
Zephyr IIO device driver enhancement: DAC support, compat fixes

petejohanson-adi:
## PR Description

This PR has a few minor fixes, and one major feature addition. With these changes combined, the Zephyr implementation is sufficiently compatible with the https://github.com/analogdevicesinc/pyadi-iio library to consume, e.g. a connected ad5592r/ad5593r and reuse existing example scripts there unchanged.

### Major enhancement

- Rename/refactor the ADC IIO device driver to be a more generic "io-channels" device driver, allowing specifying both ADC and DAC channels in the `io-channels` property, and exposing the correct input/output channels on the IIO device.

### Minor enhancements

- Use the `io-channel-names` devicetree property to name the exposed IIO channels. This is important for compat with pyadi-iio, which for a device like the ad5592r with a pin that's both ADC and DAC mode, to have two channels both named the same `voltage#` with different input/output boolean metadata.
- Allow overriding the IIO *device* name explicitly, which is also needed for compat, where pyadi-iio often requires a matching device name for drivers like the ad5592r to be used.
- ~~A small build fix for including the `sys/sockets.h` header needed by at least some of the Zephyr libc implementations.~~ (Edit: removed from this PR, was no longer necessary)

## PR Type
- [X] New feature (a change that adds new functionality)

## PR Checklist


utilities.c
```diff
@@ -25,6 +25,9 @@
 #ifdef _WIN32
 #include <ws2tcpip.h>
 #elif defined(__has_include)
+#  if __has_include(<sys/socket.h>)
+#    include <sys/socket.h>
```
dNechita: If you're including here, it doesn't make sense to include it again on the 4th line.

petejohanson-adi: Ah, sorry, this was a side effect of a rebase and me missing this was in fact fixed a few lines further down. Removed that commit from this PR. Thanks.

dNechita (approved):
Looks good to me!

MaureenHelm (approved):

PR #1425, opened 2026-04-15
 Add buffer proper direction tracking and iio_buffer_is_output() public API

nunojsa:
## PR Description

 Previously, determining whether a buffer was used for TX (output) required scanning the device's channels for output scan elements via the internal iio_device_is_tx() helper. This approach was indirect—it queried the device rather than the buffer itself—and the same channel-scanning logic was duplicated across several call sites (iiod responder, compat layer, etc.). More important, the channel logic would not work for devices with more than one buffer (ore we can have in and out buffers)

The kernel (from some time now) expose a direction attribute on each IIO buffer via sysfs, making it possible to know the buffer direction directly without channel scanning and when creating the buffer.

 ### What this series does

 1. Read buffer direction from sysfs in the local backend and store it as a first-class field on struct iio_buffer.
 2. Serialize/deserialize the direction in the XML representation so remote backends also carry the information.
 3. Introduce iio_buffer_is_output() as a new public API, with a fast path when the direction is known and a fallback to channel scanning for older kernels.
 4. Convert all internal callers to the new API and remove the now-unused iio_device_is_tx().
 5. Expose the property in all language bindings (C++, Python, C#).

This should be the last step before supporting multi-buffer devices in libiio.

## PR Type

## PR Checklist


local.c
```diff
@@ -1340,11 +1341,24 @@ static int add_buffer_attributes(struct iio_device *dev, const char *devpath)
...
 		if (!buffer)
 			return -ENOMEM;
 
+		/* A buffer's direction cannot change at runtime, read it now */
+		ret = (int)local_do_read_dev_attr(iio_device_get_id(dev),
+						  0, "direction",
+						  dir_buf, sizeof(dir_buf),
+						  IIO_ATTR_TYPE_BUFFER);
+		if (ret > 0) {
+			if (!strcmp(dir_buf, "in"))
```
dNechita, line 1356: I assume that it's no possible to a different value other than "in" and "out" right? Or should we be extra careful and check for "out" as well?

dNechita (approved):
The commits are well-organized with logical grouping that made the changes easy to follow!

local.c
nunojsa: Nope! This must not change. It would be considered an ABI breakage from the kernel

nunojsa:
Alright! It seems the CI failures are also unrelated! Will merge this then

dNechita:
> Alright! It seems the CI failures are also unrelated! Will merge this then

The CI failures are related to a cmake script that does an audit checking files licenses. Will most likely be fixed soon. So yes, it isn't related to your changes.

PR #1435, opened 2026-04-29
iio_reg: fix context leak on successful register access

999Marius:
## PR Description
`read_reg()` and `write_reg()` returned directly from `main()`, bypassing `err_destroy_context` and leaking the entire IIO context on every successful invocation. Caught with Valgrind during testing.
Store the return value and fall through to the common cleanup block instead. This also fixes `err_destroy_context` incorrectly returning `EXIT_SUCCESS` when device lookup fails.

### Changes
- Store result of `read_reg()`/`write_reg()` in `ret` instead of returning directly
- Return `ret` instead of hardcoded `EXIT_SUCCESS` at cleanup label

## PR Type

## PR Checklist

### DIFF
--- a/utils/iio_reg.c
+++ b/utils/iio_reg.c
@@ -119,15 +119,15 @@ int main(int argc, char **argv)
 	addr = sanitize_clamp("register address", argw[optind + 1], 0, UINT32_MAX);
 
 	if ((argc - optind) == 2) {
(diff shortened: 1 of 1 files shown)


dNechita (approved):
That's a really good finding!
The fix looks good!

PR #1438, opened 2026-05-05
iiod-client: null-terminate XML string in non-ZSTD path

stefpopa:
## PR Description

In iiod_client_create_context_private_new(), the XML string received from iiod via the PRINT command is not null-terminated in the non-ZSTD code path. The buffer is then passed to iio_create_context_from_xml(), which eventually calls xmlReadMemory() with strlen(arg) to determine the length. Since the buffer is not null-terminated, strlen() reads past the XML data into uninitialized heap memory, causing intermittent XML parse errors ('Extra content at the end of the document').

The ZSTD code path correctly null-terminates the decompressed data, but the uncompressed path was missing this step.

Fix by allocating one extra byte and writing a null terminator after receiving the XML data.

## PR Type

## PR Checklist

### DIFF
--- a/iiod-client.c
+++ b/iiod-client.c
@@ -948,7 +948,7 @@ iiod_client_create_context_private_new(struct iiod_client *client,
 	bool is_zstd;
 	int ret;
 
-	xml = malloc(xml_len + uri_len);
+	xml = malloc(xml_len + uri_len + 1);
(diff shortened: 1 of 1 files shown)


dNechita (approved):
Great catch! Thanks for spotting and fixing this!

PR #1039, opened 2023-08-24
Rename "tests" to "utils" + remove deprecated parameters

pcercuei:
Move the whole "tests" folder to now be called "utils".

Remove the "-n", "-x" and "-s" parameters of those utility programs.

PR #1441, opened 2026-05-08
io_common: fix params forwarding for local backend, remove dead prefix code

999Marius:
## PR Description

Fix `iio_genxml -e` not including attribute values in the generated XML when using the local backend. The `-e` flag sets `IIO_CTX_XML_INCLUDE_VALUES` in `iio_context_params. This fix ensures that flag reaches the context for all backends.
### Changes

- Pass `&params` instead of `NULL` to `iio_create_context()` for the local backend path in `handle_common_opts()` in `utils/iio_common.c`
- Remove dead prefix variable and unreachable else if (prefix) block in handle_common_opts() — prefix was never assigned after f1b6706d removed `--ip/--xml`
### Backwards Compatibility

All other utilities (`iio_info`, `iio_attr`, `iio_rwdev`, `iio_reg`, `iio_event`) pass `NULL` as `default_params`, so `params` stays NULL in `handle_common_opts()`.

## PR Type

## PR Checklist

### DIFF
--- a/utils/iio_common.c
+++ b/utils/iio_common.c
@@ -368,7 +368,7 @@ struct iio_context * handle_common_opts(char * name, int argc,
 	else if (backend == IIO_URI)
 		ctx = iio_create_context(&params, arg);
 	else
-		ctx = iio_create_context(NULL, NULL);
+		ctx = iio_create_context(&params, NULL);
 
 	err = iio_err(ctx);
 	if (err && !do_scan && !detect_context) {
--- a/utils/iio_common.c
+++ b/utils/iio_common.c
@@ -262,7 +262,7 @@ struct iio_context * handle_common_opts(char * name, int argc,
 	if (default_params)
 		params = *default_params;
 	enum backend backend = IIO_LOCAL;
-	const char *arg = NULL, *prefix = NULL;
+	const char *arg = NULL;
 	bool do_scan = false, detect_context = false;
 	char buf[1024];
 	struct option *opts;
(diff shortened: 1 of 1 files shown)


999Marius:
**Note:**
 a second `iio_create_context(NULL, buf)` call exists for a prefix-URI path in `handle_common_opts()`, but that path seems to be dead code `prefix` is always `NULL`, left over from when `--ip` and `--xml` options existed for [prefix](https://github.com/analogdevicesinc/libiio/pull/1039/changes/b7509b26df99f02d26505d8265b07e966e80c830#diff-1b532735160cac438f373e62d20edfed223523390a161e9eab0d103bcfad6e95L303)

dNechita:
> **Note:** a second `iio_create_context(NULL, buf)` call exists for a prefix-URI path in `handle_common_opts()`, but that path seems to be dead code `prefix` is always `NULL`, left over from when `--ip` and `--xml` options existed

Can you check if it's indeed dead code? if yes, it makes sense to remove it

999Marius:
> > **Note:** a second `iio_create_context(NULL, buf)` call exists for a prefix-URI path in `handle_common_opts()`, but that path seems to be dead code `prefix` is always `NULL`, left over from when `--ip` and `--xml` options existed
> 
> Can you check if it's indeed dead code? if yes, it makes sense to remove it

Yes, it was dead code since prefix was initialized and never assigned again, prefix used to be assigned to `xml` or `ip` but those were removed when `--xml` `--ip` were deprecated in favor of `-u`.
### The last commit removed the dead code.

dNechita (approved):

PR #1443, opened 2026-05-12
Fix buflist_lock deadlock and simplify close path

nunojsa:
## PR Description

Fix a deadlock in handle_open_buffer() where buflist_lock was acquired  early but never released on error paths. While at it, simplify handle_close_buffer() which was doing a redundant double traversal of the buffer list with a TOCTOU race in between.

 - Patch 1 fixes the deadlock by deferring the lock to just before the list insertion, and switches to get_iio_buffer_entry() for the initial lookup.
 - Patch 2 reworks the close path to hold the lock across lookup and removal in a single critical section.

## PR Type

## PR Checklist

### DIFF
--- a/iiod/responder.c
+++ b/iiod/responder.c
@@ -377,7 +377,7 @@ static void handle_open_buffer(struct parser_pdata *pdata,
 	if (!dev)
 		goto err_send_response;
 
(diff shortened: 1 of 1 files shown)


dNechita (approved):
Looks good!

PR #1449, opened 2026-05-15
Add multi-buffer support

nunojsa:
## PR Description

This series adds support for devices that expose multiple IIO buffers (e.g. ADRV9104 with separate RX1, RX2, and TX buffers), and fixes a long-standing bug with cyclic TX transfers.

### Multi-buffer support

The Linux IIO subsystem gained multi-buffer support in v5.13, where a device can have multiple buffer directories (buffer0, buffer1, ...) each with its own set of scan elements. libiio previously hardcoded buffer0 and stored scan element metadata as per-channel "protected" attributes — an approach that made multi-buffer support cumbersome.

This series refactors scan element ownership from channels to buffers, introducing a struct iio_scan_element that lives in the buffer object. The local backend is reworked to discover all available buffers during context creation and associate scan elements with the correct buffer.  XML serialization/deserialization is extended to carry per-buffer channel mappings, with a legacy fallback for older IIOD versions.

 Two new public API functions are added:
 - iio_buffer_get_scan_elements_count()
 - iio_buffer_get_scan_element()

The iio_info and iio_rwdev utilities are updated to make use of the new API, with iio_rwdev gaining a -i/--buffer-index option.

### create_device() fix

We got a segfault and a double free if we fail during device creation. First patch is fixing that!

### Cyclic TX fix

The -c flag in iio_rwdev was silently broken: the stream API internally calls iio_block_enqueue() with cyclic=false, so the
hardware never actually looped the data. The fix bypasses the stream API for cyclic mode, using the block API directly with iio_block_enqueue(cyclic=true).

### Tested on ADRV9104 (ZCU102) with both cyclic and non-cyclic TX at 61.44 Msps.

 TODO

 There is still work to be done on top of this series:
 - Other utilities (e.g. iio_attr) need to gain multi-buffer awareness
 - Language bindings (C++, Python, C#) need to be updated for the new
 iio_buffer_get_scan_elements_count() / iio_buffer_get_scan_element()
 public APIs

## PR Type

## PR Checklist


nunojsa:
Changes in v2:
    * Fix spelling typos in local.c comments ("attributess", "muliple")
    * Fix iio_info buffer overflow: increase sprintf buffer and use snprintf
    * Add missing FMT_BUF macro for MSVC builds

nunojsa:
Changes in v3:
    * Remove unused variable `ret` in print_buffer() to fix MSVC -Werror (C4101)

local.c
```diff
@@ -1215,6 +1153,74 @@ static int detect_and_move_global_attrs(struct iio_device *dev)
...
+	attr_path = strchr(strstr(path, dev->id), '/') + 1;
+
+	channel_id = get_channel_id(dev, attr);
+	if (!channel_id)
+		return -ENOMEM;
+
+	chn = iio_device_find_channel(dev, channel_id, attr[0] == 'o');
+	if (!chn) {
+		/* If we can't find it, it means the channel is meant only for buffering which means
+		 * it has not other attributes than the scans. We still need to add it to the
```
dNechita: typo: `has not` -> `has no`

local.c
```diff
@@ -1311,16 +1308,77 @@ static int foreach_in_dir(const struct iio_context *ctx,
 	return ret;
 }
 
-static int add_scan_elements(struct iio_device *dev, const char *devpath)
+static int add_scan_element(void *d, const char *path)
```
dNechita: This `add_scan_element` function is defined twice.

dNechita: I was reviewing one commit at a time, then I saw the duplicate was removed later.

dNechita (approved):
Looks good to me!

dNechita:
@999Marius can help updating `iio_attr` including the code generation part. Would that work for you @nunojsa?

local.c
```diff
@@ -1674,9 +1679,9 @@ local_open_buffer(const struct iio_device *dev, unsigned int idx,
 
 	/* Finally, update the channels mask by reading the hardware again,
 	 * since some channels may be coupled together. */
-	for (i = 0; i < dev->nb_channels; i++) {
-		chn = dev->channels[i];
-		if (chn->index >= 0) {
+	for (i = 0; i < buffer->nb_scans; i++) {
+		chn = buffer->scans[i]->chn;
+		if (iio_channel_is_enabled(chn, mask)) {
```
999Marius: was iio_channel_is_enabled() intentional? The old code read states for all channels unconditionally to handle hardware coupled channels.

nunojsa:
> @999Marius can help updating iio_attr including the code generation part. Would that work for you @nunojsa?

Sure

local.c
nunojsa: Hmm I'm really not sure what that "hardware coupled channels" means tbh! So I kind of went with what makes sense. Just reading the channels we actually changed state. But ok, to be on the safe side, I'll keep the old behavior.

local.c
nunojsa: Oh, that does not matter :sweat_smile:. Clearly I messed up in my rebasing. We should never have a duplicated function in any commit! Will fix, thanks!

local.c
nunojsa: ack

nunojsa:
Changes in v4:
 * Fix typo "has not" → "has no" in local.c comment, per @dNechita
 * Remove duplicate add_scan_element() from commit 3 (rebase artifact), per @dNechita
 * Remove iio_channel_is_enabled() guard when reading back hardware state — preserves old behavior for coupled channels, per @999Marius

999Marius:
For `iio_attr` the flag `-i` is already taken (used to filter the input/output variant of a channel attribute). Using two different short flags for the same --buffer-index option across tools would be confusing. We may want to pick a letter that is free in both tools or fall back to long-form only.  There aren't many free flags in `iio_attr` that would make sense, `-n`, `-x`. 
I'd like a bit of feedback.

dNechita:
> For `iio_attr` the flag `-i` is already taken (used to filter the input/output variant of a channel attribute). Using two different short flags for the same --buffer-index option across tools would be confusing. We may want to pick a letter that is free in both tools or fall back to long-form only. There aren't many free flags in `iio_attr` that would make sense, `-n`, `-x`. I'd like a bit of feedback.

If we can't find a short flag for the buffer index that could be used by both `iio_rwdev` and `iio_attr` then I guess we can use `-b` for the 'iio_attr` to indicate which buffer to use.

It is safe to say that each tool in the utils directory is a specialised utility with its own concepts and functionality. These functions have distinct names and semantics, and it is possible for different tools to define functions or concepts that start with the same letter—resulting in single-letter flags that represent entirely different meanings depending on the tool.
A good example is that `iio_rwdev` uses `-c` to enable cyclic data transmission, whereas `iio_attr` uses `-c` to read from or write to channel attributes.

One could argue that certain concepts—such as (C)ontext, (d)evice, (c)hannel, (b)uffer, and (a)ttribute—are common across tools, and that the corresponding letters (C, d, c, b, a) could be reserved for these meanings. However, not all tools use—or are likely to ever use—this full set. For instance, `iio_rwdev` does not need concepts such as Context or channel, but it would still be inefficient not to use `-c` to represent the cyclic option.

I would favour using different letters that are intuitive and clearly tied to the concept, rather than enforcing consistency across tools with letters that are not self-explanatory—unless we find a single letter that satisfies both criteria.

PR #1459, opened 2026-05-27
device: fall back to attr method if backend reg op returns -ENOSYS

999Marius:
## PR Description

`iio_device_reg_write()` and `iio_device_reg_read()` now work correctly when a v1 client connects to a v0 (ASCII protocol) server over the network. Previously both functions returned `-ENOSYS` in that scenario, making register access completely broken against legacy servers.

### Changes
- `iio_device_reg_write()` in `device.c`: treat `-ENOSYS` as not supported on the legacy version and fall through to the `direct_reg_access` attribute method
- `iio_device_reg_read()` in `device.c`: same fix

### Testing

Tested against a v0 IIOD server (AD9361 platform) using `iio_device_reg_write` and `iio_device_reg_read` on the `cf-ad9361-dds-core-lpc` device. A full write -> read round-trip  returned the expected value (`0xA5A5A5A5`). Before the fix both functions returned `-38` (`-ENOSYS`).
## PR Type

## PR Checklist

### DIFF
--- a/device.c
+++ b/device.c
@@ -417,7 +417,9 @@ int iio_device_reg_write(struct iio_device *dev,
 
 	/* Use atomic backend operation if available */
 	if (ops && ops->reg_write) {
-		return ops->reg_write(dev, address, value);
+		int ret = ops->reg_write(dev, address, value);
+		if (ret != -ENOSYS)
+			return ret;
 	}
 
 	/* Fallback to the original attribute-based implementation.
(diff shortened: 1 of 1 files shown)


dNechita (approved):
Great catch!
The fallback was originally intended to handle backends that don’t implement `reg_read/reg_write`, by reverting to classic attribute read/write. However, it didn’t account for the case where a libiio v1 client connects to a libiio v0 IIOD, which doesn’t recognise the reg_read/reg_write commands. This is another scenario where falling back to classic attribute access is necessary.

PR #1461, opened 2026-06-01
iio_attr: add -f flag to write raw bytes from a file

mhennerich:
Adds a -f / --input-file flag that re-interprets the trailing [value] positional as a path. The file's bytes are loaded into memory and written via the iio_*_attr_write_raw API variant, allowing binary payloads (e.g. FIR filter configs) to be pushed to attributes that don't accept a string form.

Combinations that don't make sense are rejected up front: -C (context attributes), --generate-code, and wildcard matches. Empty files are rejected, and a short write (fewer bytes accepted than supplied) is reported as a WARNING on stderr. The man page synopsis and OPTIONS section are updated accordingly.

### DIFF
--- a/tests/iio_attr.c
+++ b/tests/iio_attr.c
@@ -7,6 +7,10 @@
  *         Robin Getz <robin.getz@analog.com>
  * */
 
+#ifdef _MSC_VER
+#define _CRT_SECURE_NO_WARNINGS
+#endif
+
 #include <errno.h>
 #include <getopt.h>
 #include <iio.h>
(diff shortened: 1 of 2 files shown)


tests/iio_attr.c
```diff
@@ -7,6 +7,10 @@
  *         Robin Getz <robin.getz@analog.com>
  * */
 
+#ifdef _MSC_VER
+#define _CRT_SECURE_NO_WARNINGS
```
dNechita, line 11: This opens the door for using other functions that are considered not secure on Windows. While `fopen_s()` provides limited additional security—only verifying that the file pointer is not NULL, which is already handled here—other functions, such as `strcpy()`, could benefit from being replaced with safer alternatives like `strcpy_s()` in the future, as these help prevent buffer overflows.
I think this approach is better:
```
#ifdef _MSC_BUILD
	fopen_s()
#else
	fopen()
#endif
```

mhennerich: I created this PR as draft just to get WIN build.
If you're ok with eventually merging it into v0.x I'll fix those warnings properly.
The _CRT_SECURE_NO_WARNINGS was just a quick way to get a build...

dNechita: Since this is more of a feature than a bug fix, I am inclined to say that it should not get merged to v0.x.

dNechita: > I created this PR as draft just to get WIN build. 
There is also the option to create a branch starting with "staging/" which triggers a win build. Full list of conditions to trigger Azure is here: https://github.com/analogdevicesinc/libiio/blob/libiio-v0/azure-pipelines.yml#L4:L12

PR #1471, opened 2026-06-17
gh-pages: restructure site for versioned documentation

999Marius:
## PR Description
This PR restructures the libiio GitHub Pages site to work with the adi_doctools Sphinx theme and its version switcher dropdown. The old hand-crafted HTML landing page and its assets are replaced with a redirect to `main/`, and a `tags.json` file is added at the site root so the version switcher can list all deployed versions. A companion PR against `main` will add CI automation to keep `tags.json` updated after each deployment.

### Benefits
- Root URL now lands on the new Sphinx-based docs instead of the old static page
- All v0.x Doxygen docs remain accessible, nothing is removed
- `tags.json` enables the adi_doctools version switcher to list all past and future versions

### Changes
- Replace root `index.html` with a redirect to `main/`
- Remove old `img/` assets (style.css, version.js, logos) that belonged to the old landing page
- Add `tags.json` listing all deployed versions for the adi_doctools version switcher

### Backwards Compatibility
All v0.x Doxygen documentation directories remain untouched and accessible at their existing URLs. Only the root landing page and its associated assets are replaced.

## PR Type
- [ ] Bug fix (a change that fixes an issue)
- [x] New feature (a change that adds new functionality)
- [ ] Breaking change (a change that affects other repos or cause CIs to fail)

## PR Checklist
- [x] I have conducted a self-review of my own code changes
- [ ] I have commented new code, particularly complex or unclear areas
- [x] I have checked that I did not introduce new warnings or errors (CI output)
- [ ] I have checked that components that use libiio did not get broken
- [ ] I have updated the documentation accordingly (GitHub Pages, READMEs, etc)

PR #1472, opened 2026-06-17
CI: add versioned doc deployment with tags.json support

999Marius:
## PR Description
This PR wires up automatic `tags.json` maintenance to the documentation CI, enabling the version switcher dropdown provided by the adi_doctools Sphinx theme to reflect all deployed versions of the libiio docs. After each deploy to `main` or a release tag, the workflow scans the live gh-pages branch, validates discovered versions against remote git tags, and pushes an updated `tags.json` to the root of gh-pages. Works in conjunction with #1471  which restructures the gh-pages branch to support versioned deployments.

### Benefits
- Version switcher dropdown in the adi_doctools theme stays up to date automatically after every deployment
- Legacy v0.x Doxygen docs are included in the dropdown without manual maintenance, only directories backed by a real git tag are shown
- `ADOC_DOC_VERSION` ensures the theme highlights the correct version during each build

### Changes
- Set `ADOC_DOC_VERSION` env var on the "Check doc build" step so the theme knows which version is being built
- Add "Checkout gh-pages for scanning" step that checks out the gh-pages branch into `gh-pages-content/`
- Add "Generate tags.json" step that discovers deployed Sphinx versions via `objects.inv`, cross-references `v0.*` directories against remote git tags, and writes a combined `tags.json`
- Add "Deploy tags.json to gh-pages" step that pushes the result to the root of gh-pages using `peaceiris/actions-gh-pages` with `keep_files: true`

## PR Type

## PR Checklist


dNechita (approved):
Looks good!

PR #1477, opened 2026-06-22
 add iio_context_ping() API and iio_ping utility

999Marius:
## PR Description

This PR adds `IIOD_OP_NOP` (opcode 27), a new no-operation command to the IIOD binary protocol, along with a public `iio_context_ping()` API and an `iio_ping` command-line utility. It gives applications a way to verify that a remote IIO context (network, USB, serial) is still alive and responsive , without touching any hardware or changing device state.

### Benefits
- Enables connection checks without reading attributes or changing state
- Works across all remote backends (network, USB, serial) through a single API call
- Includes `iio_ping` utility for  connection testing

### Changes
  - Add IIOD_OP_NOP (opcode 27) to iiod-responder.h, handle_nop() to iiod/responder.c, and iiod_client_nop()
  to iiod-client.c
  - Add .ping to iio_backend_ops and implement iio_context_ping() in context.c with -EINVAL fallback to
  iio_context_set_timeout()
  - Add network_ping(), usb_ping(), serial_ping(), local_ping(), emu_ping() to their respective backends
  - Add iio_ping utility with -c and -i options, and man page

### How it works

1. The caller invokes `iio_context_ping(ctx)`.
2. `iio_context_ping()` calls `ctx->ops->ping(ctx)` for example `network_ping()` for a network context.
3. `network_ping()` calls `iiod_client_nop(pdata->iiod_client)`.
4. `iiod_client_nop()`  builds an `iiod_command` with `op = IIOD_OP_NOP` and sends it via `iiod_io_exec_simple_command()`. 
5. The reader thread dispatches opcode 27 to `handle_nop()`, which calls `iiod_io_send_response_code(io, 0)` and the client returns 0.

### Usage

```c
/* API */
struct iio_context *ctx = iio_create_context(NULL, "ip:192.168.1.10");

if (iio_context_ping(ctx) < 0) {
    iio_context_destroy(ctx);
    ctx = iio_create_context(NULL, "ip:192.168.1.10");
}
```

```sh
# CLI
iio_ping -u ip:192.168.0.2 -c 4 -i 1000
PING ip backend (Linux pi5 6.12.62+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.12.62-1+rpt1 (2025-12-18) aarch64 192.168.0.2)
Reply from ip: seq=0 time=1.616 ms
Reply from ip: seq=1 time=2.038 ms
Reply from ip: seq=2 time=2.732 ms
Reply from ip: seq=3 time=1.983 ms

--- ip ping statistics ---
4 requests sent, 4 received, 0.0% loss
```

## PR Type

## PR Checklist

### DIFF
--- a/context.c
+++ b/context.c
@@ -667,6 +667,14 @@ int iio_context_set_timeout(struct iio_context *ctx, int timeout)
 	return 0;
 }
 
+int iio_context_ping(struct iio_context *ctx)
+{
+	if (!ctx->ops->ping)
+		return -ENOTSUP;
+
+	return ctx->ops->ping(ctx);
+}
+
 const struct iio_backend *const iio_backends[] = {
 	IIO_IF_ENABLED(WITH_LOCAL_BACKEND, &iio_local_backend),
 	IIO_IF_ENABLED(WITH_NETWORK_BACKEND && !WITH_NETWORK_BACKEND_DYNAMIC, &iio_ip_backend),
--- a/iiod-client.c
+++ b/iiod-client.c
@@ -1947,3 +1947,19 @@ int iiod_client_reg_write(struct iiod_client *client, const struct iio_device *d
 
 	return iiod_io_wait_for_command_done(io);
 }
+
+int iiod_client_nop(struct iiod_client *client)
+{
+	struct iiod_command cmd;
+	struct iiod_io *io;
+
+	if (!iiod_client_uses_binary_interface(client))
+		return -ENOTSUP;
+
+	cmd.op = IIOD_OP_NOP;
+	cmd.dev = 0;
+	cmd.code = 0;
+
+	io = iiod_responder_get_default_io(client->responder);
+	return iiod_io_exec_simple_command(io, &cmd);
+}
--- a/bindings/cpp/iio.hpp
+++ b/bindings/cpp/iio.hpp
@@ -683,6 +683,7 @@ public:
     optional<Device> device(unsigned int idx) const { return impl::maybe<Device>(iio_context_get_device(p, idx)); }
     optional<Device> find_device(cstr name) const {return impl::maybe<Device>(iio_context_find_device(p, name));}
     void set_timeout(int timeout_ms){impl::check(iio_context_set_timeout(p, timeout_ms), "iio_context_set_timeout");}
+    void ping(){impl::check(iio_context_ping(p), "iio_context_ping");}
     iio_context_params const * params() const {return iio_context_get_params(p);}
     void set_data(void * data){iio_context_set_data(p, data);}
     void * data() const {return iio_context_get_data(p);}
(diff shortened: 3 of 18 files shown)


context.c
```diff
@@ -667,6 +667,24 @@ int iio_context_set_timeout(struct iio_context *ctx, int timeout)
...
+
+	if (!ctx->ops->ping)
+		return -ENOTSUP;
+
+	ret = ctx->ops->ping(ctx);
+
+	/* Old servers don't know IIOD_OP_NOP and return -EINVAL.
+	 * Fall back to set_timeout which does a real round-trip
+	 * without touching hardware. */
+	if (ret == -EINVAL)
```
dNechita: there should be no fallback. If the IIOD does not have support then that is that. The contract that is mentioned in iio.h states that "does not access hardware or change any device state" but this fallback can potentially change timeout which changes the state of the IIOD and how libiio interacts with the hardware. Also, setting timeout is only supported in v1 which means that if v0 is present then the fallback won't work anyway.

dNechita (commented):
There is a typo in the first commit message. It's a tab splitting a word:
"is st	ill alive without side effects."

iiod-client.c
```diff
@@ -1947,3 +1947,27 @@ int iiod_client_reg_write(struct iiod_client *client, const struct iio_device *d
...
+		return iiod_io_exec_simple_command(io, &cmd);
+	} else {
+		char buf[] = "NOP\r\n";
+		int ret;
+
+		iio_mutex_lock(client->lock);
+		ret = iiod_client_exec_command(client, buf);
+		iio_mutex_unlock(client->lock);
+
+		return ret;
```
dNechita: The ASCII IIOD does not recognise the NOP command, so there is no clear reason to send it. We should either implement support for it on the IIOD side or return a “not supported” error in the fallback (else) branch.
In general, no new features were added to the ASCII IIOD beyond the timeout handling, and nothing was implemented for event attributes. The current approach is therefore to introduce new functionality only in the binary IIOD.

context.c
999Marius: Fixed, removed the v0 fallback. Now iio_context_ping is exlusive to v1 binary protocol.

iiod-client.c
999Marius: Fixed, removed the ASCII else branch. NOP is now sent exclusively through the binary protocol.

999Marius:
**Update v1:**
 - Removed ASCII fallback from iiod_client_nop(), NOP is binary-protocol only
 - iio_context_ping() now returns the backend result of ping directly, no special handling for v0 server

dNechita (approved):
Looks good!

999Marius:
**Update 2:**
- Added python/csharp/cpp bindings for `iio_context_ping`

dNechita (approved):

PR #1487, opened 2026-07-02
Add buffer-specific scan element properties and attributes

dNechita:
## PR Description

Make scan element properties (index, data format, sample size) buffer-specific to properly support the channel-buffer relationship in multi-buffer scenarios.

### Background

libiio already supports multiple buffers per device through the `iio_buffer` API. However, scan element properties were previously treated as channel-global, accessed without a buffer context. This worked for single-buffer devices but was incorrect for multi-buffer scenarios where the same channel can have different properties depending on which buffer it's accessed through.

Linux kernel 5.11+ exposed this limitation by changing the sysfs structure to make scan element attributes explicitly per-buffer:

**Legacy kernel (single buffer):**
/sys/bus/iio/devices/iio:deviceX/
├── scan_elements/             # scan element index, type, en
└── in_voltageY_*             # channel attributes

**Modern kernel (multi-buffer):**
/sys/bus/iio/devices/iio:deviceX/
├── buffer0/                   # scan elements embedded in buffer
│   ├── in_voltageY_index     # channel's index in buffer0
│   ├── in_voltageY_type      # channel's format in buffer0
│   └── in_voltageY_en
├── buffer1/                   # different scan elements per buffer
│   ├── in_voltageY_index     # may differ from buffer0
│   └── in_voltageY_type      # may differ from buffer0
├── scan_elements/             # legacy compat (kernel provides)
└── in_voltageY_*             # channel attributes (device-level)

In multi-buffer configurations, a channel's scan element properties can legitimately differ between buffers - different index ordering, different data formats, etc.

#### API changes - scan element properties now require buffer context

**Before (incorrect for multi-buffer):**
```c
long idx = iio_channel_get_index(chn);
const struct iio_data_format *fmt = iio_channel_get_data_format(chn);
size_t sample_sz = iio_device_get_sample_size(dev, mask);

After (buffer-specific):
long idx = iio_channel_get_index(chn, buf);
const struct iio_data_format *fmt = iio_channel_get_data_format(chn, buf);
size_t sample_sz = iio_buffer_get_sample_size(buf, mask);
// Pass NULL for single-buffer use cases: iio_channel_get_index(chn, NULL)

Modified functions:
- iio_channel_get_index(chn, buf) - index is buffer-specific
- iio_channel_get_data_format(chn, buf) - format is buffer-specific
- iio_channel_convert(chn, dst, src, buf) - conversion uses buffer-specific format
- iio_channel_convert_inverse(chn, dst, src, buf) - conversion uses buffer-specific format
- iio_channel_is_scan_element(chn, buf) - channel may be scan element in one buffer but not another
- iio_buffer_get_sample_size(buf, mask) - replaces device-level function (sample size is inherently per-buffer)
```

Passing NULL for the buffer parameter defaults to buffer0 for backwards compatibility and convenience in single-buffer applications.

Scan element attributes

- Scan elements can now have additional attributes beyond the core index, type, en
- New API functions expose per-buffer scan element attributes
- XML serialization updated to capture per-buffer scan element configuration

Backend implementation

- Local backend: Detects kernel version and reads from appropriate paths
  - Buffer0: uses legacy scan_elements/ path for old kernels, buffer0/ for new kernels
  - BufferN (N>0): reads from bufferN/ directory (multi-buffer kernels only)
  - Maintains backwards compatibility with single-buffer kernels
- emu backend: Extended to support per-buffer scan element attributes in XML contexts
- iiod protocol: Extended to read/write scan element attributes over network

Backwards compatibility

- NULL buffer parameter defaults to buffer0 for easy migration
- Buffer0 reads from legacy paths when available for maximum old-kernel compatibility
- Single-buffer applications work with minimal changes (add NULL parameters)

## PR Type

## PR Checklist


nunojsa:
> This worked for single-buffer devices but was incorrect for multi-buffer scenarios where the same channel can have different properties depending on which buffer it's accessed through.

Not sure I understand the above? A scan element is something we define as part of the channel and will be pretty much the same for all buffers (if the channel can be used in multiple buffers). That said, a channel can have multiple scan elements but only one active at a time (eg: channel has some oversampling attr). But that is a different problem.

> /sys/bus/iio/devices/iio:deviceX/
├── buffer0/ # scan elements embedded in buffer
│ ├── in_voltageY_index # channel's index in buffer0
│ ├── in_voltageY_type # channel's format in buffer0
│ └── in_voltageY_en
├── buffer1/ # different scan elements per buffer
│ ├── in_voltageY_index # may differ from buffer0
│ └── in_voltageY_type # may differ from buffer0
├── scan_elements/ # legacy compat (kernel provides)
└── in_voltageY_* # channel attributes (device-level)

Nope. The above is a wrong assumption. As I said, a channel can have multiple scan elements (and the thing that actually differs is the [type](https://elixir.bootlin.com/linux/v7.1.2/source/include/linux/iio/iio.h#L268)) but only one active at a time and it will be the same for all buffers where that channel is available. And actually, for all real multibuffer usecases we have (and that I'm aware of), each channel needs to be routed only through one buffer.

Now I agree the channel APIs to handle scan elements are not great. We could expose an iio_scan object and an iio_scan related API (I did thought about introducing it when I reworked the buffer stuff) but I don't really think that pays that much and it will slow down the release again :D

dNechita:
> > This worked for single-buffer devices but was incorrect for multi-buffer scenarios where the same channel can have different properties depending on which buffer it's accessed through.
> 
> Not sure I understand the above? A scan element is something we define as part of the channel and will be pretty much the same for all buffers (if the channel can be used in multiple buffers). That said, a channel can have multiple scan elements but only one active at a time (eg: channel has some oversampling attr). But that is a different problem.

Can a channel be positioned as the 1st channel in buffer0 and the 2nd in buffer1?

 
> > /sys/bus/iio/devices/iio:deviceX/
> > ├── buffer0/ # scan elements embedded in buffer
> > │ ├── in_voltageY_index # channel's index in buffer0
> > │ ├── in_voltageY_type # channel's format in buffer0
> > │ └── in_voltageY_en
> > ├── buffer1/ # different scan elements per buffer
> > │ ├── in_voltageY_index # may differ from buffer0
> > │ └── in_voltageY_type # may differ from buffer0
> > ├── scan_elements/ # legacy compat (kernel provides)
> > └── in_voltageY_* # channel attributes (device-level)
> 
> Nope. The above is a wrong assumption. As I said, a channel can have multiple scan elements (and the thing that actually differs is the [type](https://elixir.bootlin.com/linux/v7.1.2/source/include/linux/iio/iio.h#L268)) but only one active at a time and it will be the same for all buffers where that channel is available. And actually, for all real multibuffer usecases we have (and that I'm aware of), each channel needs to be routed only through one buffer.

If there are 4 channels:
 - voltage0
 - voltage1
 - voltage2
 - voltage3
, and voltage0 and voltage3 are used in buffer0 (I assume these channels will have indexes 0 and 1, right?)
, and voltage1 and voltage2 are used in buffer1 (I assume these channels will have indexes 0 and 1, right?)
then calling:
```c
iio_channel_get_index(voltage0); // will return 0
iio_channel_get_index(voltage1); // will also return 0 
```
will return the same index but for different buffers. Is this assumption correct?

>As I said, a channel can have multiple scan elements (and the thing that actually differs is the [type](https://elixir.bootlin.com/linux/v7.1.2/source/include/linux/iio/iio.h#L268)) but only one active at a time and it will be the same for all buffers where that channel is available

Do you mean that `type` can change at runtime but absolutely cannot have 2 or more different values at the same time for a channel?

nunojsa:
> Can a channel be positioned as the 1st channel in buffer0 and the 2nd in buffer1?

Yes. The key is that it will have the same index in both buffers. The device scan_mask will just account for all existing channels that are scan_elements.

> , and voltage0 and voltage3 are used in buffer0 (I assume these channels will have indexes 0 and 1, right?)
, and voltage1 and voltage2 are used in buffer1 (I assume these channels will have indexes 0 and 1, right?)
then calling:

No. In theory and assuming linearity:

* voltage0 - index 0
* voltage1 - index 1
* voltage2 - index 2
* voltage3 - index 3

And the above is true independent on which buffer(s) the scan is available.

> will return the same index but for different buffers. Is this assumption correct?

Hopefully with the above the answer is clear :)

> Do you mean that type can change at runtime but absolutely cannot have 2 or more different values at the same time for a channel?

Yes! And it will be the same type for all the buffers. I think we already discussed this (maybe there's even an open issue for it). But this is another story. Use case is converters that have some oversampling functionality that increases the sample resolution. That means that number of the sample realbits will change when such a feature is enabled on the device. This was the reason why allowing multiple scan types per element was introduced

dNechita:
Closing this PR because I created it based on the mistaken assumption that index and type are per-buffer attributes. In reality, they are defined per channel.

PR #1488, opened 2026-07-02
network: Fix pipe2() fallback handling

larsclausen:
When `pipe2()` succeeds, `create_cancel_fd()` still falls through to `pipe()`. This overwrites the stored cancel pipe file descriptors and leaks the descriptors returned by `pipe2()`.

Return immediately after a successful `pipe2()` call so `pipe()` is only used for the intended `ENOSYS` fallback case.

Fixes: d2ca7db254bd ("network: Add cancellation support")


mhennerich:
Welcome back!
I assume this fix needs to go into the libiio-v0 branch as well.
Greets,
Michael

dNechita (approved):
Looks good!
Thanks Lars!

PR #1491, opened 2026-07-06
zephyr: samples/iiod: fix native_sim network test races

dimitrije-lilic:
## PR Description

Two independent races made the pytest sample fail intermittently on the
Zephyr CI ubuntu-22.04 runner:

  1. Port collision. sample.iiod.network (console harness) and
     sample.iiod.network.pytest (pytest harness) built two separate
     native_sim zephyr.exe images that both bind TCP port 30431. Twister
     runs native_sim tests in parallel, so whichever DUT lost the bind()
     race got errno 112 (EADDRINUSE) and never advertised its IIO
     context:

         <err> libiio: Failed to bind socket: -1 (errno: 112)
         <err> libiio: Failed to create test server

     sample.iiod.network only checked that the DUT prints "Booting Zephyr
     OS", which is a strict subset of what the pytest variant already
     verifies (it also boots the DUT and then talks the full IIOD
     protocol to it). Drop the redundant console-harness test so nothing
     else contends for port 30431.

  2. Client-side race. The IIOD server thread opens its listen socket
     before iio_context_add_device() is called for each configured
     device. If pytest connects between those two steps -- which the old
     fixture did as soon as socket.create_connection() succeeded -- it
     read back an empty XML and cached an empty device list, causing
     every test to fail on `iiod_context.devices`. Retry iio.Context()
     from the fixture until it reports at least one device or a 10 s
     deadline expires.

With these fixes, `west twister -T samples --integration` passes 10/10
native_sim test cases under default parallel execution.

Signed-off-by: Dimitrije Lilic <dimitrije.lilic@orioninc.com>

## PR Type

## PR Checklist

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -60,7 +60,7 @@ jobs:
           if [ "${{ runner.os }}" = "Windows" ]; then
             EXTRA_TWISTER_FLAGS="--short-build-path -O/tmp/twister-out"
           fi
-          west twister -T samples -v --force-color --inline-logs --integration --fixture iiod_tcp_30431 $EXTRA_TWISTER_FLAGS
+          west twister -T samples -v --force-color --inline-logs --integration $EXTRA_TWISTER_FLAGS
 
   regression-tests:
     name: Regression tests against Zephyr IIOD backend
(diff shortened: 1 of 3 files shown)


MaureenHelm (commented):
I agree we should remove `sample.iiod.network` since it's covered by `sample.iiod.network.pytest`. But let's squash the changes into one commit rather than introducing `--fixture iiod_tcp_30431` in one commit and removing it in the following commit

PR #1493, opened 2026-07-08
Add runtime channel format refresh with IIOD protocol support

dNechita:
## PR Description

Implements `iio_channel_refresh_format()` API to update cached channel formats at runtime when hardware configuration changes (e.g., modifying `oversampling_ratio` attribute). Includes full IIOD protocol support to synchronize format changes between client and server.

## Motivation

When IIO device attributes like `oversampling_ratio` change, the underlying scan element format can change dynamically (e.g., from `le:s24/32>>0` to `le:s28/32>>0`). Previously, libiio cached the format at device enumeration time with no way to refresh it without re-creating the entire context. This caused:

- **Data corruption** when reading buffers after format changes
- **Sample size mismatch** between client and iiod server cached values
- Forced application restarts to pick up hardware configuration changes

Adding this PR as a replacement for #1368.

### New API

```c
int iio_channel_refresh_format(struct iio_channel *chn);
```

Refreshes the cached format for a scan element channel. Backend-specific implementation:
- Local backend: Re-reads sysfs type attribute
- Network/USB/Serial backends: Sends IIOD_OP_REFRESH_FORMAT to server

IIOD Protocol Extension

Added new REFRESH command:
Client → Server: REFRESH <device-id> <channel-id>
Server → Client: le:s24/32>>0\n  (format string)

The server:
1. Atomically re-reads the format from sysfs
2. Returns the format string to the client
3. Invalidates cached sample_size for the device (entry->update_mask = true)

Why Format String Response?

- Atomic operation: Prevents TOCTOU race conditions in multi-client scenarios
- Self-describing: Client receives authoritative format from server
- Future-proof: Compatible with protocol versioning and extensions
- Consistent: Matches existing XML serialization convention

iio_attr Tool Integration

Added -r/--refresh option:
```
iio_attr -r <device>           # Refresh all scan element channels
iio_attr -r <device> <channel> # Refresh specific channel
iio_attr -r -v <device>        # Verbose output with format details
```
Implementation Details

Client-Side

- Generic iio_channel_refresh_format() delegates to backend ops->refresh_format()
- Each backend (local, network, serial, USB) implements format refresh
- Client parses format string and updates chn->format structure

Server-Side (iiod)

- REFRESH command handler in iiod/responder.c
- Calls iiod_device_refresh_format() which:
  - Re-reads format via iio_device_get_channel_format()
  - Invalidates DevEntry->update_mask to force sample_size recalculation
  - Returns format string to client
- Thread-safe: uses existing device locking mechanisms

Backward Compatibility

- Old clients with new servers: no impact (don't send REFRESH)
- New clients with old servers: command returns error, client falls back gracefully
- Protocol versioning handles capability detection
## PR Type

## PR Checklist


dNechita:
V1:

- fix Windows compiler warnings about signed vs. unsigned integer mismatches
- added bindings for C#, Python, C++

bindings/csharp/Channel.cs
```diff
@@ -383,5 +386,21 @@ public uint write(Block block, byte[] array, bool raw = false)
 
             return count;
         }
+
+        /// <summary>
+        /// Refresh the channel's data format from the hardware.
+        /// Call this after changing any attribute that affects channel format
+        /// (e.g., oversampling_ratio) and before opening a buffer.</summary>
+        /// <exception cref="IioLib.IIOException">The format could not be refreshed.</exception>
+        public void refresh_format()
```
gastmaier, line 395: Wouldn't be better to call refresh_scan_type?
iio_channel_refresh_format -> iio_channel_refresh_scan_type

channel.c do use "format" instead of "scan_type" like the kernel does, still, I don't think "format" naming is exposed yet.
(same everywhere)

doc/source/tools/iio_attr.rst
```diff
@@ -42,6 +43,11 @@ The iio_attr utility has a few main options, which control what the main utility
 **-D --debug-attr**
    Read and Write IIO Debug attributes
 
+**-r, --refresh**
+   Refresh channel format(s). Device must be buffer-capable. Channel must
+   be a scan element. If no channel is specified, all scan element channels
+   on the device are refreshed.
```
gastmaier:
```suggestion
   Refresh channel scan type. If no channel is specified, all scan element
   channels on the device are refreshed.
```

"Device must be buffer-capable." fair, but the channel only have a scan type if it is buffer capable; like this it implies that there are channels with scan types that cannot be refreshed because they have no buffer.

utils/iio_attr.c
```diff
@@ -858,6 +879,73 @@ int main(int argc, char **argv)
...
+					}
+
+					/* If specific channel was requested, we're done */
+					if (channel_index)
+						break;
+				}
+
+				if (quiet != ATTR_QUIET && !channel_index) {
+					printf("Refreshed %d channel format%s on device %s",
+					       refresh_count, refresh_count != 1 ? "s" : "",
```
gastmaier: always printed, even without `-v`
```
$ iio_attr -u ip:10.44.3.94 -r iio:device0 -v
Refreshed channel voltage0: bits=20 length=32 shift=0 repeat=1
Refreshed 1 channel format on device ad4062
$ iio_attr -u ip:10.44.3.94 -r iio:device0
Refreshed 1 channel format on device ad4062
```
I don't think this message `Refreshed %d channel format%s on device %s` is needed on success without `-v`.

utils/iio_attr.c
```diff
@@ -858,6 +879,73 @@ int main(int argc, char **argv)
...
+					if (err < 0) {
+						if (quiet != ATTR_QUIET) {
+							IIO_PERROR(err, "Failed to refresh format for channel %s", ch_id);
+						}
+						refresh_err++;
+					} else {
+						if (quiet == ATTR_VERBOSE) {
+							const struct iio_data_format *fmt = iio_channel_get_data_format(ch);
+							printf("Refreshed channel %s: bits=%u length=%u shift=%u repeat=%u\n",
+							       ch_id, fmt->bits, fmt->length, fmt->shift, fmt->repeat);
```
gastmaier: Only printed with `-v`
```
$ iio_attr -u ip:10.44.3.94 -r iio:device0 voltage0 -v
Refreshed channel voltage0: bits=20 length=32 shift=0 repeat=1
```

gastmaier (changes requested):
ok, so like #1368 it is an explicit flag, we offload the responsibility to gui apps, pyadi-iio

I didn't get the expected behavior :/

verified:
```
$ iio_attr -h | grep refre -A 1
        -r, --refresh
                        Refresh channel format(s). Device must be buffer-capable. Channel must be scan element.
$ iio_attr -u ip:10.44.3.94 -c iio:device0
dev 'ad4062', channel 'voltage0' (input, index: 0, format: be:S16/16>>0), found 5 channel-specific attributes

$ iio_attr -u ip:10.44.3.94 -c iio:device0 voltage0
dev 'ad4062', channel 'voltage0' (input), attr 'calibscale', value '1.000000000'
dev 'ad4062', channel 'voltage0' (input), attr 'oversampling_ratio', value '1'

$ iio_attr -u ip:10.44.3.94 -c iio:device0 voltage0 oversampling_ratio 2 

# Old value (expected)
$ iio_attr -u ip:10.44.3.94 -c iio:device0
dev 'ad4062', channel 'voltage0' (input, index: 0, format: be:S16/16>>0), found 5 channel-specific attributes

$ iio_attr -u ip:10.44.3.94 -r iio:device0
Refreshed 1 channel format on device ad4062

$ iio_attr -u ip:10.44.3.94 -r iio:device0 voltage0
# Wirdly no output

# Old value still (wrong)
$ iio_attr -u ip:10.44.3.94 -c iio:device0
dev 'ad4062', channel 'voltage0' (input, index: 0, format: be:S16/16>>0), found 5 channel-specific attributes
```

At the target
```
#  iio_attr -c iio:device0
dev 'ad4062', channel 'voltage0' (input, index: 0, format: be:s20/32>>0), found 5 channel-specific attributes
```

gastmaier:
there seem to be also a dangerous `be:S16/16` (with ` -u ip:`) and ` be:s16/16` (without) `S` casing mismatch

dNechita:
> ok, so like #1368 it is an explicit flag, we offload the responsibility to gui apps, pyadi-iio
> 
> I didn't get the expected behavior :/
> 
> verified:
> 
> ```
> $ iio_attr -h | grep refre -A 1
>         -r, --refresh
>                         Refresh channel format(s). Device must be buffer-capable. Channel must be scan element.
> $ iio_attr -u ip:10.44.3.94 -c iio:device0
> dev 'ad4062', channel 'voltage0' (input, index: 0, format: be:S16/16>>0), found 5 channel-specific attributes
> 
> $ iio_attr -u ip:10.44.3.94 -c iio:device0 voltage0
> dev 'ad4062', channel 'voltage0' (input), attr 'calibscale', value '1.000000000'
> dev 'ad4062', channel 'voltage0' (input), attr 'oversampling_ratio', value '1'
> 
> $ iio_attr -u ip:10.44.3.94 -c iio:device0 voltage0 oversampling_ratio 2 
> 
> # Old value (expected)
> $ iio_attr -u ip:10.44.3.94 -c iio:device0
> dev 'ad4062', channel 'voltage0' (input, index: 0, format: be:S16/16>>0), found 5 channel-specific attributes
> 
> $ iio_attr -u ip:10.44.3.94 -r iio:device0
> Refreshed 1 channel format on device ad4062
> 
> $ iio_attr -u ip:10.44.3.94 -r iio:device0 voltage0
> # Wirdly no output
> 
> # Old value still (wrong)
> $ iio_attr -u ip:10.44.3.94 -c iio:device0
> dev 'ad4062', channel 'voltage0' (input, index: 0, format: be:S16/16>>0), found 5 channel-specific attributes
> ```
> 
> At the target
> 
> ```
> #  iio_attr -c iio:device0
> dev 'ad4062', channel 'voltage0' (input, index: 0, format: be:s20/32>>0), found 5 channel-specific attributes
> ```

I am just realising that the problem is a design problem. By this I mean that -r command does get the new type but once a new iio_attr instance is started it will still get the old iio context from the server because that is what the server does, at startup it scans the sysfs and builds an iio context which stores it in an XML format and will also compress it with Zstd library and during its lifetime it will serve the same context to any client that connects to it.

One idea would be to decompress the context, edit it (update only channel type information) and compress it back and during this time to make sure no client asks for the context. This wouldn't be my first option.

Another approach, discussed in #1368, would be to have iio_channel_get_data_format() always retrieve the data format from sysfs. With this in place, we could eventually deprecate the cached data format information currently embedded in the XML.

nunojsa:
Just some random idea :)...

Why not treating this just another channel attribute read? In the end of the day format is just another scan element attribute. Not 100% sure of all the implications but you would avoid a new kind of message. Basically all backends would automatically be able to read it without any new code. OTOH, I guess we would still likely need an explicit API (which would be just a wrapper around the normal attr read) to make the intent clear about what's being done and to have a place to properly document the importance of refreshing the _type attr in some systems.

dNechita:
> Just some random idea :)...
> 
> Why not treating this just another channel attribute read? In the end of the day format is just another scan element attribute. Not 100% sure of all the implications but you would avoid a new kind of message. Basically all backends would automatically be able to read it without any new code. OTOH, I guess we would still likely need an explicit API (which would be just a wrapper around the normal attr read) to make the intent clear about what's being done and to have a place to properly document the importance of refreshing the _type attr in some systems.

I previously explored this idea. One advantage is that it would not require introducing a new message (i.e. an IIOD command) and could instead leverage the existing libiio infrastructure for reading attributes.

However, the main drawback is that, in a client-server configuration, the client could issue a request to refresh the channel data format, but the server would be unable to distinguish it from a regular attribute read. As a result, the server would not update its own state, leading to a desynchronisation between the client and the server. The server uses the format of the channels to demultiplex data.
Having an IIOD command ensures that whenever the client asks for a refresh the server knows about it and can update it's own set of channel format data. But this approach isn't sufficient. If there are 2 or more clients connected to the same server, those clients that did not request a refresh will stil use the outdated channel data format.

bindings/csharp/Channel.cs
dNechita: The term "format" is already present in the API and in the bindings. Since the refresh function updates most of the members of C struct `iio_data_format` (7 out of 10) I would keep the name "refresh + format" so that there is a logical connection between `iio_channel_get_data_format` and `iio_channel_refresh_format`. The first is a getter and the second reads from hardware and refreshes the cached data.

doc/source/tools/iio_attr.rst
dNechita: This is not longer applicable. All utils now display an update version of the channel scan type.

utils/iio_attr.c
dNechita: You are right. I removed the -v entirely. It's no longer needed.

dNechita:
> there seem to be also a dangerous `be:S16/16` (with ` -u ip:`) and ` be:s16/16` (without) `S` casing mismatch

capital 'S' should no longer show up. Solved.

dNechita:
V2:

- Do a read of the channel scan type upon context creation, on the same place where channel scale and offset are being read.
- Remove `-v` command line option from `iio_attr` as it's no longer needed. The above change ensures iio_attr (and iio_info for that matter or any newly created libiio clients) displays the updated channel scan type.
- Fix displaying upper case 'S' on remote clients and lower case 's' on local target. Always show lower case 's'.

gastmaier:
@mhennerich for
> Fix displaying upper case 'S' on remote clients and lower case 's' on local target. Always show lower case 's'.

It is still a result of me misunderstanding it right? We should keep 'S' uppercase on libiio despite not being on linux upstream?

dNechita:
> @mhennerich for
> 
> > Fix displaying upper case 'S' on remote clients and lower case 's' on local target. Always show lower case 's'.
> 
> It is still a result of me misunderstanding it right? We should keep 'S' uppercase on libiio despite not being on linux upstream?

No, you noticed well that 'S' was being printed only on remote. This was added accidentally by me in this PR so I reverted that change.
But libiio continues to support both 's' and 'S'.

gastmaier (approved):
Hi Dan, confirmed the changes works

```
$ iio_attr -u ip:10.44.3.94 -c iio:device0
dev 'ad4050', channel 'voltage0' (input, index: 0, format: le:s12/32>>0), found 6 channel-specific attributes
$ iio_attr -u ip:10.44.3.94 -c iio:device0 voltage0
dev 'ad4050', channel 'voltage0' (input), attr 'calibscale', value '1.000000000'
dev 'ad4050', channel 'voltage0' (input), attr 'oversampling_ratio', value '1'
dev 'ad4050', channel 'voltage0' (input), attr 'oversampling_ratio_available', options '1' '2' '4' '8' '16' '32' '64' '128' '256'
dev 'ad4050', channel 'voltage0' (input), attr 'raw', value '0'
dev 'ad4050', channel 'voltage0' (input), attr 'sampling_frequency', value '2000000'
dev 'ad4050', channel 'voltage0' (input), attr 'scale', value '2.441406250'
$ iio_attr -u ip:10.44.3.94 -c iio:device0 voltage0 oversampling_ratio 2
2
$  iio_attr -u ip:10.44.3.94 -c iio:device0
dev 'ad4050', channel 'voltage0' (input, index: 0, format: le:s14/32>>0), found 6 channel-specific attributes
```

commit: e05c11f641185ed8ceca52f183ad024f5984c1e8

## GUI apps state:

for iio-osc, with oversampling 2 (12->14):

image

then on opening again:
image

that is, the gui app itself don't pick-up the scan type again, "wrapping" 2 bits, unless the user restarts the gui app.
(key change: no restart iiod needed anymore)

(scopy behaviour not checked)

nunojsa:
> However, the main drawback is that, in a client-server configuration, the client could issue a request to refresh the channel data format, but the server would be unable to distinguish it from a regular attribute read. As a result, the server would not update its own state, leading to a desynchronisation between the client and the server. The server uses the format of the channels to demultiplex data.
Having an IIOD command ensures that whenever the client asks for a refresh the server knows about it and can update it's own set of channel format data. But this approach isn't sufficient. If there are 2 or more clients connected to the same server, those clients that did not request a refresh will stil use the outdated channel data format.

Not really. If we treat `_type` just as a regular attr, every time we read it, we should get the real value (so I mean, no caching at all). And we could just get the string as-is. Then we can have some new APIs on top of the "regular" read to give the special treatment the attribute needs. It can have other drawbacks though... Or be a bigger change (more refactors) but (if I'm not missing nothing) I believe that in the end would be an easier change.

dNechita:
> > However, the main drawback is that, in a client-server configuration, the client could issue a request to refresh the channel data format, but the server would be unable to distinguish it from a regular attribute read. As a result, the server would not update its own state, leading to a desynchronisation between the client and the server. The server uses the format of the channels to demultiplex data.
> > Having an IIOD command ensures that whenever the client asks for a refresh the server knows about it and can update it's own set of channel format data. But this approach isn't sufficient. If there are 2 or more clients connected to the same server, those clients that did not request a refresh will stil use the outdated channel data format.
> 
> Not really. If we treat `_type` just as a regular attr, every time we read it, we should get the real value (so I mean, no caching at all). And we could just get the string as-is. Then we can have some new APIs on top of the "regular" read to give the special treatment the attribute needs. It can have other drawbacks though... Or be a bigger change (more refactors) but (if I'm not missing nothing) I believe that in the end would be an easier change.

Is this similar to what you had in mind? https://github.com/analogdevicesinc/libiio/pull/1512

dNechita:
After comparing both approaches:
 - approach 1 - this PR
 - approach 2  - PR #1512

, it seems that neither comes free of disadvantages.
I will not go with approach 2 because it affects the overall timing of the data capture/data transmit path. For each block created there needs to be N read operations of the "type" attribute, where N = number of channels.
If I think of the 80-20 rule, I think that scenarios where "type" changes dynamically falls into the 20%. Therefore, I wouldn't add a performance penalty for the rest of 80%. At least not in this libiio version given it's current architecture.

So, I am leaving the decision to re-read the channel format to the client code, which in most cases is aware of the hardware in use and, more importantly, knows when a configuration change may affect the channel data format.

PR #1494, opened 2026-07-09
Fix buffer detection for v0.x server with v1.0 client

dNechita:
## PR Description

Fixes compatibility issue where v1.0 clients connecting to v0.x servers
reported "0 buffers found" even though devices were buffer-capable.

**Root cause:** Legacy compatibility code only created buffers when
`<buffer-attribute>` tags existed. v0.x servers send channels with
`<scan-element>` tags but no explicit `<buffer>` tags, so no buffer
was created.

**Fix:** Enhanced both xml and emu backends to:
- Check if channels have scan-elements
- Create legacy buffer if needed
- Add all scan-elements to the buffer

## PR Type

## PR Checklist

### DIFF
--- a/emu.c
+++ b/emu.c
@@ -792,12 +792,34 @@ static int create_device(struct iio_context *ctx, xmlNode *n)
 		}
 	}
 
+	/* Handle legacy XML formats that have scan-elements but no explicit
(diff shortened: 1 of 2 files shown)


999Marius (approved):
lgtm

PR #1496, opened 2026-07-09
Fix Python bindings on embedded systems without dev tools

### LINKED ISSUE #1379: python bindings fail to find libiio.so on (embedded) systems without gcc, objdump, ldconfig and ld
libiio python bindings use ctypes `find_library()` helper function to get the correct library name on all platforms. On linux this relies on external tools being present as documented [here](https://docs.python.org/3/library/ctypes.html#finding-shared-libraries):

> On Linux, [find_library()](https://docs.python.org/3/library/ctypes.html#ctypes.util.find_library) tries to run external programs (/sbin/ldconfig, gcc, objdump and ld) to find the library file. It returns the filename of the library file.

These external programs are not guaranteed to be present on i.e. an embedded system built with buildroot. The only tool out of these you could possible have on the target system is ldconfig (as development tools like gcc or binutils are [not supported](https://buildroot.org/downloads/manual/manual.html#faq-no-compiler-on-target)) but that is an optional component that's only available for glibc based systems and only if it's part of the toolchain package the system is built with.

The result is that doing `import iio` just fails with the following backtrace:
```
>>> import iio
    import iio
  File "/usr/lib/python3.13/site-packages/iio.py", line 229, in <module>
  File "/usr/lib/python3.13/ctypes/__init__.py", line 412, in __getattr__
  File "/usr/lib/python3.13/ctypes/__init__.py", line 417, in __getitem__
AttributeError: python: undefined symbol: iio_get_backends_count
```

I've submitted a patch for this to buildroot just now: https://patchwork.ozlabs.org/project/buildroot/patch/20251222172638.839720-1-buildroot@bubu1.eu/ which just replaces the `find_libary()` call with the hardcoded `libiio.so` return value. This is always going to be correct, at least on a buildroot built system and probably on most linux distributions.

But having this working upstream would be a lot better than having to carry this downstream patch, so I wanted to ask your opinion on solving this upstream here. Maybe if `find_library()` returns `None` and we are on a linux platform, we can just try to load the canonical library name as a fallback (i,e, libiio.so and now also libc.so on git main)?

dNechita:
## PR Description

The Python bindings use ctypes.util.find_library() to locate libiio, which relies on external programs (ldconfig, gcc, objdump, ld) that are often absent in minimal embedded environments like buildroot or Yocto. When these tools are missing, find_library() returns None, causing the bindings to fail with "undefined symbol" errors.

This commit adds a fallback mechanism that activates only when find_library() returns None on Linux systems. It attempts to load the library directly using canonical soname patterns (libiio.so.1, libiio.so), validating each candidate before accepting it.

Closes #1379.

## PR Type

## PR Checklist

### DIFF
--- a/bindings/python/iio/__init__.py
+++ b/bindings/python/iio/__init__.py
@@ -328,6 +328,21 @@ if "Windows" in _system():
 else:
     # Non-windows, possibly Posix system
     _lib_loc = find_library("iio")
+
+    if _lib_loc is None:
+        # find_library failed - try direct loading for embedded systems
+        # (common on buildroot/yocto systems without gcc/ldconfig)
+        import sys
+        if sys.platform.startswith('linux'):
+            for name in ['libiio.so.1', 'libiio.so']:
+                try:
+                    # Attempt load to verify it's valid
+                    _cdll(name)
+                    _lib_loc = name
+                    break
+                except OSError:
+                    continue
+
     if _lib_loc is not None:
         if _path.islink(_lib_loc):
             _lib_loc = _path.realpath(_lib_loc)


999Marius (approved):
lgtm

PR #1507, opened 2026-07-17
iiod-client: fix reg_write not waiting for its actual response

999Marius:
## PR Description

`iiod_client_reg_write()` sent its command but never actually waited for the server's response - it only confirmed the outgoing write was flushed, so it never registered as a reader for the reply. That unclaimed response could then be consumed by whatever unrelated command issued the next request on the shared `default_io` channel, corrupting that command's result.

### Changes
- Replace the manual `iiod_io_send_command()` + `iiod_io_wait_for_command_done()` sequence in `iiod_client_reg_write()` with `iiod_io_exec_command()`, which registers for and waits on the matching response

### How it works
1. `iiod_client_reg_write()` builds the `IIOD_OP_REG_WRITE` command and value buffer as before.
2. It now calls `iiod_io_exec_command(io, &cmd, &buf, NULL)`, matching the pattern already used by `iiod_client_reg_read()`.
3. `iiod_io_exec_command()` registers as the reader before sending, so the reply is always matched to the call that issued it.

### Testing
Verified against real M2k hardware (`ad9963`): wrote/read back a register 50 times, including back-to-back writes immediately followed by an unrelated attribute read (the interleaving that previously risked misdelivery). All 50 iterations passed with correct values and no mismatches.

## PR Type

## PR Checklist

### DIFF
--- a/iiod-client.c
+++ b/iiod-client.c
@@ -1932,7 +1932,6 @@ int iiod_client_reg_write(struct iiod_client *client, const struct iio_device *d
 	struct iiod_io *io = iiod_responder_get_default_io(client->responder);
 	struct iiod_command cmd;
 	struct iiod_buf buf;
-	int ret;
 
 	cmd.op = IIOD_OP_REG_WRITE;
 	cmd.dev = (uint8_t)iio_device_get_index(dev);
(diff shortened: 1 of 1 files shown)


dNechita (approved):
LGTM!

PR #1510, opened 2026-07-23
bindings: python: fix iio_device_get_trigger for v1.0 API

MaureenHelm:
The v1.0 C API changed iio_device_get_trigger to return the trigger device pointer directly (NULL if none, error-encoded pointer on failure) instead of writing to an out-parameter. Update the Python binding to match.

Assisted-by: Claude:claude-opus-4.6

## PR Type

## PR Checklist

### DIFF
--- a/bindings/python/iio/__init__.py
+++ b/bindings/python/iio/__init__.py
@@ -80,6 +80,12 @@ def _check_ptr_err(result, func, arguments):
     return result
 
 
+def _check_ptr_err_allow_null(result, func, arguments):
+    if result is None:
+        return result
+    return _check_ptr_err(result, func, arguments)
+
+
 # pylint: enable=unused-argument
 
 # Python 2 and Python 3 compatible _isstring function.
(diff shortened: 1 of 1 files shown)


bindings/python/iio/__init__.py
```diff
@@ -599,12 +599,8 @@ class ChannelType(Enum):
...
 
 _d_get_trigger = _lib.iio_device_get_trigger
-_d_get_trigger.restype = c_int
-_d_get_trigger.argtypes = (
-    _DevicePtr,
-    _POINTER(_DevicePtr),
-)
-_d_get_trigger.errcheck = _check_negative
+_d_get_trigger.restype = c_void_p
+_d_get_trigger.argtypes = (_DevicePtr,)
```
dNechita, line 609: Good catch!
To avoid duplicating the logic that subtracts 4096 from the error code, you can reuse the existing `_check_ptr_err` function by adding:
`_d_get_trigger.errcheck = _check_ptr_err`

There are several function in the API than can return an encoded error pointer, such as:
`iio_create_context`
`iio_device_create_event_stream`
`iio_buffer_create_stream`
etc.,
 and all of these use `_check_ptr_err`.

dNechita (approved):
Looks good!

PR #1513, opened 2026-07-31
VITA 49.2 Base Functionality

WithanagePerera:
Adds support for Data Packets, Control Packets, Control Extension Packets, Context Packets, AckX/AckV/AckS Packets.

Also creates a VITA 49.2 subsystem that gets spawned as a threadpool within IIOD. The subsystem is able to execute commands/queries for I/Q from a host machine over UDP and send the relevant V49.2 packet back.

Still in progress.


tfcollins (changes requested):
Can you remove the "pylibiio.egg-info" folder

PR #1447, opened 2026-05-13
Add emu backend test variants and improve coverage

999Marius:
## PR Description

This PR extends the libiio test suite so that every existing API test is also ran against the emu backend, running entirely in CI without any physical hardware. Each API test is registered as an additional CTest variant that points `TESTS_API_URI` at a device XML from `tests/resources/xmls/`. CTest labels (`api`, `hardware`, `emu`, `<xml-name>`) make it easy to run or filter.

Also adds IST to the codespell ignore list to suppress a false positive in the fmcomms2.xml device description string.

### Benefits

- Runs the full API test suite against emulated hardware profiles in GitHub Actions without needing any physical devices.
- Increases GitHub Actions CI function coverage from **31.5% (206/653)** to **48.8% (336/688)** -> +17.3 gain
- Increases GitHub Actions CI line coverage from **20.7% (1719/8307)** to **34.6% (3099/8957)** -> +13.9 gain
   - Here is a direct link to the 2 coverages [before emu tests](https://github.com/analogdevicesinc/libiio/actions/runs/25796230850/job/75774120153#step:6:335) => [after emu tests](https://github.com/analogdevicesinc/libiio/actions/runs/25800687209/job/75789584505?pr=1447#step:6:394)
- Replaces `pluto.xml` with `fmcomms2.xml` across tests to use a device profile that is already present in the CI XML set and is generated with libiio v1 semantics

### Changes

- Add `-DWITH_EMU_BACKEND=ON` to the `cmake` invocation in `.github/workflows/tests.yml` so emu test variants are built and executed by `make coverage`
- Add `LABELS "api"` to all API CTest entries in `tests/api/CMakeLists.txt` for selective filtering
- Add emu CTest variants in `tests/api/CMakeLists.txt`: for each XML file under `tests/resources/xmls/` and each API test target, register `${test}_emu_${xml_name}` with `TESTS_API_URI=emu:${xml_file}` and labels `"emu;${xml_name}"`
- Add `LABELS "hardware"` to all hardware CTest entries in `tests/hardware/CMakeLists.txt`
- Update `bindings/csharp/tests/IntegrationTests.cs` and `bindings/csharp/tests/README.md` to use `fmcomms2.xml`

### How it works

1. When `WITH_EMU_BACKEND=ON` is passed to CMake, `tests/api/CMakeLists.txt` globs all `*.xml` files under `tests/resources/xmls/`.
[...]

PR #1515, opened 2026-08-07
tests: api: skip instead of crashing on null context

DanielMBouyou:
When libiio is built without the local backend and TESTS_API_URI is unset, three api tests (test_context, test_channel, test_lowlevel) dereference a NULL context instead of skipping: two segfault, one fails an assertion. This makes their skip guards work.

Why the guards were dead: create_test_context() normalises every failure to NULL, it returns NULL whenever iio_err() is set and otherwise returns the context (tests/test_helpers.h:22-27). So by the time a test body runs, iio_err(ctx) is always 0, the if (iio_err(ctx)) guards could never fire, and the code fell through to a NULL dereference. The fix guards on the pointer instead (if (!ctx) / if (!test_ctx)), the skip idiom the other api tests already use.

The skip policy these tests were meant to honour was added in #1447 (May 2026): "Skip tests if no backend available ... This ensures tests run in environments with varying backend support" (tests/api/README.md:104-115). That PR added the README and the emu variants but not the guards in these three bodies.

Verified in a container build with no local backend: the three now skip cleanly instead of two segfaults and one failure. CI config (WITH_LOCAL_BACKEND=ON): full suite still 49/49, zero added skips. Emu subset: 36/36.

setup_test_channel() keeps a fourth, redundant iio_err() check on the same normalised return; harmless, left as-is, happy to clean it up in a follow-up.

Not tested: Linux x86_64 in a container only, no real hardware.

## PR Type

## PR Checklist

### DIFF
--- a/tests/api/test_channel.c
+++ b/tests/api/test_channel.c
@@ -190,6 +190,11 @@ TEST_FUNCTION(channel_mask_operations)
 {
 	setup_test_channel();
 
+	if (!test_ctx) {
+		DEBUG_PRINT("  SKIP: No test context available\n");
+		return;
+	}
+
 	/* Find a channel that is a scan element in order to test enabling/disabling */
 	struct iio_channel *iio_chn = NULL;
 	unsigned int nb_devices = iio_context_get_devices_count(test_ctx);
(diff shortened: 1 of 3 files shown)


999Marius (approved):
The fixes look good and I've tested them myself as well, solid root cause writeup too.
Thanks for contributing!

dNechita (approved):
Thanks for the fix!

PR #1530, opened 2026-08-24
usb: fix teardown hang when cancellation races transfer submission

MaureenHelm:
## PR Description

Destroying a USB context intermittently hangs. All I/O completes, then iio_context_destroy() never returns: the iiod-responder reader thread sits in an 8-byte bulk IN read and the main thread blocks in the join inside iiod_responder_wait_done().

That read carries a timeout_ms of -1, which usb_sync_transfer() maps to a libusb timeout of 0, meaning wait forever, so the only way out of it is libusb_cancel_transfer(). iiod_responder_stop() merely sets a flag and cannot interrupt a read already in progress, and usb_cancel() can only cancel a transfer that is already in flight. When cancellation lands in the window before the reader submits, nothing interrupts the transfer it submits next: the cancelled flag is consulted before submitting and never again once the completion loop is entered.

Wait for completion in bounded steps and re-check the flag, cancelling the in-flight transfer when cancellation was requested after it was submitted. Cancellation no longer depends on where the reader happens to be.

Reproducing it needs a context with more than one device, so that enough command exchanges occur to fall in the window; a single-device context never hit it. Over repeated iio_info runs against a two-device target the hang appeared in four of six attempts before and in none of twenty-four after.

Assisted-by: Claude:claude-opus-5

cc: @ralucabozdog 

## PR Type

## PR Checklist

### DIFF
--- a/usb.c
+++ b/usb.c
@@ -607,6 +607,7 @@ static int usb_sync_transfer(struct iio_context_pdata *pdata, struct iiod_client
 {
 	unsigned char ep;
 	struct libusb_transfer *transfer = NULL;
+	bool cancel_requested = false;
 	int completed = 0;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


MaureenHelm:
Superseded by #1532

PR #1541, opened 2026-08-28
compat, iiod: fix buffered I/O for v0 applications

999Marius:
## PR Description

Three fixes to the v0 compat path: two in the compat layer, one in iiod's v0 protocol support.

### Benefits

- libm2k works through the compat layer at all — it calls `iio_channel_write()` on every push, which failed to resolve
- Input buffers work with `nb_kernel_buffers` set to 1, which previously could not read a sample
- Fixes the Scopy Signal Generator freeze on M2K: Run hung the UI and left the DAC at `-EBUSY` until iiod restarted

### Changes

- Add `iio_channel_read()`, `iio_channel_read_raw()`, `iio_channel_write()`, `iio_channel_write_raw()`
- Queue every input block in `iio_device_create_buffer()` before `iio_buffer_stream_start()`, matching `stream.c`; set `compat->enqueued` so `refill()` doesn't queue block 0 twice
- Restrict `rw_thd()`'s empty-block pre-enqueue to input buffers
- Add `tx_all_enqueued` to `struct DevEntry`, guarding `iio_block_dequeue()` on output buffers
- Skip the dequeue for cyclic transfers, which have no completion callback
- Signal writer threads as well as readers on completion

### Backwards Compatibility

No API or ABI change — four unresolved functions added, none changed. TX still keeps block 0 for the application to write. The iiod change drops a transfer that was never meaningful.

### Testing

| test | before | after |
|---|---|---|
| any libm2k use | `undefined symbol: iio_channel_write` at import | works |
| Scopy Signal Generator Run | UI freeze, DAC stuck at `-EBUSY` | works |
| `test_pattern_generator_pulse` | hangs forever | passes |
| `test_noncyclic_buffer` | hangs forever | passes, 0.6 s |

- v0 ABI now complete: 147 symbols, no gaps against a real 0.26 build
- `iio_info`, `iio_attr`, `iio_genxml` byte-identical to the pre-fix layer across 9 XML contexts; `iio_attr` matches 0.26 in all four modes
- `dummy-iiostream`, `iio_readdev` and `iio_writedev` from the v0 branch, cyclic and non-cyclic
- `ad9361-iiostream` on an FMCOMMS2: simultaneous RX+TX, non-cyclic 1 MiS buffers, ~46 MSmp each way over 20 s through the compat layer and ~48 MSmp with a real 0.26 client
- builds clean with `LIBIIO_COMPAT`, `WITH_IIOD` and `WITH_IIOD_V0_COMPAT` each on and off
- libm2k suite: 70 pass, 1 fail, 4 error, 1 skip of 76; the failures are pre-existing

Also wrote a v0-API hardware harness, run two ways — through the compat layer against a stock 1.x iiod, and with a real 0.26 client against a patched iiod, since only the latter reaches `rw_thd()`. It covers: RX refill with `nb_kernel_buffers` at 1 and at the default, all four new channel helpers, TX non-cyclic across more cycles than there are kernel buffers, TX cyclic start/teardown/reopen, and an induced dequeue failure checking the daemon is not left wedged.

## PR Type

## PR Checklist

### DIFF
--- a/compat.c
+++ b/compat.c
@@ -140,6 +140,11 @@ struct compat {
 	void (*iio_channel_convert)(const struct iio_channel *, void *, const void *);
 	void (*iio_channel_convert_inverse)(const struct iio_channel *, void *, const void *);
 
+	size_t (*iio_channel_read)(const struct iio_channel *, const struct iio_block *,
+			void *, size_t, bool);
+	size_t (*iio_channel_write)(const struct iio_channel *, struct iio_block *,
+			const void *, size_t, bool);
+
 	void (*iio_channel_enable)(const struct iio_channel *, struct iio_channels_mask *);
 	void (*iio_channel_disable)(const struct iio_channel *, struct iio_channels_mask *);
 	bool (*iio_channel_is_enabled)(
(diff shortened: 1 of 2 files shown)


dNechita (approved):
Looks good!

PR #1547, opened 2026-09-01
zephyr: iiod: usb: Give the IIO class its own buffer pools

MaureenHelm:
## PR Description

The class took its bulk buffers from udc_ep_pool, which is a single allowance for the whole device: every class on it, both directions, plus control transfers. Its 1024-byte default cannot arm three 512-byte pipes, so pipes 1 and 2 never armed and the host timed out connecting to a data pipe.

Define an RX and a TX pool per instance instead, sized from the same devicetree properties that size the endpoints: one receive buffer per pipe, and one transmit chunk per pipe since iiod_usb_pipe_write() sends them one at a time. This is what the CDC ACM class does for the same reason, and it costs less than raising CONFIG_UDC_BUF_POOL_SIZE would - 1728 bytes rather than 3072 - while leaving the shared pool to the rest of the device.

Assisted-by: Claude:claude-opus-5

@dimitrije-lilic @ralucabozdog please give this a try for the USB "failed to allocate" error you observed on pipe 1.

## PR Type

## PR Checklist

### DIFF
--- a/zephyr/iiod/usb.c
+++ b/zephyr/iiod/usb.c
@@ -124,6 +124,8 @@ struct iio_usb_data {
 	uint16_t tx_buf_size;
 	uint16_t rx_fifo_size;
 	uint8_t *rx_fifo_data;
+	struct net_buf_pool *rx_pool;
+	struct net_buf_pool *tx_pool;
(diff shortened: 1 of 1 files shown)


dNechita (approved):
Looks good.

PR #1550, opened 2026-09-02
.github/docs: dirhtml instead of html, documentation -> system-level

gastmaier:
To not require the .html, use Sphinx dirhtml suffix that wraps
non-index pages into subdirectories, for example:

  software/app/cli.rst -> software/app/cli/
  
To consolidate with the developer.analog.com/docs name,
rename documentation to system-level.


dNechita (approved):
Looks good! Thanks!

PR #1553, opened 2026-09-03
zephyr: Add Zephyr implementation of the libiio lock and thread API

MaureenHelm:
## PR Description

Zephyr builds compiled `lock-dummy.c`, so every `iio_mutex_lock()` was an empty function, `iio_cond_wait()` returned `-ETIMEDOUT` immediately and `iio_thrd_create()` returned `-ENOSYS`. The network and USB transports already run one iiod interpreter per client over a shared context, mutating the file-scope buffer and event-stream lists in `iiod/responder.c` with their locks compiled out.

Add a `k_mutex`/`k_condvar`/`k_thread` backend behind a new `CONFIG_LIBIIO_MULTITHREADING`, default `n` so existing builds are unchanged. The option switches the lock backend and `NO_THREADS` together: the `NO_THREADS` paths in `task.c` run task bodies inline and release a mutex they do not hold, which is only harmless while the mutexes are no-ops.

Nothing here allocates. Mutex, condvar and thread objects come from static pools with sys_bitarray slot tracking, following the idiom Zephyr's POSIX layer uses for `pthread_mutex_t` and `pthread_cond_t`: `k_mutex_init()` and `k_condvar_init()` run on allocation rather than once at boot, so a slot released while its k_mutex is still held cannot hand the next caller a mutex that still records the old owner. Exhaustion returns `-ENOMEM` and logs which pool ran out. The pool sizes are Kconfig, defaulting higher for the network server, which serves several clients at once.

The 4096 stack default comes from measurement: static analysis puts the deepest path near 1.8 KB on Cortex-M4, and peaks on apard32690 serving an attribute read are 616 bytes for the reader thread and 320 for the writer. The headroom is for the buffer paths still to land.

### Testing

Tested with the network transport on native_sim and the USB transport on apard32690.

Below is a SystemView trace of an `iio_info` transaction (partial) over the USB transport on apard32690. Note the `reader-thd` (purple) and `writer-task` (blue), and `iiod_usb_0` (green) threads.
image

## PR Type

## PR Checklist

### DIFF
--- a/iiod/responder.c
+++ b/iiod/responder.c
@@ -1350,10 +1350,12 @@ static void iiod_responder_free_resources(struct parser_pdata *pdata)
 int binary_parse(struct parser_pdata *pdata)
 {
 	struct iiod_responder *responder;
+	int ret;
 
 	responder = iiod_responder_create(&iiod_responder_ops, pdata);
-	if (!responder)
-		return -ENOMEM;
+	ret = iio_err(responder);
+	if (ret)
+		return ret;
 
 	/* TODO: poll main thread pool FD */
 
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -60,7 +60,7 @@ jobs:
           if [ "${{ runner.os }}" = "Windows" ]; then
             EXTRA_TWISTER_FLAGS="--short-build-path -O/tmp/twister-out"
           fi
-          west twister -T samples -v --force-color --inline-logs --integration \
+          west twister -T samples -T tests -v --force-color --inline-logs --integration \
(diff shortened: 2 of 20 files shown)


iiod/responder.c
```diff
@@ -1350,10 +1350,12 @@ static void iiod_responder_free_resources(struct parser_pdata *pdata)
 int binary_parse(struct parser_pdata *pdata)
 {
 	struct iiod_responder *responder;
+	int ret;
 
 	responder = iiod_responder_create(&iiod_responder_ops, pdata);
-	if (!responder)
-		return -ENOMEM;
+	ret = iio_err(responder);
```
dNechita, line 1356: That's a good find!

dNechita (approved):
LGTM

PR #1556, opened 2026-09-08
ci: fix format.sh auto-running on source in FormatCheck

999Marius:
The Azure `FormatCheck` job sources `format.sh` to reuse its helper functions, but `format_all()` was being called unconditionally at the bottom of the script. 
Sourcing therefore auto-formatted every file in the PR checkout in place before the dry-run check ran, so real formatting issues never made the check fail.

### Changes
- Guard the `format_all` call in `format.sh` to only run when the
  script is executed directly, not when sourced

## PR Type

## PR Checklist

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -61,4 +61,6 @@ format_all() {
     done
 }
 
-format_all
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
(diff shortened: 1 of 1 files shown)


dNechita (approved):
good one! Thanks!

PR #1557, opened 2026-09-08
Zephyr Thread Improvements

MaureenHelm:
## PR Description

This branch fixes a crash in the IIOD responder's command-enqueue error path and then reduces the thread and stack resources the Zephyr IIOD responder needs per connection.

The fix (`treewide: don't store a failed iio_task_enqueue() result` + its regression test) addresses a bug where a failed enqueue left a poisoned pointer in a field that callers treat as a plain liveness check, leading to a stale-pointer dereference and permanent wedging of the connection; this is reachable on Zephyr whenever `CONFIG_LIBIIO_COND_POOL_SIZE` is exhausted and was confirmed causing a bus fault on apard32690 over USB and a segfault on native_sim.

The remaining commits eliminate an unnecessary per-connection reader thread on the server side (the IIOD daemon and Zephyr transports only ever block waiting for the connection to end, so they don't need an async reader), resize the Zephyr thread pool defaults and test expectations to match the new one-thread-per-connection model, and shrink the network client thread's stack to a size measured via GCC call-graph analysis and cross-checked at runtime on hardware.

Net effect on Zephyr: each IIOD connection now costs one thread pool slot instead of two, and the network configuration's RAM footprint on apard32690 drops from 96% to 77% of the target's 128 KB.

## PR Type

## PR Checklist

### DIFF
--- a/block.c
+++ b/block.c
@@ -143,6 +143,8 @@ int iio_block_enqueue(struct iio_block *block, size_t bytes_used, bool cyclic)
 	struct iio_buffer_stream *buf_stream = block->buf_stream;
 	const struct iio_device *dev = buf_stream->buf->dev;
 	const struct iio_backend_ops *ops = dev->ctx->ops;
+	struct iio_task_token *token;
+	int err;
 
 	if (bytes_used > block->size)
 		return -EINVAL;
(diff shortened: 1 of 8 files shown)


block.c
dNechita, line 1: This makes sense. Thanks!

dNechita (approved):

PR #1564, opened 2026-09-14
Add iiod-emu server utility

dNechita:
## PR Description

Add a cross-platform IIOD server based on libtinyiiod, called `iiod-emu`. Its purpose is to enable more advanced hardware emulation by allowing multiple clients to connect to the same server and share state, unlike the emu backend, which duplicates the state for each client.

This makes it possible to create hardware models (via dynamic backends) that better emulate real hardware, rather than representing only a snapshot of the state (the IIO attribute values at a given point in time).

Additionally, `iiod-emu` can be used for CI testing when real hardware is not available.

## PR Type

## PR Checklist


999Marius (approved):
Looks good!

PR #1569, opened 2026-09-16
CMake: Fix missing public headers in macOS framework bundle

999Marius:
## PR Description

macOS `iio.framework` shipped with 4 of 5 public headers missing (only `iio.h` present).

Cause: `add_library(iio ...)` never listed headers as target sources. CMake's Makefiles/Ninja generators only embed a `PUBLIC_HEADER` into a framework if it's also a source (see `ConstructSourceFileFlags()` in CMake's `cmGeneratorTarget.cxx` — it does `Makefile->GetSource(relFile)`, which returns null for unregistered files, so they're silently skipped). Xcode generator doesn't have this restriction, so it never showed there.

Fix: `target_sources(iio PRIVATE ${IIO_PUBLIC_HEADERS})` — one line.

Tested on macOS:
- `Headers/` empty without the line, all 5 present with it
- Confirmed in both plain build and CI-style packaged tarball
- Ran `ad9361-iiostream` example against fixed headers — works
- No effect on Linux (headers as non-compiled sources are a no-op there)

## PR Type

## PR Checklist


999Marius:
**Update:** Guarded the `target_sources()` call behind `OSX_FRAMEWORK`

dNechita (approved):

PR #1572, opened 2026-09-17
 Pin down the block contracts

dNechita:
## PR Description

`struct iio_block` is the object that carries sample data between libiio and the
application. Today every block libiio hands out is a chunk of memory that libiio
itself allocated and `mmap()`ed, so the pointer accessors — `iio_block_start()`,
`iio_block_first()`, `iio_block_end()` — always have something to return, and the
documentation says so unconditionally.

In the future this will stop being true.

### Background: the two directions of DMABUF sharing

A DMABUF is a buffer plus a file descriptor that acts as a handle to it. Handing
that descriptor to another subsystem lets it DMA into the *same* physical memory,
with no copy in between. There are two directions, and they are not symmetric:

* **libiio exports.** libiio allocates the buffer and hands out the descriptor
  via `iio_block_get_dmabuf_fd()`. Whoever imports it — a GPU, a NIC, another
  driver — attaches to memory libiio chose. This works today and needs nothing
  new.

* **libiio imports.** Someone else allocates the buffer and libiio attaches to
  it. This is the direction we do not support yet, and it is the direction some
  hardware forces on us. Where an allocator insists on owning its memory — CUDA
  on NVIDIA hardware being the case in front of us — libiio has no choice but to
  be the importer. That requires one new constructor, roughly
  `iio_buffer_create_block_from_dmabuf(buffer, fd, size)`, which is purely
  additive: a new symbol, no soname bump, and `struct iio_block` is private to
  `block.c` so its layout is ours to change.

## What this does
 - Updates the contract (API documentation) about `iio_block_start()`, `iio_block_first()` and `iio_block_end()` stating they can return an encoded error pointer when access to the block data is restricted.
 - Updates the bindings such that the return type is decoded

## PR Type

## PR Checklist


include/iio/iio.h
```diff
@@ -1308,9 +1366,25 @@ __api __check_ret ssize_t iio_block_foreach_sample(const struct iio_block *block
...
+/** @brief Get the amount of valid data in the block, in bytes
+ * @param block A pointer to an iio_block structure
+ * @return The number of bytes of valid data in the block, counted from the
+ * address returned by iio_block_start.
+ *
+ * <b>NOTE:</b> Before the block has been enqueued for the first time, this
+ * function returns zero. After a successful iio_block_enqueue, it returns the
+ * value that was passed to it, or the size of the block if that value was zero.
+ * An enqueue that fails leaves the previously reported value unchanged. */
+__api __pure size_t iio_block_get_bytes_used(const struct iio_block *block);
```
999Marius: `iio_block_get_bytes_used()` maybe shouldn't be `__pure` - `bytes_used` is mutated by `iio_block_enqueue()`, possibly from another thread. `pure` only holds for write-once fields (like `block->data` behind `iio_block_is_mapped`).  Suggest dropping `__pure` here.

dNechita: makes sense

include/iio/iio.h
```diff
@@ -1251,24 +1279,31 @@ __api int iio_block_disable_cpu_access(struct iio_block *block, bool disable);
...
  * the mapped data is currently coherent. See iio_block_disable_cpu_access. */
 __api __pure bool iio_block_is_mapped(const struct iio_block *block);
 
 /** @brief Get the start address of the block
  * @param block A pointer to an iio_block structure
- * @return A pointer corresponding to the start address of the block */
+ * @return A pointer corresponding to the start address of the block
+ * @return NULL if the block has no CPU mapping
+ *
+ * <b>NOTE:</b> Returning NULL is not an error condition. It indicates that the
```
nunojsa: We have a way to encode errors in pointers. I would use that (it just scales much better)

block.c
```diff
@@ -221,6 +224,10 @@ void *iio_block_first(const struct iio_block *block, const struct iio_channel *c
 	unsigned int i;
 	size_t len;
 
+	/* The block has no CPU mapping; there is no pointer to return. */
+	if (!block->data)
+		return NULL;
```
nunojsa: I'm not sure checking the pointer is the proper way for this! We have two things going:

1. As the API stands, mmap() cannot return NULL (it's an hard error anyways). What can happen is that cpu access is disabled (I actually think there's no internal path on libiio to give the user such data but we do have public API to disable access) but that not means data is NULL, just means is probably outdated/garbage (this cpu IOCTL are pretty much cache maintenance in the kernel to make sure data is up to date).
2. A future import API where data cannot be mapped at all! In that case, data will indeed be NULL but given 1. we should already have something to flag that we should not touch the data! In fact, in these scenario we can use both the pointer + a flag to detect misuse like "enabling cpu access on something that we have no mapping for" (also just realized that iio_block_disable_cpu_access() has no documentation for `return`).

To sum it up I think the most important thing now is to make sure APIs like iio_block_end() stay consistent in terms of behavior and we can ensure that with 1. already! But what we have now is cpu enabled/disabled state and so I would use that.

For 2. I think it can be added later without changing API behavior. At the very least, we can return an additional error code but I don't think that counts as API breakage

include/iio/iio.h
```diff
@@ -1225,24 +1225,52 @@ __api __check_ret struct iio_block *iio_buffer_stream_create_block(
...
  * @param block A pointer to an iio_block structure
  * @return The file descriptor of the underlying DMABUF object.
  * If the iio_block is not backed by a DMABUF object, -EINVAL is returned.
- * Otherwise, the file descriptor will be valid until the block is destroyed. */
+ * Otherwise, the file descriptor will be valid until the block is destroyed.
+ *
+ * <b>NOTE:</b> The returned file descriptor is owned by the iio_block. The
+ * caller must not close it; it is closed by iio_block_destroy. To retain a
+ * reference to the DMABUF object beyond the lifetime of the block, the caller
+ * should duplicate the file descriptor, e.g. with dup() or fcntl(F_DUPFD). */
```
nunojsa: I think the above is actually misleading. Not sure about the added value of doing `dup()`. Destroying the block is not just about closing a flie descriptor. I would drop the above for now.

include/iio/iio.h
```diff
@@ -1225,24 +1225,52 @@ __api __check_ret struct iio_block *iio_buffer_stream_create_block(
...
+ * iio_block_is_mapped. A block may therefore be in one of three states:
+ *
+ * * mapped and synchronized: the pointer accessors return valid pointers and
+ *   the data they refer to is coherent with the device;
+ * * mapped with CPU access disabled: the pointer accessors still return valid
+ *   pointers, but the data they refer to may be stale and must not be used;
+ * * not mapped: the block has no CPU mapping at all, the pointer accessors
+ *   return NULL, and iio_block_is_mapped returns false. This can happen for
+ *   blocks whose memory is not CPU-accessible, for instance memory that lives
+ *   behind a PCI BAR.
```
nunojsa: I would not document things we do not support as of now. As said, we care about behavior of the API which means we can return proper errors and not blindly access the data. A part from data I would keep docs synchronized with what we support today

999Marius (approved):
Looks good!

include/iio/iio.h
```diff
@@ -1311,6 +1325,16 @@ __api __check_ret ssize_t iio_block_foreach_sample(const struct iio_block *block
...
 
+/** @brief Get the amount of valid data in the block, in bytes
+ * @param block A pointer to an iio_block structure
+ * @return The number of bytes of valid data in the block, counted from the
+ * address returned by iio_block_start.
+ *
+ * <b>NOTE:</b> Before the block has been enqueued for the first time, this
+ * function returns zero. After iio_block_enqueue, it returns the value that
+ * was passed to it, or the size of the block if that value was zero. */
+__api size_t iio_block_get_bytes_used(const struct iio_block *block);
```
nunojsa: Don't think we should be adding any of the above APIs!

* `iio_block_is_mapped()` - there's no usecase for it as of today so adding it later should be perfectly fine
* `iio_block_get_bytes_used()` - did not looked much so maybe it makes sense but then It would be nice  to see some user of it rather than bindings. We have examples and tools so that one could see the added value of the API

nunojsa (commented):
Here it goes. There's things where I we're trying to guess how it will look like. Don't agree tbh. Yes, we should be careful and plan ahead about the APIs block accessor APIs. But as stated, AFAICT, we just need to make sure they don't change behavior and we can already get that done with what we have today without trying to guess how things will be in the future.

include/iio/iio.h
dNechita: but that changes behavior making the functions return different things (encoded error pointer instead of raw pointer). This also means the clients calling them need to change how the handle the returned value.

include/iio/iio.h
dNechita: Agree

include/iio/iio.h
dNechita: alright

include/iio/iio.h
nunojsa: Nope! Only changes if you change the error code. Adding new ones should not be a problem. Or?

dNechita: Adding new error codes should be a problem.
LATER EDIT: I made a typo, I meant it should not be a problem. 

It's just that changing the contract (until now it returns valid address or NULL and what you are suggesting is adding encoded pointer as well) makes these `iio_bllock_start()`, `iio_block_end()`, `iio_block_first()` a bit harder to use by needing to remember to also decode.

block.c
dNechita: Removing the NULL check.

include/iio/iio.h
dNechita: Removing `_is_mapped()` but keeping `get_bytes_used()`.

include/iio/iio.h
dNechita: It's probably better to wait and see how the future implementation will look like and then act upon it.

nunojsa: > It's just that changing the contract (until now it returns valid address or NULL and what you are suggesting is adding encoded pointer as well) 

Well, of course it changes behavior when compared to what we have today. I was just saying that adding a new error code (after we change to encoding errors in pointers) would not break the API. but given that we still did not released v1, we still have time (was kind of my point :)).

> makes these iio_bllock_start(), iio_block_end(), iio_block_first() a bit harder to use by needing to remember to also decode.

Sure but it's way more meaningful than NULL if we start to have different error conditions and we want to know what was the cause. So what seems likely to libiio is:

1. You can't access a block because cpu access is disabled (possible today)
2. You can't access a block because data was not mapped at all

How do we differentiate with NULL? That was the whole reasoning of "NULL as return does not scale very nicely). But I have no strong feelings as well!

nunojsa:
I think you dropped the most valuable change which is returning NULL (or a proper error - whatever you decide :)) on the block APIs! At the very least we should document the proper APIs that NULL can be returned if the memory cannot be accessed. It is already documented that one should not use that memory if `iio_block_disable_cpu_access()` is set on the block but having a more clear "return value kind of doc" in the  affected APIs would make the intent (that users should check for the returned pointer before using it) more clear. Like stating that NULL can be returned! Then we're free to make the change internally in the library even if we do not return NULL as of today. And I do think that having something like you had before on `block->data` but instead of data using a `cpu_access` flag (we already have such a flag in pdata) should be fairly straight. But improving docs (leaving things prepared for future) is arguably more important.

And FWIW, I still don't see the value in the new bytes_used API. Did anyone explicitly asked for it? Because looking at the current PR, the only users I see are bindings and unit test which do not count IMO.

include/iio/iio.h
dNechita: > Sure but it's way more meaningful than NULL if we start to have different error conditions and we want to know what was the cause

Now that you put it like that, yes, it makes sense to have a way to differentiate between cpus access disabled state and not mapped state.

dNechita:
Ready for reviewing.

include/iio/iio.h
```diff
@@ -1257,14 +1257,20 @@ __api int iio_block_disable_cpu_access(struct iio_block *block, bool disable);
 
 /** @brief Get the start address of the block
  * @param block A pointer to an iio_block structure
- * @return A pointer corresponding to the start address of the block */
+ * @return On success, a pointer corresponding to the start address of the block
+ * @return On failure, a pointer-encoded error is returned
+ *
+ * <b>NOTE:</b> Check the returned pointer with iio_err before dereferencing it.
+ * A block's memory is not guaranteed to be reachable by the CPU; memory that cannot
+ * be mapped, would be reported this way rather than as a valid address. */
```
nunojsa: a small nit: `mapped or accessed`

nunojsa (approved):
Looks good! Next (which should not stop the release) would be to check tools and examples to properly check for errors on  the block APIs. People might look at our examples so we better have good ones :)

include/iio/iio.h
dNechita: done

PR #1574, opened 2026-09-17
Library Symbol Version Map

dNechita:
## PR Description

Add a library symbol version map.
This would allow adding new functions post v1.0.0 more easily, at least on Linux.

Add some fixes that were found while implementing this.

## PR Type

## PR Checklist


library.c
```diff
@@ -26,11 +26,16 @@ static void libiio_exit(void)
...
-		platform_get_ticks_us = cb;
-	}
+	platform_get_ticks_us = cb;
+
+	/* Log timestamps are reported relative to library_startup_time_us,
+	 * which was read from the built-in clock when the library was loaded.
+	 * Re-base it onto the time source that is now in effect, otherwise
+	 * every timestamp would be the difference between two unrelated
+	 * clocks. */
+	library_startup_time_us = iio_read_counter_us();
```
999Marius: Guard was dropped in the rename, calling this again with the same cb now always resets `library_startup_time_us `. Would be safer to be guarded in the old ` if (platform_get_ticks_us != cb)`

nunojsa:
I would not leave those Claude tags! We (humans) should still be the ones responsible and signing of the committs. We coud like linux and have an `Assisted-by:` tag or a co-authored.

My 2 cents

library.c
dNechita: you're right. Fixed!

dNechita:
> I would not leave those Claude tags! We (humans) should still be the ones responsible and signing of the committs. We coud like linux and have an `Assisted-by:` tag or a co-authored.
> 
> My 2 cents

Yes, I need to sign them.
Since the LLM assisted, I considered to keep its tag as an information for other humans to know who made what.
Thanks!

999Marius (approved):
Looks good!

PR #1580, opened 2026-09-21
iio-debug: don't pass a null pointer to a "%s" directive

dNechita:
## PR Description

dev_dbg() and friends print the device/channel ID through a "%s" directive whose argument was __dev_id_or_null(dev), that is NULL whenever the object itself is NULL. Passing a null pointer for "%s" is undefined behaviour: glibc happens to print "(null)", but nothing requires that, and the Zephyr build supplies its own iio_prm_printf().

At -O3 GCC threads the dev == NULL path and sees the null argument for what it is, so the two call sites in xml.c that log an unknown XML field fail the build with -Werror:

  include/iio/iio-debug.h:69:60: error: '%s' directive argument is null
    [-Werror=format-overflow=]

The fix:
Substitute the literal "(null)" rather than a null pointer, and rename the macros to __dev_id_str()/__chn_id_str() as they no longer can return NULL.

## PR Type

## PR Checklist


999Marius (approved):
Looks good!

PR #1581, opened 2026-09-21
Small CI fixes on how the asset is prepared

dNechita:
## PR Description

macOS .pkg rename:
  The rename sliced the file name with byte offsets (cut -b 3-26, then
  cut -b 1-20), which only lines up for one particular combination of
  version and abbreviated hash lengths. It breaks without any version
  change at all, because git rev-parse --short auto-grows with the repo
  and now yields 8 characters where it used to yield 7, so old_name came
  out as "libiio-1.0.0.g3e5c55f6.p" and mv failed. Derive the new name
  from the file itself with a suffix substitution instead.

Windows headers:
  The public headers moved into an iio/ subdirectory in 07f40a61, and the
  artifact manifest was updated to match in 790a6c93, but this script was
  not: it still copied and removed a top-level iio.h. Handle the whole
  directory, which also gives Windows.zip the include/iio/iio.h layout
  that the <iio/iio.h> include form expects.

README.txt:
  Use ${BUILD_SOURCESDIRECTORY} rather than hardcoding /home/vsts/work/1/s.
  Identical on the agent, and it lets the script be run outside of CI.

## PR Type

## PR Checklist


999Marius (approved):
Looks good!

PR #1582, opened 2026-09-21
doc: fix Doxygen group nesting hiding public API docs

999Marius:
The `Buffer` Doxygen group opened with `@{` but was never closed before the nested `BufferStream` group started, so every group after it (BufferStream, Block, Stream, Hwmon, Events, Debug) ended up nested inside `Buffer`. This made Sphinx's `doxygengroup::` directives pull in unrelated content on the Buffer Functions page, including fields from `iio_event` further down in the header.

Also, the pointer-encoded error helpers used `@defgroup Functions for handling pointer-encoded errors`, which Doxygen parses as group name "Functions" with no page referencing it, so iio_ptr(), iio_err() and iio_err_cast() never appeared in the generated docs at all.

- Close the Buffer group before BufferStream opens
- Rename the malformed group to `ErrorHandling`
- Reference BufferStream from buffer.rst and ErrorHandling from top.rst

## PR Type

## PR Checklist


dNechita (approved):
Looks good!

